PR #16, opened 2023-01-25
deps: latest toolchain updates (bytecodealliance/wit-bindgen#457)

guybedford:
This updates to the latest toolchain updates including the new worldfile syntax.

For now the self-build still uses the old world version as this project itself always builds itself with an older version of itself currently to avoid the associated issues.

Includes the full update from `gen-host-js` being updated into the `crates/js-component-bindgen` fork.

The wasm-tools component does not yet support the full pathing semantics, that is pending preview2 support. For now, only the direct component input forms are supported.

The wasi preview2 support is also pending in the tests, which have been skipped for now pending the next wasi preview2 adapter update.


src/jsct.js
```diff
@@ -65,13 +66,10 @@ program.command('parse')
 
 program.command('new')
   .description('create a WebAssembly component adapted from a component core Wasm [wasm-tools component new]')
-  .argument('[module]', 'Wasm core module filepath')
+  .argument('[core-module]', 'Wasm core module filepath')
   .requiredOption('-o, --output <output-file>', 'Wasm component output filepath')
   .option('--name <name>', 'custom output name')
-  .option('--wit <wit>', 'WIT file to use')
```
alexcrichton, line 71: FWIW this argument is still useful to have but for the `wasm-tools` CLI I moved it to a new `wasm-tools component embed` subcommand which is explicitly only about embedding information in a core wasm module which is then later fed to `wasm-tools component new`. Similar to the `--encoding` flag that's where those moved to if you're intersted in having an `embed` command here as well.

Although to some degree if you're interested I think it would be reasonable to expose the `wasm-tools` CLI as a clap-based library. That way you wouldn't have to rewrite the CLI here if you'd like.

update-tests.sh
```diff
@@ -0,0 +1,34 @@
+git submodule init
+git submodule update
+cd wit-bindgen
```
alexcrichton, line 3: I was originally hoping that this project wouldn't depend on `wit-bindgen` which would allow `wit-bindgen` itself to depend on this project. If you'd rather stick to this route though then I'll update `wit-bindgen` to use the Wasmtime support for components and remove the js-host dependency. 

In the end I just want to remove the js-host dependency from `wit-bindgen` to not duplicate it there from here. I was hoping to use JS to run host tests since it lends itself to a bit speedier iteration times since Wasmtime doesn't have to be built, nor the embedding for Wasmtime. 

I realize though that writing tests for the component model is not easy right now. You can see [in my PR to update `wasmtime-py`](https://github.com/bytecodealliance/wasmtime-py/pull/121) how I've got a bunch of handwritten `*.wat` files but they're a pain to update and not easy to extend.

So, essentially, would you  prefer to stick to this testing strategy? If so I'll drop JS from wit-bindgen and run tests with Wasmtime. If not then the tests here would be rewritten to not use `wit-bindgen` and I'll drop JS from wit-bindgen to use this project.

src/jsct.js
guybedford: I'd like to implement the `embed` command and will need to integrate the `dummy` support as well. It's certainly a bunch of wrapper work, but I think it's useful that it can run in any JS environment including a web browser without it being tied to a CLI / WASI world. That said, with the new resolver stuff that will certainly depend on WASI I believe. Hoping to iterate through this further with the preview2 update.

update-tests.sh
alexcrichton: A major drawback of precompiled tests, which is why I do not use them, is that they basically cannot be debugged, modified, or recreated. Debugging is often crucial when the generator is changing, for example. Modifying becomes important when bugs are fixed or new features are added and tests want to be added to reflect this. Creating new tests is similar for new features.

If the JS integration here relies on wit-bindgen for tests that means that if wit-bindgen were to rely on JS for running tests that's a circular dependency. For breaking changes, like the ones that just happened, there would be no way to actually update the repositories and keep CI working. That's why I want to maintain a non-circular sequence of steps to landing breaking changes. It sounds though like I should drop the JS host generator and bring back the Wasmtime support for running the tests.

alexcrichton (approved):

update-tests.sh
guybedford: That's a great point, and I can get behind moving towards a future where all the tests are compiled internally. Perhaps a future goal to aim for might be to have all the tests written in componentize-js themselves when that is ready.

src/jsct.js
guybedford: Sure, let me see how the WASI integration goes here first. On the one hand it's nice not to have to write to temporary files for in-memory ops (eg we do currently support `import { componentNew } from 'js-component-tools'` being fully in-memory), but with the latest resolver systems we will need WASI as a dependency now anyway.

alexcrichton: To clarify, though, all of that is supported by the libraries of `wit-parser`, `wit-component`, and the `wasm-tools` crates. I'm mostly just putting this out there of there's possible more code to reuse if you'd like.

update-tests.sh
alexcrichton: I think it's reasonable to see how this plays out over time. I think a good number of tests are eventually going to be along the lines of "test this unsual shape of a component" which wouldn't be something a tool like `componentize-js` could do, but there's a lot of other runtime-y tests which could be tested too.

No matter what happens I find this aspect of the component model particularly difficult because it's not trivial to create components yet and it's also not entirely trivial to run components, so testing everything in tandem often feels a bit janky.

src/jsct.js
guybedford: At the very least we need a component build of the library interfaces though to achieve those use cases in a JS tool without requiring the Rust toolchain.

update-tests.sh
guybedford: Certainly let's continue to reassess here.

PR #27, opened 2023-01-30
feat: raw bindgen option for custom wrapping

guybedford:
This adds a new `raw_bindgen` generation option, along with flattening the generation output so that import functions are declared flat as opposed to within the initialization closure.

This option disables the imports, exports and initialization, effectively exposing a contract of:

* Imports must define functions `lowering{idx}Callee`
* Lowered imports can be referenced via `lowering{idx}`
* `memory{idx}`, `realloc{idx}`, `exports{idx}` are defined variables, but must be set externally
* Exported interfaces are provided by their exact names

Intrinsics are still output exactly, modular / separated intrinsics could be further looked into as a possible future feature.

The above bindgen with some well-defined wrapping enables more flexible embedding scenarios. I think this embedding use case is important and it's somewhat similar to the custom instantiation output except that there is no instantiation actually happening in this use case of dummy world operation, as it's a direct singular binding mapping. I'm not sure how to better categorise this use case though, to further distinguish the above defined contract, suggestions very much welcome too.

### DIFF
--- a/crates/js-component-bindgen-component/Cargo.toml
+++ b/crates/js-component-bindgen-component/Cargo.toml
@@ -13,7 +13,7 @@ crate-type = ["cdylib"]
 anyhow = { workspace = true }
 heck = { workspace = true }
 clap = { workspace = true, optional = true }
-js-component-bindgen = { path = "../js-component-bindgen" }
+js-component-bindgen = { path = "../js-component-bindgen", features = ['raw-bindgen'] }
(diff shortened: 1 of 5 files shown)


guybedford:
I've setup `raw-bindgen` as a feature for now to enable further experimentation here, we can consolidate and refactor in due course if the use case gets proven out.

elliottt (approved):
Seems reasonable to me!

PR #28, opened 2023-01-30
feat: wasm-tools component embed support

guybedford:
This adds support for the `wasm-tools component embed` command which is the new way to embed world information into component binaries now that it has been separated from the `wasm-tools component new` command.

This enables the former component new use cases within the JS toolchain and also updates Cargo dependencies.


elliottt (approved):
Looks reasonable to me!

PR #32, opened 2023-02-15
Rename to jco

guybedford:
This renames the project to "jco", with both the project name, CLI and library being under this single name. This would then be published as `@bytecodealliance/jco` on npm.

Will leave this open for now for feedback, but the plan will be to release with this renaming along with the latest changes in the next couple of days.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,4 +4,4 @@
 /obj
 /dist
 /test/output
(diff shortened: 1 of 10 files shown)


guybedford:
This is now fully live as `@bytecodealliance/jco` on npm.

PR #33, opened 2023-02-16
feat: producer metadata tooling

guybedford:
This adds the `jsct metadata` command from Wasm Tools as `wasm-tools metadata show`.

In addition `metadata` can be provided to the `embed` command and API so that when creating a component via a `new + embed` workflow, the producer metadata can be provided that way.

My tendency was not to provide the "add metadata" function since this is more designed to be a tool for consuming components than creating them, so specifically only targeting the `embed` workflow seems like it fulfills the major requirements of a standard component creation process embedding metadata such as is used in js-compute-runtime.

The embedding option didn't seem to pass through the metadata on its own so I had to add an additional write to this. Unsupported fields throw in the embedding process.

### DIFF
--- a/crates/wasm-tools-component/src/lib.rs
+++ b/crates/wasm-tools-component/src/lib.rs
@@ -158,6 +158,27 @@ impl exports::Exports for WasmToolsJs {
             binary.unwrap()
         };
 
+        let mut producers = Producers::default();
+        match opts {
+            Some(ref opts) => match &opts.metadata {
+                Some(metadata_fields) => {
+                    for (field_name, items) in metadata_fields {
+                        if field_name != "sdk"
+                            && field_name != "language"
+                            && field_name != "processed-by"
+                        {
+                            return Err(format!("'{field_name}' is not a valid field to embed in the metadata. Must be one of 'language', 'processed-by' or 'sdk'."));
+                        }
+                        for (name, version) in items {
+                            producers.add(&field_name, &name, &version);
+                        }
+                    }
+                }
+                None => {}
+            },
+            None => {}
+        };
+
         let encoded = wit_component::metadata::encode(&resolve, world, string_encoding, None)
             .map_err(|e| e.to_string())?;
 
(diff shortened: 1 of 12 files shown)


guybedford:
Nice pretty printing is also included for the output. Here's an example...

```
[component]
  processed-by: wit-component 0.7.0

  [module]
    language:     C99
    processed-by: clang 15.0.6 (https://github.com/llvm/llvm-project 088f33605d8a61ff519c580a71b1dd57d16a03f8)
    processed-by: wit-component 0.7.0
    processed-by: wit-bindgen-gen-guest-c 0.3.0

  [module wit-component:adapter:wasi_snapshot_preview1]
    language:     Rust
    processed-by: rustc 1.68.0-nightly (c7572670a 2023-01-03)

  [module wit-component:shim]
    processed-by: wit-component 0.7.0

  [module wit-component:fixups]
    processed-by: wit-component 0.7.0

  [component]
    (no metadata)
```

crates/wasm-tools-component/src/lib.rs
```diff
@@ -165,9 +169,85 @@ impl exports::Exports for WasmToolsJs {
...
+                        producer_section.field(&field_name, &field);
+                    }
+                }
+                None => {}
+            },
+            None => {}
+        };
+
+        core_binary.push(producer_section.id());
+        producer_section.encode(&mut core_binary);
```
pchickey, line 197: Instead, please use `add_to_wasm` https://docs.rs/wasm-metadata/latest/wasm_metadata/struct.Producers.html#method.add_to_wasm, which will add to the module's existing producers section of it has one.

crates/wasm-tools-component/wit/wasm-tools.wit
```diff
@@ -20,15 +20,34 @@ default world wasm-tools-js {
     /// Extract a *.wit interface from a component, optionally providing a document name to extract
     component-wit: func(binary: list<u8>, document: option<string>) -> result<string, string>
 
+    type metadata-fields = list<tuple<string, list<tuple<string, string>>>>
```
pchickey, line 23: I would just call this `producers-fields`

crates/wasm-tools-component/wit/wasm-tools.wit
```diff
@@ -20,15 +20,34 @@ default world wasm-tools-js {
...
+    variant module-meta-type {
+      module,
+      // the number of nested modules
+      component(u32),
+    }
+
+    record module-metadata {
+      name: option<string>,
+      meta-type: module-meta-type,
+      metadata: metadata-fields,
```
pchickey, line 45: and this member `producers`

pchickey (approved):
Looks good! One change and two minor nits

guybedford:
I've updated the metadata handling to use the existing `embed` path for metadata inclusion, then also separated `metadata` into `metadata show` and `metadata add`.

pchickey (approved):

PR #37, opened 2023-02-18
feat: add wasi preview2 shim for browser and nodejs

eduardomourar:
This a basic shim for WASI (preview2) in JavaScript. The browser part is almost as is from https://github.com/guybedford/wasi-command-export-printf/tree/main/wasi and for Node.js it implements a small percentage of the API. Keep in mind that this is packaged as an ES module to be consumed for both environments without any workaround or hack.


packages/wasi-common/package.json
```diff
@@ -0,0 +1,18 @@
+{
+    "name": "@bytecodealliance/wasi-common",
+    "version": "0.0.1",
+    "description": "WASI implementation in JavaScript for the Browser and Node.js",
+    "author": "Guy Bedford, Eduardo Rodrigues<16357187+eduardomourar@users.noreply.github.com>",
+    "type": "module",
+    "main": "./index.js",
+    "browser": "./browser/index.js",
```
guybedford, line 8:
```suggestion
    "exports": {
      "node": "./nodejs/index.js",
      "default": "./browser/index.js"
    }
```

packages/wasi-common/package.json
```diff
@@ -0,0 +1,18 @@
+{
+    "name": "@bytecodealliance/wasi-common",
```
guybedford, line 2: Sounds good, but perhaps let's make it release-specific and note that this is not a "proper" implementation but a "shim" -

```suggestion
    "name": "@bytecodealliance/preview2-shim",
```

guybedford (approved):
Thanks for putting this together, it's almost entirely identical to what I've been using myself here.

Definitely agreed on the overall direction, just see the few comments.

Happy to merge and iterate from there.

packages/wasi-common/nodejs/exit.js
```diff
@@ -0,0 +1,4 @@
+export function exit(status) {
+  console.log(`[exit] Exit: ${JSON.stringify(status)}`);
+  process.exit(1);
```
guybedford, line 3: This should probably be code-specific:

```suggestion
  process.exit(status.tag === 'err' ? 1 : 0);
```

packages/wasi-common/types/wasi-exit.d.ts
```diff
@@ -0,0 +1,3 @@
+export namespace WasiExit {
+  export function exit(status: Result<void, void>): void;
```
guybedford, line 2:
```suggestion
  export function exit(status: { tag: 'err' | 'ok' }): void;
```

guybedford:
We should probably be generating all the types directly from the WASI wit definitions as well to ensure we're in sync.

eduardomourar:
I have made a commit with the experimental outbound HTTP request work I have been doing (related to this other PR [here](https://github.com/bytecodealliance/preview2-prototyping/pull/85)). I guess I should revert that, right?

guybedford:
I'm more than happy to just ship things in any state here, as it is highly experimental and also because it is just a shim - so any iterative improvement in terms of things working will always be a win.

I'm tempted to post a first npm publish later this afternoon, unless you can think of any reasons not to!

eduardomourar:
> I'm more than happy to just ship things in any state here, as it is highly experimental and also because it is just a shim - so any iterative improvement in terms of things working will always be a win.
> 
> 
> 
> I'm tempted to post a first npm publish later this afternoon, unless you can think of any reasons not to!

Publishing to NPM will actually make my life so much easier!

PR #38, opened 2023-02-20
fix: treat synckit as an optional dependency

guybedford:
This treats `synckit` used in the Node.js http shim as an optional dependency.

@eduardomourar ideally I think we should fully inline any dependency usage if we can, as it would be great to have no dependencies for this.

### DIFF
--- a/packages/preview2-shim/lib/nodejs/http.js
+++ b/packages/preview2-shim/lib/nodejs/http.js
@@ -1,12 +1,13 @@
-import { createSyncFn } from "synckit";
-import path from "node:path";
+import { createRequire } from 'node:module';
 import { fileURLToPath } from "node:url";
 import { UnexpectedError } from "../http/error.js";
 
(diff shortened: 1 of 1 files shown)


eduardomourar:
fully agree on this

eduardomourar:
i guess we should include synckit to the [optionalDependencies within package.json](https://docs.npmjs.com/cli/v9/configuring-npm/package-json#optionaldependencies), right?

guybedford:
Unfortunately optionalDependencies in Node.js don't really work since they are only based on "engines" field filtering, as opposed to actual usage.

Usually I'd just prefer to use devDependencies and bundle any dependencies in with a build. It depends on the footprint though. Synckit doesn't seem _too_ bad (https://bundlephobia.com/package/synckit@0.8.5).

eduardomourar:
The underlying problem that I tried to solve is this one here: https://github.com/bytecodealliance/jco/issues/39

PR #50, opened 2023-03-20
fix: define export interface

eduardomourar:


guybedford (approved):
Thanks for this, can you just run `rust fmt` for the CI to pass?

PR #66, opened 2023-04-26
feat: bindgen refactor

guybedford:
This refactors the bindgen into low-level `function_bindgen` which is then used by separate `transpile_bindgen`, `componentize_bindgen` and `ts_bindgen`.

The componentize bindgen has been fully refactored out, replacing the need for the previous raw_bindgen.


crates/js-component-bindgen/src/intrinsics.rs
```diff
@@ -0,0 +1,372 @@
+use crate::source::Source;
```
guybedford, line 1: Intrinsics generation is separated out.

crates/js-component-bindgen/src/function_bindgen.rs
```diff
@@ -0,0 +1,1195 @@
+use crate::intrinsics::Intrinsic;
```
guybedford, line 1: Function bindgen is pretty much copy-pasted from the original, but as the target for low-level bindgen.

crates/js-component-bindgen/src/ts_bindgen.rs
```diff
@@ -0,0 +1,675 @@
+use crate::files::Files;
```
guybedford, line 1: TypeScript bindgen is separated into a separate bindgen pass, as distinct from transpilation bindgen. Instead treating bindgen types as their own separate passes entirely.

crates/js-component-bindgen/src/transpile_bindgen.rs
```diff
@@ -0,0 +1,711 @@
+use crate::files::Files;
```
guybedford, line 1: Transpile bindgen is included, and made a feature. Raw bindgen is removed. Instead componentize bindgen can be built on top of function bindgen directly. The entry point to the bindgen is treated as a function instead of a stateful options initialization.

alexcrichton (approved):

Issue #42, opened 2023-02-26
Browser support


bjorn3:
I did like to use jco for handling wasi preview2 in browser_wasi_shim. Is it possible to use jco directly in the browser to take a wasm component, run the transpilation step and directly run it as a shim until browsers natively support the wasm component model? Or does it only run in nodejs and does it only support writing out javascript files to the filesystem that can later be imported rather than directly evaluated using `new Function` or `eval`?

guybedford:
A browser build would be straightforward to add, and we could support all API methods except for `optimize`. It would just require careful handling of the imports in the API entry point to not throw in browsers. Marking this issue to track the feature.

DougAnderson444:
+1 as I'm running into this now when I try to import `transpile` in a browser context:

```js
// in the browser, doesn't work
import { transpile } from '@bytecodealliance/jco';
```

It can't be done because top level await and a module called "module" that Webpack inserts. 

Even if I were to import it, could I call `transpile` in the browser?

DougAnderson444:
Dug a bit deeper, I noticed that [`js-component-bindgen-component`](https://github.com/bytecodealliance/jco/blob/f1364c9faa128f9fd518cff77b6dbbf4567057a8/crates/js-component-bindgen-component/wit/js-component-bindgen.wit#L54) exports `{ generate }` which output the files without any NodeJS deps:

```js
let { files, imports, exports } = generate(component, opts);
```

...we could use [rollup](https://rollupjs.org/faqs/#how-do-i-run-rollup-itself-in-a-browser) to bundle those deps into a script directly usable by the browser.

lee-orr:
That only works if you work on the assumption that the artifact can/should be changed for different targets. If you want to be able to deploy a single wasm component and have it run in multiple environments, having to then do a separate step for just the browser defeats a decent chunk of the purpose. Especially for use-provided logic.

DougAnderson444:
@guybedford I finished writing the Rollup Plugin 🚀

https://github.com/DougAnderson444/rollup-plugin-wit-component

This plugin essentially bundles all the dependencies from

```js
// import { generate } from 'obj/js-component-bindgen-component.js'
let { files, imports, exports } = generate(wasmBytes, opts);
```

into a single file that browsers can use to `import` the javascript functions.

This bundling step can even occur in the browser itself! That is what the demo does.

[...]

PR #84, opened 2023-05-26
[WIP] move node modules into single dependency

lee-orr:
Starting work towards supporting  #42 (in progress)


guybedford:
Browser support for jco is now working, see the example in https://github.com/bytecodealliance/jco/blob/main/test/browser.html.

Issue #110, opened 2023-07-20
Missing original export name with "inline" wit interface


kajacx:
## Summary

When I have a wasm component with a wit world that has "inline interfaces" (see bellow), when I transpile it with `jco transpile component.wasm --instantiation -o out-dir`, the generated TS bindings for the exported "inline interface" are wrong. Other TS types seem to be correct.

## Full example

Full reproducible example can be found [here](https://github.com/kajacx/wasm-playground/tree/wit-bindgen-inline-interface-wrong-ts), see the [`plugin-rust`](https://github.com/kajacx/wasm-playground/tree/wit-bindgen-inline-interface-wrong-ts/plugin-rust) guest and [`runtime-rust-web`](https://github.com/kajacx/wasm-playground/tree/wit-bindgen-inline-interface-wrong-ts/runtime-rust-web) host.

## Wit file
```wit
package example:protocol

interface host-imports {
  print-line: func(msg: string)
}

interface guest-exports {
  run: func()
}

world my-world {
  import host-imports
  export guest-exports

  import inline-imports: interface {
    add-one: func(num: s32) -> s32
  }
  export inline-exports: interface {
    add-three: func(num: s32) -> s32
  }
}
```

## Generated TS bindings
```ts
export interface Root {
  'example:protocol/guest-exports': typeof ExportsExampleProtocolGuestExports,
  'inline-exports': typeof ExportsInlineExports,
  guestExports: typeof ExportsExampleProtocolGuestExports,
  inlineExports: typeof ExportsInlineExports,
}
```

## Actual behaviour

The actual "instance" object is missing the `'inline-exports'` object, it only has `inlineExports`.

Ideally, it would have both to make it consistent with how "named" exports also have both variants.

## Edit

I changed the name, because it seems that the TS bindings are "correct" and the JS code is wrong. In other words, it would be better to update the JS code to match the TS bindings and not the other way around.

guybedford:
Thanks for pointing out this disparity! None of this is set in stone at all, and we are working out these conventions, but the complexity here is that the component model very much does specify that language specific naming conventions should be used where appropriate. Ie conversion from kebab to camel case.

The mental model I use here is to think of the interface names (with a `:`) as in "canonical interface space" but to think of normal exports as in "kebab name space" that can be converted into "camel case space" because that is what you would use in JS. The set of exports is then a union of these spaces with aliases into the canonical interface space but not the kebab name space.

[...]

PR #111, opened 2023-07-20
Unify interface aliasing

kajacx:
Fixes https://github.com/bytecodealliance/jco/issues/110

## Quick summary:

"Named" interface exports have both original name and aliased name in exports, but "local"/"inline" interfaces have only aliased export in the exported JS object.

```
interface guest-exports {
  run: func()
}

world my-world {
  export guest-exports // "named" interface

  export inline-exports: interface { // "local"/"inline" interface
    add-three: func(num: s32) -> s32
  }
}
```

TS bindings are already "correct" (according to what *should* be exported):
```ts
export interface Root {
  'example:protocol/guest-exports': typeof ExportsExampleProtocolGuestExports,
  'inline-exports': typeof ExportsInlineExports, // was missing in actual JS code
  guestExports: typeof ExportsExampleProtocolGuestExports,
  inlineExports: typeof ExportsInlineExports,
}
```

But the JS code did not export the non-aliased local interface export.

### DIFF
--- a/crates/js-component-bindgen/src/ts_bindgen.rs
+++ b/crates/js-component-bindgen/src/ts_bindgen.rs
@@ -329,23 +329,30 @@ impl TsBindgen {
     ) -> String {
         let local_name =
             self.generate_interface(export_name, resolve, id, files, AbiVariant::GuestExport);
-        if instantiation {
-            uwriteln!(
-                self.export_object,
-                "{}: typeof {local_name},",
-                maybe_quote_id(export_name)
-            );
-        } else {
-            if export_name != maybe_quote_id(export_name) {
-                // TODO: TypeScript doesn't currently support
-                // non-identifier exports
-                // tracking in https://github.com/microsoft/TypeScript/issues/40594
-            } else {
+
+        // Only export the original interface name if it comes from an external interface,
+        // BUT export "inline" interface if the original and aliased names are the same,
+        // because it would be skipped later
+        // See https://github.com/bytecodealliance/jco/pull/111
+        if export_name.contains(":") || export_name == local_name {
+            if instantiation {
                 uwriteln!(
                     self.export_object,
-                    "export const {}: typeof {local_name};",
+                    "{}: typeof {local_name},",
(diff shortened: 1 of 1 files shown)


guybedford (commented):
Note my feedback on this PR is as per https://github.com/tc39/proposal-source-phase-imports/issues/49#issuecomment-1593876714.

guybedford:
I realise I gave the wrong link to my feedback, I meant to link to https://github.com/bytecodealliance/jco/issues/110#issuecomment-1644321422.

The point remains that we should not merge this as-is, since the model is correct, and it is in fact that TS case that is incorrect providing the kebab name.

kajacx:
Ok, I almost forgot about this issue. The workaround is really easy, you can see it [here](https://github.com/kajacx/wasm-bridge/blob/master/crates/wasm-bridge-js/src/component/exports.rs#L80), so it's not a big deal if the JS code stays as it is. But it's true that the TS types should reflect the JS code, I'll try to change the PR.

kajacx:
Ok, this code seems to "fix" the discrepancy in the TS types, but the code is kind of bad. It "exploits" the fact that external interface names contain a ":" in their name in a really bad way.

The proper solution would have been to store the interface name in a struct that would properly contain fields for the interface name itself, as well as the path of where it comes from, and then decide based on that instead of looking at the complete string representation.

But that would would require a significant refactor of the code, which is out of scope of this PR.

I still think this is the wrong choice, and exporting the original name from JS as well would have been better, but it's your project, you decide in which direction you want to go.

guybedford:
Thank you @kajacx for the update here, this would be great to land as soon as the tests are passing.

To reiterate the main argument with respect to naming it is that JavaScript exports use idiomatic camel casing not kebab casing. It's only interface names which exist as canonical kebab interface names on exports for this specific deduping case. Perhaps in future we will have better middle ground deduping like namespaced camel conversions that we use instead here.

guybedford:
Closing as discussed in https://github.com/bytecodealliance/jco/issues/110#issuecomment-1644321422.

PR #118, opened 2023-07-27
Update js wasi clocks

kajacx:
performance.now() returns milliseconds, and we want nanoseconds, so we need to multiply by a million, not a billion.

Using one `Date.now()` prevents a "bad" case when one first Date.now() call returns a different date than the second Date.now() call, which can be especially bad if the two dates "cross over" the millisecond boundary.


guybedford (approved):
Thanks so much for PRing

PR #119, opened 2023-08-03
Update the preview1 adapter

elliottt:
Change the shell script to remove the right files in `lib`, and run it to update the preview1 adapters.


guybedford:
Thanks for the PR! This should be superseded by https://github.com/bytecodealliance/jco/pull/120 now.

PR #124, opened 2023-08-07
Update to Wasmtime 12

alexcrichton:
This commit updates the Wasmtime dependency to the, currently unreleased, Wasmtime 12.0.0 branch. A number of changes happened to component translation in Wasmtime 12 to account for resources which required changes here as well. Additionally Wasmtime 12 disallows empty types in components which WASI was previously using, so this additionally updates the preview1 shims and test components.

PR #125, opened 2023-08-07
 Support polyfilling multi-memory 

alexcrichton:
Currently JS does not generally support the WebAssembly multi-memory
proposal. Wasmtime will, however, generate adapter modules which use
multi-memory to communicate between components. This means that composed
components are typically not compatible with the transpile process as
they produce a core module that doesn't actually run in any JS runtime.

This commit fixes this issue by adding polyfill support for
multi-memory. Whenever an adapter is produced that uses multi-memory jco
will now rewrite the module such that any references to memory that
isn't at index 0 to be indirected through functions. These imported
functions then operate on the specified memory on behalf of the wasm
itself. This is a horribly slow process because all memory reads/writes
become function calls, but given that the baseline is otherwise "does
not work" it's hopefully a bit better than before.

The end goal here is to work up towards strings between components, but
for now this just gets everything else working with multi-memory such as
transferring lists.

### DIFF
--- a/crates/js-component-bindgen/src/core.rs
+++ b/crates/js-component-bindgen/src/core.rs
@@ -1,0 +1,794 @@
+//! Support for transpiling core modules using multi-memory to those that don't
+//! use multi-memory.
+//!
+//! Wasmtime's implementation of adapter modules between components requires the
+//! usage of multi-memory for copying data back and forth between two
+//! components. The multi-memory proposal is, at this time, not stable in any JS
+//! engine. This module is an attempt to polyfill this until at such a time that
+//! multi-memory can be used natively.
+//!
+//! The purpose of this module is to identify core wasms which require
+//! multi-memory coming out of Wasmtime. These wasms are rewritten to not
+//! actually use more than one memory. The implementation here is to replace all
+//! memory instructions operating on memory index 1 or greater with function
+//! calls where JS is the one that does the load/store/etc. This is not expected
+//! to be fast at runtime but is intended to be just enough to get this working
+//! in JS environments at this time. The true speed is expected to come with the
+//! multi-memory proposal.
+//!
+//! This module exports a [`Translation`] which wraps a [`ModuleTranslation`]
+//! either as a pass-through "normal" or an "augmented" version where
+//! "augmented" means that the original wasm is not used but instead a
+//! recompiled copy without multiple memories is used. When calculating the
+//! imports for the "augmented" module the arguments for JS functions that
+//! read/write memory are automatically injected and handled.
+//!
+//! Callers of this module need to have an implementation in JS for all of the
+//! entries listed in [`AugmentedOp`], likely through the `DataView` class in
+//! JS.
+//!
+//! Note that at this time this module is not intended to be a complete and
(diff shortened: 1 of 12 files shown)


alexcrichton:
I'll note that this is based on https://github.com/bytecodealliance/jco/pull/124 because Wasmtime 12 has a [minor fix](https://github.com/bytecodealliance/wasmtime/pull/6783) required to get this working.

crates/js-component-bindgen/src/core.rs
```diff
@@ -0,0 +1,794 @@
...
+        }
+    }
+}
+
+struct Translator<'a, 'b> {
+    func: &'a mut wasm_encoder::Function,
+    augmenter: &'a Augmenter<'b>,
+}
+
+// TODO
```
guybedford, line 613: Can you explain what this todo item is?

crates/js-component-bindgen/src/transpile_bindgen.rs
```diff
@@ -331,6 +342,45 @@ impl Instantiator<'_, '_> {
...
+            // component and may change encodings. This is left unimplemented
+            // for now since it can't be tested and additionally JS doesn't
+            // support multi-memory which transcoders rely on anyway.
+            Trampoline::Transcoder {
+                op: _,
+                from: _,
+                from64: _,
+                to: _,
+                to64: _,
+            } => unimplemented!(),
```
guybedford, line 363: Is this path then effectively skipped in the polyfill case?

guybedford (commented):
Fantastic, I've been worried about this. Will try it out first thing tomorrow.

crates/js-component-bindgen/src/core.rs
```diff
@@ -0,0 +1,794 @@
...
+//! JS.
+//!
+//! Note that at this time this module is not intended to be a complete and
+//! general purpose method of compiling multiple memories to single-memory
+//! modules. This does not handle all instructions that use memory for example,
+//! but only those that Wasmtime's adapter modules emits. It's possible to add
+//! support for more instructions but such support isn't required at this time.
+//! Examples of unsupported instructions are `i64.load8_u` and `memory.copy`.
+//! Additionally core wasm sections such as data sections and tables are not
+//! supported because, again, Wasmtime doesn't use it at this time.
```
guybedford, line 37: I was wondering if it might make sense to use a full wasm2js solution here (eg something like https://github.com/evanw/polywasm), but this certainly seems adequate as well.

crates/js-component-bindgen/src/transpile_bindgen.rs
alexcrichton: Nah this'll still needs to get filled out, but I wanted to tackle this PR first as a first step before tackling strings

crates/js-component-bindgen/src/core.rs
alexcrichton: I think any of those would work myself, this was just the first thing I thought of so I figured I'd take a stab implementing it while stuck delayed in an airport :)

If you'd prefer to go with those though I think it's ok to hold this open or close it, either way's fine by me

PR #139, opened 2023-08-29
feat(preview2-shim): update wasi http to latest

eduardomourar:
This will update to the latest definition of the WASI HTTP proposal. Additionally, some WIT files have been removed as they are not being used anymore.


guybedford:
Great to see this, thanks for continuing to maintain this work!

Issue #145, opened 2023-09-04
Invalid function parameters in TypeScript definition


eduardomourar:
Whenever trying to use the interfaces from the preview2-shim directly, you get the following compilation error:
> ERROR in /project/imports/wasi-poll-poll.d.ts 3:29-31
[tsl] ERROR in /project/imports/wasi-poll-poll.d.ts(3,30)
      TS1359: Identifier expected. 'in' is a reserved word that cannot be used here.

PR #146, opened 2023-09-04
fix(js-component-bindgen): reserved word in typescript function parameter

eduardomourar:
We will prevent reserved words of being used as function parameters within generate TypeScript definitions. It will append an extra `_` (underscore) character

Fixes: https://github.com/bytecodealliance/jco/issues/145


guybedford (approved):
Looks good, thanks

Issue #141, opened 2023-08-30
Preview2 shim instantiation


guybedford:
Just like Node.js WASI supports instancing, we should make our preview2 shim implementation instancing based.

That is, still have the exact same exports we have right now, but have a new entry point like `@bytecodealliance/preview2-shim/factory` that provides a high-level class that instantiates into the various interfaces. Internally we should ideally then use this as the base dependency, while trying to keep cross-dependence to a minimum.

Perhaps each subsection will be its own factory - `@bytecodealliance/preview2-shim/factory/filesystem` say.

This will enable custom virtualization / reinstancing workflows. Eg for testing etc.

PR #147, opened 2023-09-04
feat: extend browser wasi support

guybedford:
This extends browser WASI interfaces to support a virtual filesystem and stdio streams.


guybedford:
Whether we maintain this direction or not is still a little unclear, but let's land and see how it goes.

With https://github.com/bytecodealliance/jco/issues/141, we will at the very least need an instantiable FS implementation anyway so this allows a path to developing the codebase.

Issue #149, opened 2023-09-08
Include tracing to host bindgen


eduardomourar:
Similar to how it is being done in wasmtime (code [here](https://github.com/bytecodealliance/wasmtime/blob/181d005c45965b580b05df7bdaa29a8a4b4e5827/crates/wit-bindgen/src/lib.rs#L94)), we should add tracing details for the host's generated binding code.

PR #150, opened 2023-09-10
feat(jco): add flag to enable tracing details

eduardomourar:
Similar to how it is being done in wasmtime (code [here](https://github.com/bytecodealliance/wasmtime/blob/181d005c45965b580b05df7bdaa29a8a4b4e5827/crates/wit-bindgen/src/lib.rs#L94)), this will add tracing details for the host's generated binding code.

Closes: https://github.com/bytecodealliance/jco/issues/149.


crates/js-component-bindgen/src/intrinsics.rs
```diff
@@ -96,6 +97,12 @@ pub fn render_intrinsics(
                 const hasOwnProperty = Object.prototype.hasOwnProperty;
             "),
 
+            Intrinsic::ToResultString => output.push_str("
+                function toResultString(obj) {
+                    return JSON.stringify(obj, (_, v) => typeof v === 'bigint' ? v.toString() : v);
```
guybedford, line 102: Just wondering if this might be quite noisy in the case of big return objects. Perhaps we can make the tracing a config object, and make this an option, `{ results: true }`, say?

guybedford (approved):
Great to see this moving forward, will be glad to release it and iterate from there.

crates/js-component-bindgen/src/intrinsics.rs
eduardomourar: i would rather us limit the string size and/or object depth (like [here](https://github.com/formfunction-hq/formfunction-monorepo/blob/6d049d5b1ff852e6d9f0126b80924e989e1f257e/packages/shared/src/utils/jsonStringify.ts#L9)) with some sensible defaults.

guybedford: I think blanket turning off would be useful but we can land separately as well. For example - currently `new Uint8Array([1,2,3])` will output its bytes in this position, which is definitely not what we want.

guybedford: Can we just disable array buffers perhaps for now?

eduardomourar: instances of `Uint8Array` have been disabled. your example will now print `... result="[Uint8Array (3)]"`

crates/js-component-bindgen/src/intrinsics.rs
```diff
@@ -96,6 +97,19 @@ pub fn render_intrinsics(
                 const hasOwnProperty = Object.prototype.hasOwnProperty;
             "),
 
+            Intrinsic::ToResultString => output.push_str("
+                function toResultString(obj) {
+                    return JSON.stringify(obj, (_, v) => {
+                        if (v instanceof Uint8Array) {
+                            return `[Uint8Array (${v.byteLength})]`;
+                        } else if (typeof v === 'bigint') {
```
guybedford, line 105: We do also support other data array type lifting in returns, so I think this should be something more like:

```suggestion
                        if (v instanceof Uint8Array.prototype) {
                            return `[${v[Symbol.toStringTag]} (${v.byteLength})]`;
                        } else if (typeof v === 'bigint') {
```

PR #158, opened 2023-09-13
fix(preview2-shim): missing response http headers

eduardomourar:
This will fix the response headers for WASI HTTP. Additionally, this will add a round-trip (from creating a component in JavaScript, transpiling it and running it in Node.js) test for the WASI HTTP.


eduardomourar:
@guybedford, because we are using the fetch API, I had to set the minimum Node.js version to 18.

package.json
```diff
@@ -3,6 +3,9 @@
   "version": "0.11.2",
   "description": "JavaScript tooling for working with WebAssembly Components",
   "author": "Guy Bedford",
+  "engines" : { 
+    "node" : ">=18"
```
guybedford, line 7: I generally prefer to leave this out, because it's just such a bad bad UX for users. Rather users can just hit the bug when running on 16. It's EOL next month anyway so I have no problem with that.

So please let's keep the package.json unrestricted here.

.github/workflows/main.yml
```diff
@@ -45,9 +45,9 @@ jobs:
         echo "WASI_SDK_PATH=`pwd`/wasi-sdk-16.0" >> $GITHUB_ENV
       if : matrix.os == 'windows-latest'
 
-    - uses: actions/setup-node@v2
+    - uses: actions/setup-node@v3
       with:
-        node-version: '16'
+        node-version-file: "package.json"
```
guybedford, line 50: Let's just make this `lts`, then it should be roughly what we want going forward.

guybedford (approved):
Thanks, great to see the thorough tests here.

Since 16 is EOL next month anyway I'm totally fine with landing this as lts. But let's remove the package.json restriction for compat too.

Issue #153, opened 2023-09-12
Jco transpile: nonblocking implementation of poll-oneoff


pchickey:

yoshuawuyts:
@pchickey can you assign me to this issue please? Thank you!

vados-cosmonic:
Hey @yoshuawuyts can we close this?

vados-cosmonic:
Closing this as we have an implementation of [poll](https://github.com/bytecodealliance/jco/blob/e8103124a31d916188555ec8dd39bbd3e708b975/packages/preview2-shim/lib/io/worker-io.js#L486) -- happy to re-open if something was still missing here!

PR #161, opened 2023-09-19
feat: add pollable promise

eduardomourar:
This will prepare the ground by defining a `PollablePromise` class that can execute the inner promise every time `poll-oneoff` is called.

Related to: https://github.com/bytecodealliance/jco/issues/153.


guybedford:
Thanks @eduardomourar this looks like a huge step in the right direction to me.

If we had comprehensive tests, I would merge straight away, but I'm looking into these async cases right now so will likely be making further changes here in the coming days to the code base. I don't want to break your work here unless we have a full end to end test suite for it.

So I'd ask you to consider which option you'd prefer - wait a couple of days to rebase / land this, add tests and land this, or land this now and risk my subsequent changes altering things?

eduardomourar:
I don't mind us merging as is because we are exercising most of the poll logic through the `wasi-http-proxy` integration test.  Besides, I do expect this to be modified anyway as part of your upcoming changes.

guybedford:
This is now superseded by https://github.com/bytecodealliance/jco/pull/244.

PR #162, opened 2023-09-20
deps: upgrade toolchain & WASI versions to latest

guybedford:
This upgrades the toolchain dependencies to the latest.

* wit-bindgen: https://github.com/bytecodealliance/wit-bindgen/commit/637389a0409e83597807b71bade607c383ad0cf8
* wasm-tools: https://github.com/bytecodealliance/wasm-tools/commit/c7bb472f72b7d6fe2afac81919b9b65ce8dd5039
* wasmtime: https://github.com/bytecodealliance/wasmtime/commit/a4d036ca8e9ea56c23521ade7c847000d624a368
* WASI: a4d036ca8e9ea56c from Wasmtime


crates/js-component-bindgen/src/core.rs
```diff
@@ -347,7 +347,7 @@ impl Augmenter<'_> {
         // before.
         let mut types = TypeSection::new();
         for ty in self.types.iter() {
-            assert!(!ty.is_final);
+            // assert!(!ty.is_final);
```
guybedford, line 350: This assertion removal was necessary to get the upgrade tests to pass, but needs to be looked into further.

@alexcrichton would value your eyes on this if you have a moment.

eduardomourar:
@guybedford, as there are a lot of breaking changes in the WASI definitions, could we land [this PR](https://github.com/bytecodealliance/jco/pull/163) first, please? That way we will have a commit in main that matches the (soon to be) latest stable version of wasmtime.

crates/js-component-bindgen/src/core.rs
alexcrichton: Ah here I think you can change `self.types.iter()` to [`into_iter_err_on_gc_types`](https://docs.rs/wasmparser/latest/wasmparser/type.TypeSectionReader.html#method.into_iter_err_on_gc_types) which should handle this for you now (that didn't exist when I first wrote this)

guybedford: Thanks.

PR #163, opened 2023-09-20
feat(wasi): sync wit definition with wasmtime v13

eduardomourar:
This will update the WIT definitions and adapters to match the [wasmtime v13 release](https://github.com/bytecodealliance/wasmtime/releases/tag/v13.0.0).

The implementation for the flushing logic is missing as we are simply returning a huge amount for `check-write`. The tests are exercising the expected behavior as done in the test in wasmtime, though.


package.json
```diff
@@ -54,7 +54,7 @@
     "build": "cargo build --workspace --target wasm32-wasi --release && cargo run",
     "build:types:preview2-shim": "PREVIEW2_SHIM_TYPES=* cargo build -p jco",
     "lint": "eslint -c eslintrc.cjs lib/**/*.js packages/*/lib/**/*.js",
-    "test": "mocha -u tdd test/test.js --timeout 120000"
+    "test": "rm -rf $(pwd)/node_modules/@bytecodealliance/jco/lib && ln -s $(pwd)/lib $(pwd)/node_modules/@bytecodealliance/jco/ && mocha -u tdd test/test.js --timeout 120000"
```
guybedford, line 57: This shouldn't be necessary, and is brittle giving the lib folder special precedence. Any code that uses jco should refer to the local `src/jco.js` path.

packages/preview2-shim/lib/http/wasi-http.js
```diff
@@ -2,18 +2,18 @@
 // https://github.com/bytecodealliance/wasmtime/blob/8efcb9851602287fd07a1a1e91501f51f2653d7e/crates/wasi-http/
 
 /**
- * @typedef {import("../../types/imports/wasi-http-types").Fields} Fields
```
guybedford, line 5: Curious why this didn't regenerate with the script previously?

guybedford (approved):
I'd be glad to merge this, but was hoping to do a release today.

package.json
eduardomourar: it is failing without it because the adapters are incompatible with jco in npm (being pulled from componentize-js dependency)

guybedford: Ah I see, in that case, this actually seems like a great workaround, thanks for explaining!

packages/preview2-shim/lib/http/wasi-http.js
eduardomourar: i don't think this file is update by the scripts at all. there is probably some way to get the types from the entrypoint instead, but i could not get it to work a while back

Issue #172, opened 2023-10-05
Consider switching to a `cargo xtask`-based runner


yoshuawuyts:
As I'm diving into the repo I've noticed there are number of scripts in the project root, as well as a `bin/` directory containing what seems to be utility scripts. I would like to suggest we unify these approaches using the [cargo xtask](https://github.com/matklad/cargo-xtask) pattern, which would give us a common interface to perform the various actions on the repo. I was thinking something along these lines:

```sh
$ cargo xtask update tests     # replaces ./update-tests.sh
$ cargo xtask update wasi      # replaces ./update-wasi.sh
$ cargo xtask update preview2  # replaces ./update-preview2.sh
$ cargo xtask test             # frontend for `npm test`
$ cargo xtask build            # frontend for `npm install && npm run build`
```

This approach is quite common in the Rust ecosystem, including now being adopted by `rust-lang/rust`. To view an example of this in action see [yoshuawuyts/html](https://github.com/yoshuawuyts/html), which provides a common `xtask` frontend for various scraping/parsing/codegen commands.

guybedford:
Sure, I'm all for it. Separately, I'm currently working on a PR for Windows support for CI and fixing the Windows support, which I'm hoping to get up today. So I'd recommend basing to that PR if possible.

yoshuawuyts:
I think now that it's setup we can continue working on it, and this should be fine to close.

PR #173, opened 2023-10-05
Initialize `cargo xtask`

yoshuawuyts:
Makes progress towards #172, setting up the skeleton for `cargo xtask` and moving the `self_build` command into it as `cargo xtask build`. I intentionally kept this minimal, so it wouldn't interfere with the existing shell script refactoring. Thanks!


yoshuawuyts:
I tried moving all of `bin/self_build` to exist under `xtask`. But it seems that the `build.rs` in jco's root is load-bearing. I'm not exactly sure when it's supposed to run or what it's for - so some guidance on that would be helpful.

yoshuawuyts:
Yay okay, I think I'm getting the hang of the various build scripts. The issue was that I wasn't updating the `package.json` scripts to point to `xtask`, so every time I moved something it would break. I'm feeling somewhat confident I can move more things now!

yoshuawuyts:
Okay, I think this should be ready for review now. To explain the changes I've made in this PR:

- Added a new `cargo xtask` sub-command. This provides a central location for all Rust-based project scripting, like a counterpart to npm scripts.
- Added `cargo xtask build-self`, which replaces the previous `cargo run` functionality.
- Added `cargo xtask build`, which is equivalent to calling `cargo build --workspace --target wasm32-wasi --release && cargo xtask build-self`.
- Added `cargo xtask build-shims`, which replaces `PREVIEW2_SHIM_TYPES=* cargo build -p jco`

The only strange bit I haven't been able to figure out is that seem to need a  binary being built in the project root, even if it's entirely a no-op. I've removed both the `build.rs` and `main.rs` files, as well as all dependencies and the project builds. But if I remove the project, we get a linker error. I'm still not sure why, but for now I've just named the binary `dummy` and stripped it from everything else. I think we can fix this in a follow-up, but help on why this might be needed would be very welcome!

PR #179, opened 2023-10-10
Exercise wasmtime conformance tests

### LINKED ISSUE #178: Reuse wasmtime's conformance test suite
In order to validate that our bindings are working as intended, we need to introduce a test suite which exercises those bindings. Ideally there would be a standalone "platform conformance test suite" which could be reused by all host implementations. Unfortunately there is no dedicated project for that yet, but wasmtime *does* have a comprehensive set of test programs we should be able to run and test.

What I propose we do is this: pull in `wasmtime` as a submodule, build their test programs, and then run them against our host implementation. We'll have to write a custom test runner which can compile the test programs, load them into Node.js, and then validate their output. But I don't think that should be too hard.

Having a comprehensive test suite should be a priority for us, as it should allow us to implement the host bindings faster. So I'll make this a priority for me to implement.

yoshuawuyts:
Reuses wasmtime's conformance test suite. Closes #178.

### DIFF
--- a/xtask/src/test.rs
+++ b/xtask/src/test.rs
@@ -1,0 +1,30 @@
+use std::fs;
+
+use xshell::{cmd, Shell};
+
+pub fn run() -> anyhow::Result<()> {
+    let sh = Shell::new()?;
+
+    // Compile wasmtime test programs
+    let guard = sh.push_dir("submodules/wasmtime/crates/test-programs");
+    cmd!(sh, "cargo build --target wasm32-wasi --release").run()?;
+    drop(guard);
+
+    for entry in fs::read_dir("submodules/wasmtime/target/wasm32-wasi/release")? {
+        let entry = entry?;
+
+        // skip all files which don't end with `.wasm`
+        if entry.path().extension().map(|p| p.to_str()).flatten() != Some("wasm") {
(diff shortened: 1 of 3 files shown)


xtask/src/test.rs
```diff
@@ -0,0 +1,36 @@
...
+            Some(path) => path,
+            None => continue,
+        };
+
+        // only iterate over the `preview2` test programs
+        if !path.contains("preview2") {
+            continue;
+        }
+
+        cmd!(sh, "src/jco.js run {path}").run()?;
```
guybedford, line 32: Are all the test programs commands? Or do we need to invoke the specific test instructions against them?

guybedford (approved):

xtask/src/test.rs
yoshuawuyts: Ah, yeah, so I was going to rewrite this. Right now this just runs every command (many of which are still failing) one-by-one. But what I'd like to do instead is refactor this to _generate_ tests, check them in the repo, and then make them runnable via `cargo test`. That way an individual test failure doesn't mark the end.

But that requires adding a test runner crate, which I think I might just quickly write in a separate PR.

Issue #182, opened 2023-10-11
Rely on `wasmtime` as a git submodule for the entire project


yoshuawuyts:
In https://github.com/bytecodealliance/jco/pull/179 I'm adding wasmtime as a submodule, in order to access the `test-programs` crate. However we're already relying on wasmtime for `wasi_preview1_component_adapter.reactor.adapter` and `wasi_preview1_component_adapter.command.adapter`. Rather than doing this via a git clone into a temporary directory, it would probably be better if this could reuse the submodule. This would also take care of managing the `wasi-version` in the git root for us.

I believe all this requires is updating the `update_wasi.sh` script and we should be good. Though `update_preview2.sh` also gets artifacts from wasmtime's release stream, which perhaps we may want to build ourselves from the source instead?

cc/ @guybedford

yoshuawuyts:
Assigning myself here in case we decide we'd like to do this.

guybedford:
I'm 100% agreed on this direction fwiw.

yoshuawuyts:
It feels like we've just about implemented this.

PR #186, opened 2023-10-13
use git submodules for all git dependencies

yoshuawuyts:
This updates the shell scripts to make use of git submodules, closing #182. We're still using an unspecified version of `wasm-tools` to edit the metadata - we should address that in a follow-up PR.

I've also taken the liberty to document the existing shell scripts, in anticipation we'll want to manage them via `cargo xtask` in the future. This should make that step easier.

However locally I'm having tests fail now. I'm not sure why that's happening, but I expect it likely has something to do with either #178 or #174 (or possibly the interaction between both). If it's the interaction between both we should consider enabling GitHub's merge queue to prevent future failures of that nature happening again.


guybedford:
Weird that CI is passing while tests are locally failing. Will look into that.

Otherwise this seems great to merge, thank you.

PR #226, opened 2023-10-24
Generate conformance tests

### LINKED ISSUE #178: Reuse wasmtime's conformance test suite

yoshuawuyts:
Closes #178. This finishes generating conformance tests from Wasmtime's conformance test suite.

The main logic of this PR lives in `xtask/generate/tests.rs` and `crates/test-utils/src/lib.rs` files. These are responsible for generating the tests in `tests/generated/`. This test intentionally invokes `jco new` and `jco run` on every run to exercise both the generation and execution of the bindings. Right now it's not yet integrated in CI, but I can do that in a future PR if we want to - though probably as a non-blocking check.

Right now 8/88 test programs are passing. Running the entire test suite on my machine takes roughly five seconds. I think this will be a helpful tool to start validating our implementation with. Thanks!


guybedford:
This is really fantastic, thank you!

PR #228, opened 2023-10-24
Adding file extension to interface imports

### LINKED ISSUE #223: `--map` not working as expected for type imports from `./interfaces`
I have a project generated by `jco transpile ... --map "./interfaces#*=./interfaces/#*.js"` that looks like:

```sh
.
├── wit_ts_generator_component.core.wasm
├── wit_ts_generator_component.d.ts
├── wit_ts_generator_component.js
└── interfaces
    └── ts-generator-generator.d.ts
```

In `wit_ts_generator_component.d.ts` is the line:
```ts
import { TsGeneratorGenerator } from './interfaces/ts-generator-generator';
```

Which causes my typescript build to fail because I am using `--moduleResolution nodenext` and that requires adding the file extension `.js` at the end of the import path. I would have expected the `--map "./interfaces/*=./interfaces/#*.js"` option to remap that `import` statement to work correctly, but it remains unchanged. 

It is likely that I am missing something about how to use `--map`, but it seems like it only works for imports to a `world`? I can add `"skipLibCheck": true` to my `tsconfig` to get around the issue for now, but skipping the extra type checking isn't ideal long term (and I think would not work for consumers of this package).

landonxjames:
Adding the file extension will allow the interface imports to work for typescript packages using both `moduleResolution` `bundler` as well as `nodenext`. This is also more closely aligned with the node spec for imports so should be better supported. Closes #223

### DIFF
--- a/crates/js-component-bindgen/src/ts_bindgen.rs
+++ b/crates/js-component-bindgen/src/ts_bindgen.rs
@@ -860,7 +860,7 @@ impl<'a> TsInterface<'a> {
             Some(owned_interface_name) => {
                 uwriteln!(
                     self.src,
-                    "import type {{ {type_name} }} from '../interfaces/{owned_interface_name}';",
+                    "import type {{ {type_name} }} from '../interfaces/{owned_interface_name}.js';",
                 );
                 self.src.push_str(&format!("export {{ {} }};\n", type_name));
             }


manekinekko:
LGTM

guybedford:
Thanks for the follow-up here.

PR #238, opened 2023-10-27
use name of dealiased resource type when lifting/lowering handles

dicej:
This fixes incorrect guest code generation in cases of aliased resources.


guybedford (approved):
Thanks for the fix!

guybedford:
Does this resolve the issue in https://github.com/bytecodealliance/componentize-js/pull/67?

dicej:
> Does this resolve the issue in [bytecodealliance/componentize-js#67](https://github.com/bytecodealliance/componentize-js/pull/67)?

Yes!  Obviously `componentize-js` will need to be updated to use a Jco version with this patch.

PR #242, opened 2023-11-02
make `resourceHandleSymbol` field configurable

dicej:
This helps to enable precise disposal of imported resource handles per https://github.com/bytecodealliance/componentize-js/issues/68.  Making the `resourceHandleSymbol` field configurable allows it to be deleted from resource objects, which we will do during disposal.


dicej:
Per https://github.com/bytecodealliance/componentize-js/pull/69#discussion_r1380597503, this is not needed after all.

Issue #206, opened 2023-10-17
Jco transpile: implementation of wasi-io for Node.js


yoshuawuyts:
## Tasks

- [x] `streams.read`
- [x] `streams.blocking-read`
- [ ] `streams.skip`
- [ ] `streams.blocking-skip`
- [ ] `streams.subscribe-to-input-stream`
- [x] `streams.drop-input-stream`
- [x] `streams.check-write`
- [x] `streams.write`
- [x] `streams.blocking-write-and-flush`
- [x] `streams.flush`
- [x] `streams.blocking-flush`
- [ ] `streams.subscribe-to-output-stream`
- [ ] `streams.write-zeroes`
- [ ] `streams.splice`
- [ ] `streams.blocking-splice`
- [ ] `streams.forward`
- [x] `streams.drop-output-stream`
- [ ] `poll.drop-pollable`
- [ ] `poll.poll-oneoff`

guybedford:
Full implementation in https://github.com/bytecodealliance/jco/pull/244.

yoshuawuyts:
These should all be closed by #244, the only missing part is adding tests.

PR #244, opened 2023-11-05
Node.js IO & Poll implementation

guybedford:
This refactors the Node.js implementation to support full IO and Poll integration against Node.js streams on a worker thread.

Clocks, filesystem and http have all been updated in the process.

Resolves https://github.com/bytecodealliance/jco/issues/206.


packages/preview2-shim/lib/io/calls.js
```diff
@@ -0,0 +1,37 @@
...
+export const OUTPUT_STREAM_SUBSCRIBE = ++call_id << CALL_SHIFT;
+export const OUTPUT_STREAM_DROP = ++call_id << CALL_SHIFT;
+
+// Io Poll
+export const POLL_POLLABLE_READY = ++call_id << CALL_SHIFT;
+export const POLL_POLLABLE_BLOCK = ++call_id << CALL_SHIFT;
+export const POLL_POLL_LIST = ++call_id << CALL_SHIFT;
+
+// Http
+export const HTTP_CREATE_REQUEST = ++call_id << 24;
```
manekinekko, line 37: Out of curiosity, why are we using bit masks?

packages/preview2-shim/lib/io/worker-io.js
```diff
@@ -0,0 +1,205 @@
...
+    if (e.tag !== 'stream-error') {
+      console.error(e);
+      process.exit(1);
+    }
+    if (e.val.tag === 'last-operation-failed') {
+      const msg = e.val.val.message;
+      e.val.val = Object.assign(new IoError(), e.val);
+      throw new ComponentError("StreamError", msg, e);
+    }
+    throw new ComponentError("StreamError", "stream closed", e);
```
manekinekko, line 55: It would be interesting to provide a single implementation for error handling that we could use across all code. wdyt?

packages/preview2-shim/lib/io/worker-io.js
```diff
@@ -0,0 +1,205 @@
...
+  static _listToIds (list) {
+    return list.map(pollable => pollable.#id);
+  }
+  static _markReady (pollable) {
+    pollable.#ready = true;
+  }
+}
+
+const pollableCreate = Pollable._create;
+delete Pollable._create;
```
manekinekko, line 185: > `delete Pollable._create;`
I've seen this pattern across different files. Can you provide a short explanation (for my learning purpose)?

packages/preview2-shim/lib/io/worker-thread.js
```diff
@@ -0,0 +1,312 @@
...
+ */
+function streamError (streamId, stream, err) {
+  if (stream.end)
+    stream.end();
+  // we delete the stream from unfinishedStreams as it is now "finished" (closed)
+  unfinishedStreams.delete(streamId);
+  return {
+    tag: 'stream-error',
+    val: { tag: 'last-operation-failed', val: err }
+  };
```
manekinekko, line 27: same comment as my other note about error handing.

packages/preview2-shim/lib/io/worker-thread.js
```diff
@@ -0,0 +1,312 @@
...
+          unfinishedPolls.delete(pollCnt);
+          throw streamError(id, stream.stream, err);
+        }));
+        return pollCnt;
+      case calls.INPUT_STREAM_DROP:
+        unfinishedStreams.delete(id);
+        return;
+
+      case calls.OUTPUT_STREAM_CHECK_WRITE:
+        var { stream } = getStreamOrThrow(id);
```
manekinekko, line 142: I've noticed `var` is being used a lot here, we could limit the scope of declared variables to the current lexical env using `const` or `let` to avoid any potential hoisting issues in the future - unless there is a valid reason I am not aware of.

```suggestion
        const { stream } = getStreamOrThrow(id);
```

packages/preview2-shim/lib/io/worker-thread.js
```diff
@@ -0,0 +1,312 @@
...
+    unfinishedStreams.delete(streamId);
+    throw { tag: 'closed' };
+  }
+  return stream;
+}
+
+/**
+ * @param {number} call
+ * @param {number | null} id
+ * @param {any} payload
```
manekinekko, line 52: what's the type of `payload`?

packages/preview2-shim/lib/io/worker-thread.js
```diff
@@ -0,0 +1,312 @@
...
+        // TODO: these error handlers should be attached, and only for the duration of the splice flush
+        if (inputStream.errored) {
+          streamError(payload.src, inputStream, inputStream.errored); // error ignored?
+          throw streamError(id, outputStream, inputStream.errored);
+        }
+        if (outputStream.errored) {
+          // input stream not closed?
+          throw streamError(id, outputStream, outputStream.errored);
+        }
+        return payload.len - BigInt(bytesRemaining);
```
manekinekko, line 214: is `payload.len` always a positive integer? If it's 0 then this result would be negative. Is that expected?

test/fixtures/wit/deps/sockets/udp.wit
```diff
@@ -49,95 +64,47 @@ interface udp {
...
-        /// - <https://pubs.opengroup.org/onlinepubs/9699919799/functions/sendto.html>
-        /// - <https://pubs.opengroup.org/onlinepubs/9699919799/functions/sendmsg.html>
-        /// - <https://man7.org/linux/man-pages/man2/send.2.html>
-        /// - <https://man7.org/linux/man-pages/man2/sendmmsg.2.html>
-        /// - <https://learn.microsoft.com/en-us/windows/win32/api/winsock2/nf-winsock2-send>
-        /// - <https://learn.microsoft.com/en-us/windows/win32/api/winsock2/nf-winsock2-sendto>
-        /// - <https://learn.microsoft.com/en-us/windows/win32/api/winsock2/nf-winsock2-wsasendmsg>
-        /// - <https://man.freebsd.org/cgi/man.cgi?query=send&sektion=2>
-        send: func(datagrams: list<datagram>) -> result<u64, error-code>;
+        %stream: func(remote-address: option<ip-socket-address>) -> result<tuple<incoming-datagram-stream, outgoing-datagram-stream>, error-code>;
```
manekinekko, line 107: is `%` a typo?

packages/preview2-shim/types/interfaces/wasi-sockets-udp.d.ts
```diff
@@ -29,193 +27,276 @@ export namespace WasiSocketsUdp {
...
+      send(datagrams: OutgoingDatagram[]): bigint;
+      subscribe(): Pollable;
+    }
+    
+    export class IncomingDatagramStream {
+      receive(maxResults: bigint): IncomingDatagram[];
+      subscribe(): Pollable;
+    }
+    
+    export class UdpSocket {
```
manekinekko, line 285: are me missing the `stream()` defined in the wit file? See https://github.com/bytecodealliance/jco/pull/244/files#r1383077625

manekinekko:
Thank you @guybedford for adding support for Poll and Streams. I will wait for them to land in main, so I can rebase the wasi-socket branch.

packages/preview2-shim/lib/io/calls.js
guybedford: Having each call assigned a single numeric ID helps. The masking is used so that eg for stream calls, we can have a single part of the call that says "this is a stream flush" and another part of the number that says which stream type it is (see stream-types). From an implementation perspective, we can then implement a generic stream flush based on the first mask, or we can implement stream-specific implementations based on the combination.

packages/preview2-shim/lib/io/worker-io.js
guybedford: `Pollable` is exposed publicly to users, but users can't create their own `Pollable`s. So the create function is a private function, hence we delete it to not give users access. The reason it has to exist as a class method is because it accesses the private fields.

packages/preview2-shim/lib/io/worker-io.js
guybedford: I would be open to any refactoring here, still kind of figuring it out myself!

packages/preview2-shim/lib/io/worker-thread.js
guybedford: In this case, this function is the generic error handler for streams.

packages/preview2-shim/lib/io/worker-thread.js
guybedford: The reason I'm using vars in the switch statement is because I don't want to create a new block statement for every single case. We could add block statements though for sure.

packages/preview2-shim/lib/io/worker-thread.js
guybedford: Because this function represents every possible call, payload will be whatever is needed for the specific call. It is therefore very specifically `any`.

packages/preview2-shim/lib/io/worker-thread.js
guybedford: In the component model itself, this is typed as a u64, so it's not possible to be negative.

For components that call to WASI directly, this is therefore always guaranteed.

guybedford: Secondly, because `bytesRemaining` is originally set to `payload.len`, `bytesRemaining` can never be > 0 when `payload.len` is `0` here.

packages/preview2-shim/lib/io/worker-thread.js
guybedford: I've gone ahead and updating to using blocks.

packages/preview2-shim/lib/io/worker-thread.js
```diff
@@ -0,0 +1,312 @@
...
+        if (!stream || stream.closed || stream.errored || !stream.writableNeedDrain)
+          return 0;
+        var resolve, reject;
+        unfinishedPolls.set(++pollCnt, new Promise((_resolve, _reject) => {
+          stream
+            .once('drain', resolve = _resolve)
+            .once('error', reject = _reject);
+        }).then(() => {
+          unfinishedPolls.delete(pollCnt);
+          stream.off('error', reject);
```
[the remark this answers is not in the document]
guybedford: We aren't holding onto the `stream` variable here, so it should then just get GC'd after the end of this call I believe. We could explicitly null the variable to potentially more eagerly handle that GC operation, although I'm not sure we strictly need to?

packages/preview2-shim/types/interfaces/wasi-sockets-udp.d.ts
guybedford: This PR also updates to the Wasmtime master branch of the wits (via `cd submodules/wasmtime && git fetch origin && git reset --hard origin/main` and then `xtask update preview2` and then `xtask generate wasi-types`).

guybedford: At least on Wasmtime main, this method doesn't exist.

guybedford:
I'm going to focus on fixing the browser implementation for now, in a new jco patch, before landing this as experimental either later today or tomorrow.

packages/preview2-shim/lib/io/worker-io.js
manekinekko: This is not urgent and can be done later when we finish laying the foundation for the other implementations.

packages/preview2-shim/lib/io/worker-thread.js
manekinekko: Got it. Thanks.

packages/preview2-shim/lib/io/worker-thread.js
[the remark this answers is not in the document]
manekinekko: Not sure how GC will behave in this case with streams. Let's merge this as is for now.

packages/preview2-shim/lib/io/worker-thread.js
```diff
@@ -0,0 +1,328 @@
...
+        let bytesRemaining = Number(payload.len);
+        let chunk;
+        while (bytesRemaining > 0 && (chunk = inputStream.read(Math.min(outputStream.writableHighWaterMark - outputStream.writableLength, bytesRemaining)))) {
+          bytesRemaining -= chunk.byteLength;
+          outputStream.write(chunk);
+        }
+        // TODO: these error handlers should be attached, and only for the duration of the splice flush
+        if (inputStream.errored) {
+          streamError(payload.src, inputStream, inputStream.errored); // error ignored?
+          throw streamError(id, outputStream, inputStream.errored);
```
guybedford, line 215: Input stream errors should not throw the output stream.

Issue #246, opened 2023-11-06
JavaScript env for browsers is broken


temeddix:
Hi, I've been testing `jco` on NodeJS and it has been working well. Great project!

However, I couldn't get it to compile JavaScript files for the browser(without `import` and `export`s). How can I do that? I dug in the docs and guides, but no explanation was there that I could find. Do I have to use Webpack or Vite to transpile ES JavaScript to Browser JavaScript?

temeddix:
When I run `npx webpack`, the following error occurs:

```
PS C:\Users\temed\Documents\GitHub\jco-test> npx webpack
assets by status 2.54 MiB [cached] 2 assets
asset main.js 98 KiB [emitted] (name: main)
runtime modules 4.23 KiB 7 modules
cacheable modules 77.9 KiB (javascript) 2.54 MiB (asset)
  javascript modules 77.9 KiB
    modules by path ./node_modules/@bytecodealliance/preview2-shim/lib/ 15.1 KiB
      ./node_modules/@bytecodealliance/preview2-shim/lib/browser/cli.js 1.25 KiB [built] [code generated]
      ./node_modules/@bytecodealliance/preview2-shim/lib/browser/filesystem.js 8.24 KiB [built] [code generated]
      ./node_modules/@bytecodealliance/preview2-shim/lib/browser/clocks.js 1010 bytes [built] [code generated]
      ./node_modules/@bytecodealliance/preview2-shim/lib/common/io.js 4.62 KiB [built] [code generated]
    ./browser.js 94 bytes [built] [code generated]
    ./out_dir/jco_test.js 62.7 KiB [built] [code generated]
  asset modules 84 bytes (javascript) 2.54 MiB (asset)
    ./out_dir/jco_test.core.wasm 42 bytes (javascript) 2.51 MiB (asset) [built] [code generated]
    ./out_dir/jco_test.core2.wasm 42 bytes (javascript) 28.7 KiB (asset) [built] [code generated]

ERROR in ./node_modules/@bytecodealliance/preview2-shim/lib/browser/filesystem.js 1:0-30
Module not found: Error: Can't resolve './io.js' in 'C:\Users\temed\Documents\GitHub\jco-test\node_modules\@bytecodealliance\preview2-shim\lib\browser'
resolve './io.js' in 'C:\Users\temed\Documents\GitHub\jco-test\node_modules\@bytecodealliance\preview2-shim\lib\browser'
[...]

PR #252, opened 2023-11-07
fix: browser shim

guybedford:
Fixes the browser shim which broke during recent IO upgrades. Resolves https://github.com/bytecodealliance/jco/issues/246.


guybedford:
I'm going to fast-track and release this shortly.

Issue #245, opened 2023-11-06
`--instantiation` with non-async support


Hywan:
Hello,

Thanks for the great project, it's really useful!
I'm using it in particular context where I need to use `--instantation`, thanks for providing this option! My problem is that the generated `instantiate` function is async:

```typescript
function instantiate(
    compileCore: (…) => Promise<WebAssembly.Module>,
    imports: ImportObject,
    instantiateCore?: (…) => Promise<WebAssembly.Instance>,
): Promise<…>
```

I need a non-async `instantiate` function, such as:

```typescript
function instantiate(
    compileCore: (…) => WebAssembly.Module,
    imports: ImportObject,
    instantiateCore?: (…) => WebAssembly.Instance,
): …
```

I can work on this patch, but I would like to align on a couple of things beforehand:

1. Are you willing to merge such a feature?
2. Ideally, we want to keep `--instantiation`, but I would “duplicate” the option with `--instantation-no-async`. Internally, we would keep the `TranspileOpts::instantiation` field, but a new `TranspileOpts::instantiation_is_async` field would be added:
    * `--instantiation` would result in `TranspileOpts { instantiation: true, instantiation_is_async: true }`
    * `--instantiation-no-async` would result in `TranspileOpts { instantiation: true, instantiation_is_async: false }`.

One could argue that I could `str::replace` the output of JCO to replace `async function` by `function`, `await` by nothing, `Promise.all…` by nothing and so on, but I believe this problem might appear for other users, and having this implemented inside JCO directly would be less hacky.

Thoughts?

yoshuawuyts:
@Hywan can you elaborate why you want this function to be non-async? It would be helpful to know more about the use case motivating this.

Hywan:
I'm running the JCO output with JavaScriptCore from [a Rust bindings crate.](github.com/endoli/javascriptcore.rs/) Manipulating promises/async code in this context is really not trivial. I even wonder if that's possible with the C API of JavaScriptCore. The problem is that it's not possible to know when a `Promise` has resolved easily. Technically, I would have callbacks from JavaScript to Rust, and create functions that call this callbacks when a `Promise` is resolved, but that's clearly more complex, esp. on the Rust side (as it doesn't fit in a `Future` very well). So we have async code on the JavaScript side, and sync code on the Rust side. The only way to implement this is with a loop checking if we get a value from JavaScript. Or… just having non-async JavaScript code :-).

[...]

PR #254, opened 2023-11-08
feat: Allow `--instantiation` to generate non-async code

Hywan:
This code should ideally be reviewed patch by patch.

The idea is to implement the feature described in https://github.com/bytecodealliance/jco/issues/245.

First off, this patch updates the `--instantiation` option to receive a string value instead of a being a switch. The possible values are `async` (the default) or `sync`. The fact that `async` is the default value ensures the backward compatibility.

Next, this patch updates the `TranspileOpts::instantiation` field to be `Option<InstantiationMode>` instead of `bool`. `InstantiationMode` is a enum with 2 variants: `Async` and `Sync`. The code is updated accordingly.

Next, the same patch adds non-async code during the bindgen pass.

The next patches are about improving the test framework, and adds a new test case for non-async generated code.

### DIFF
--- a/crates/js-component-bindgen/src/ts_bindgen.rs
+++ b/crates/js-component-bindgen/src/ts_bindgen.rs
@@ -228,26 +228,26 @@ pub fn ts_bindgen(
                 bindgen.src,
                 "
                     /**
-                    * Instantiates this component with the provided imports and
-                    * returns a map of all the exports of the component.
-                    *
-                    * This function is intended to be similar to the
-                    * `WebAssembly.instantiate` function. The second `imports`
-                    * argument is the \"import object\" for wasm, except here it
-                    * uses component-model-layer types instead of core wasm
-                    * integers/numbers/etc.
-                    *
-                    * The first argument to this function, `compileCore`, is
-                    * used to compile core wasm modules within the component.
-                    * Components are composed of core wasm modules and this callback
-                    * will be invoked per core wasm module. The caller of this
-                    * function is responsible for reading the core wasm module
-                    * identified by `path` and returning its compiled
-                    * `WebAssembly.Module` object. This would use `compileStreaming`
-                    * on the web, for example.
-                    */
+                     * Instantiates this component with the provided imports and
+                     * returns a map of all the exports of the component.
+                     *
+                     * This function is intended to be similar to the
+                     * `WebAssembly.instantiate` function. The second `imports`
+                     * argument is the \"import object\" for wasm, except here it
+                     * uses component-model-layer types instead of core wasm
+                     * integers/numbers/etc.
+                     *
+                     * The first argument to this function, `getCoreModule`, is
+                     * used to compile core wasm modules within the component.
+                     * Components are composed of core wasm modules and this callback
+                     * will be invoked per core wasm module. The caller of this
+                     * function is responsible for reading the core wasm module
+                     * identified by `path` and returning its compiled
+                     * `WebAssembly.Module` object. This would use `compileStreaming`
+                     * on the web, for example.
+                     */
                     export function instantiate(
-                        compileCore: (path: string, imports: Record<string, any>) => Promise<WebAssembly.Module>,
+                        getCoreModule: (path: string) => Promise<WebAssembly.Module>,
                         imports: ImportObject,
                         instantiateCore?: (module: WebAssembly.Module, imports: Record<string, any>) => Promise<WebAssembly.Instance>
                     ): Promise<{camel}>;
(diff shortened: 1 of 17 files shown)


crates/js-component-bindgen/src/ts_bindgen.rs
```diff
@@ -217,36 +222,72 @@ pub fn ts_bindgen(
...
+                    * used to compile core wasm modules within the component.
+                    * Components are composed of core wasm modules and this callback
+                    * will be invoked per core wasm module. The caller of this
+                    * function is responsible for reading the core wasm module
+                    * identified by `path` and returning its compiled
+                    * `WebAssembly.Module` object. This would use the
+                    * `WebAssembly.Module` constructor on the web, for example.
+                    */
+                    export function instantiate(
+                        compileCore: (path: string, imports: Record<string, any>) => WebAssembly.Module,
```
guybedford, line 282: In the case of a sync instantiation, we should assume that the modules are already compiled I think, so let's rename this to something like `getCoreModule`?

crates/js-component-bindgen/src/ts_bindgen.rs
```diff
@@ -217,36 +222,72 @@ pub fn ts_bindgen(
...
+                    * used to compile core wasm modules within the component.
+                    * Components are composed of core wasm modules and this callback
+                    * will be invoked per core wasm module. The caller of this
+                    * function is responsible for reading the core wasm module
+                    * identified by `path` and returning its compiled
+                    * `WebAssembly.Module` object. This would use `compileStreaming`
+                    * on the web, for example.
+                    */
+                    export function instantiate(
+                        compileCore: (path: string, imports: Record<string, any>) => Promise<WebAssembly.Module>,
```
guybedford, line 250: It might then make sense to do the same renaming here too.

crates/js-component-bindgen/src/ts_bindgen.rs
guybedford: Also now that I'm looking at this, unrelatedly we should probably remove `imports` as a parameter to `compileCore`, as that shouldn't actually be a thing. I can post a follow-up PR if you'd rather do it that way though.

guybedford (commented):
Thanks for the PR, feedback mainly just the one point on naming, otherwise it looks good to go to me.

guybedford (commented):
Thanks for this work, the approach looks great to me.

Just a couple of comments about the naming of `compileCore` under this model.

crates/js-component-bindgen/src/ts_bindgen.rs
Hywan: Done.

crates/js-component-bindgen/src/ts_bindgen.rs
Hywan: Both are done :-).

Hywan: The `imports` argument of the `compileCore` function (now `getCoreModule`) was never used in the tests, only the `imports` argument of `instantiate` was used. It's theoretically a breaking change, but no one should notice it :-/.

crates/js-component-bindgen/src/ts_bindgen.rs
```diff
@@ -217,36 +222,72 @@ pub fn ts_bindgen(
...
+                     * integers/numbers/etc.
+                     *
+                     * The first argument to this function, `getCoreModule`, is
+                     * used to compile core wasm modules within the component.
+                     * Components are composed of core wasm modules and this callback
+                     * will be invoked per core wasm module. The caller of this
+                     * function is responsible for reading the core wasm module
+                     * identified by `path` and returning its compiled
+                     * `WebAssembly.Module` object. This would use the
+                     * `WebAssembly.Module` constructor on the web, for example.
```
guybedford, line 279: In the case of synchronous compilation, `getCoreModule` would have to be assumed to be cached. So perhaps we should update the comment to something like:

```suggestion
                     * The first argument to this function, `getCoreModule`, is
                     * used to get the compiled core wasm modules within the component.
                     * Components are composed of core wasm modules and this callback
                     * will be invoked per core wasm module. The caller of this
                     * function is responsible for reading the core wasm module
                     * identified by `path` and returning its compiled
                     * `WebAssembly.Module` object. This should ideally be an 
                     * already-cached cached compilation instead of created by the 
                     * `WebAssembly.Module` constructor on the web.
```

guybedford: I'll land it with this rewording soon, unless you have any further feedback.

Hywan: I'm fine with the rewording :-). Thanks!

Hywan:
:tada: thanks!

PR #255, opened 2023-11-09
chore: Make Clippy happier

Hywan:
`cargo clippy` wasn't happy. Now it's happier.


guybedford:
As a general rule, supporting clippy is not a goal for this project, so that this is not something we should ever test for or aim to "fix". That said, looking through the changes individually, I'm happy to land the suggestions as a code improvement / refactoring, without committing to accepting any further PRs in the same vein :)

Hywan:
Thanks :-)

Issue #253, opened 2023-11-07
Provide Typescript type declarations for `jco` API


landonxjames:
I am attempting to use `jco` as a library from a Typescript project. The following import throws an error:
```
import { transpile } from "@bytecodealliance/jco";
```
Throws: 
```
Could not find a declaration file for module '@bytecodealliance/jco'. '/Volumes/workplace/project/node_modules/@bytecodealliance/jco/src/api.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/bytecodealliance__jco` if it exists or add a new declaration (.d.ts) file containing `declare module '@bytecodealliance/jco';`ts(7016)
```

This appears to be because there are no `.d.ts` files in `node_modules/@bytecodealliance/jco`, so ts doesn't recognize the package. I can add my own declaration file to get unblocked, but this seems like a thing that will impact multiple users going forward.

From some quick research it seems like this could be handled in a few ways:
* Change `jco` to a typescript project, run `tsc` as part of the build to generate the `.d.ts` bindings
* Manually write ts bindings and distribute them with the package
* Manually write the bindings and contribute them to [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped)

The first option looks like the best, but would be fairly disruptive. The other options might be quicker.

landonxjames:
Happy to contribute to some work here, but would like feedback from project maintainers on what direction is preferable before I dive too far in.

guybedford:
Thanks for bringing this up, we should definitely resolve it!

The file in `src/api.js` directly reexports from files that are all themselves typed, so it should be possible to achieve this with just an `api.d.ts` or jsdoc comments that reexport from the same interfaces.

So it should be a relatively PR to fix, and we shouldn't need to maintain much more than just those reexport declarations.

If you're able to look into this further that would be a huge help.

landonxjames:
Almost done with this, but I'm encountering an error that I've encountered in my own projects as well. 

```js
export { tools, tools as 'local:wasm-tools/tools' }
```

This is valid javascript, but (seemingly) not valid typescript. Ts accepts `as fooBar`, but not `as "fooBar"`.  This is probably a ts bug since it should mirror the js implementation.

In my own projects I've been handling this by `sed`ing in a replacement before running `tsc`, but that feels a but hacky for the project that generates those export statements. It should work here though as nothing appears to be using that alias.

[...]

Issue #168, opened 2023-09-30
Generated export seems invalid


zifeo:
The following [code](https://github.com/bytecodealliance/jco/blob/main/crates/js-component-bindgen/src/esm_bindgen.rs#L150-L186) generate some likely invalid values:
```
export {
  core,
  runtimes,
  utils,
  core as "metatype:typegraph/core",
  runtimes as "metatype:typegraph/runtimes",
  utils as "metatype:typegraph/utils",
};
```
Same repro as https://github.com/bytecodealliance/jco/issues/102.

guybedford:
This is correct - these are exported interface names. In JavaScript it is valid to export arbitrary strings as exports.

zifeo:
@guybedford I suppose this is again a deno compat limitation. What is the exact uses of those?

guybedford:
@zifeo it's effectively the distinction between a registry name and a module export, where components can export registry-named pieces of code that can be used to globally define interfaces or implementations. The TypeScript issue to track is https://github.com/microsoft/TypeScript/issues/40594.

We could potentially have an option to customize this and in the Deno case, to disable this feature just like we do in the TypeScript generation right now.

zifeo:
@guybedford Thanks a lot!

PR #256, opened 2023-11-13
Updating build process to create `.d.ts` declaration files 

landonxjames:
Updating build process to create `.d.ts` declaration files so `jco` is usable as a library in a typescript project. 

The changes to enable this:
* Add a minimal `tsconfig.json` to generate ts types from the JSDoc comments. See the [ts docs](https://www.typescriptlang.org/docs/handbook/declaration-files/dts-from-js.html) for more info about the configuration.
* Update the `package.json` `build` script to run `tsc`
* Update `.gitignore` to ignore the new `.d.ts` and `.d.ts.map` files which are generated in the `src/` directory.
* Add new option to `--no-namespaced-exports` to `jco` to prevent generating exports like `test as "test:flavorful/test"` which are not currently compatible with typescript. See https://github.com/microsoft/TypeScript/issues/40594

Resolves https://github.com/bytecodealliance/jco/issues/168.
Resolves https://github.com/bytecodealliance/jco/issues/253


xtask/src/build/jco.rs
```diff
@@ -14,6 +13,17 @@ pub(crate) fn run() -> Result<()> {
         "wasm-tools".to_string(),
     )?;
 
+    let sh = Shell::new()?;
+    //Typescript doesn't support arbitrary module namespace identifier names so
+    //we need to remove it from wasm-tools.js before compiling
+    //See: https://github.com/microsoft/TypeScript/issues/40594
+    cmd!(
+        sh,
+        "sed -i'' -e 's/tools as \"local:wasm-tools\\/tools\"//g' ./obj/wasm-tools.js"
```
guybedford, line 22: Rather than doing this via replacement (since if jco has this issue itself then other libraries built with jco will too), perhaps we should instead make it a transpile option to omit these names?

Perhaps something like `--ts-compat` or `--no-namespaced-exports`. Would you be interested in working on a patch along those lines here?

landonxjames: Yep happy to take a look at that (and fix this windows build issue). Will probably take me a few more days or through the weekend, but any pointers to that section of the code (and potentially how cli options flow from the js to the rust bits) would be appreciated

guybedford: Sure, the exports are output by the "esm bindgen" render routine in https://github.com/bytecodealliance/jco/blob/main/crates/js-component-bindgen/src/esm_bindgen.rs#L184.

landonxjames: Alright updates are made and the `--no-namespaced-exports` option has been added

guybedford:
Great work, thanks for adding this.

Issue #201, opened 2023-10-17
Jco transpile: implementation of wasi-clocks for Node.js


yoshuawuyts:
## Tasks

- [x] `monotonic-clock.now`
- [x] `monotonic-clock.resolution`
- [x] `monotonic-clock.subscribeInstant`
- [x] `monotonic-clock.subscribeDuration`
- [x] `wallclock.now`
- [x] `wallclock.resolution`

## Conformance Tests
- [x] `api_time`
- [x] `cli_default_clocks`
- [x] `preview1_clock_time_get`
- [x] `preview2_sleep`

yoshuawuyts:
I'm currently working on this one.

PR #266, opened 2023-11-17
fix: more conformance tests

guybedford:
This adds 8 new passing tests (I've already crossed them off in the corresponding issues).

* Resolves https://github.com/bytecodealliance/jco/issues/201.
* Resolves most of the CLI tests, with just the `stdin` tests remaining to do which I'll post as a follow-up PR.
* Adds a custom `--jco-import` option to `jco run` to allow for custom environment virtualization which should work as a pattern to be able to ensure the right environment for testing.
* Adds a new `--jco-debug` option to `jco run` which doesn't delete the temporary files (so they can be debugged) and also outputs the trace calls.
* Resolves a resources bug I hit on one of the cases, where the resource table wasn't being initialized.
* Updates the utf8 encoder to avoid reallocations, since these are unsupported in the test binaries.

_Going to start adding this to my PRs:_ **Progress: 27/98 tests**

### DIFF
--- a/crates/js-component-bindgen/src/function_bindgen.rs
+++ b/crates/js-component-bindgen/src/function_bindgen.rs
@@ -5,6 +5,7 @@ use heck::*;
 use std::collections::{BTreeMap, BTreeSet};
 use std::fmt::Write;
 use std::mem;
+use wasmtime_environ::component::TypeResourceTableIndex;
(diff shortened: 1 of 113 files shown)


manekinekko:
> Adds a new --jco-debug option to jco run which doesn't delete the temporary files (so they can be debugged) and also outputs the trace calls.

Thank you!! I love this. This will same be a LOT of time :D

PR #270, opened 2023-11-23
doc: Fix API documentation of `transpileComponent`

Hywan:
The API documentation of `transpileComponent` was incorrect, leading to false-negative warnings in the code when an appropriated IDE is used.

This patch corrects the API documentation.


guybedford:
Thanks!

PR #272, opened 2023-11-23
fix(tests): fix eacces in WSL when creating virtualenvs

manekinekko:
There was a missing OS separator `sep` in this line that failing with an `EACCESS` error when running tests on WSL:
 https://github.com/bytecodealliance/jco/blob/df1f3005cfcb3ea9fb7c89b8e756c18c53b65abc/tests/virtualenvs/base.js#L8

image


guybedford:
Nice, thank you. Also need to fix a flake on the main Windows CI, will get to that soon too.

Issue #269, opened 2023-11-23
`--js` with `--instantiation` is kind of broken


Hywan:
Hello,

I've noticed that `--js` combined with `--instantiation` is broken:

* `import` statements are present (while they should be removed with `--instantiation`),
* `const instantiateCore = WebAssembly.Instance` is present, but `--js` implies that `WebAssembly` isn't defined,
* `--instantiation=sync` generates a non-async `instantiate` function, but `$init` is still async (that's my fault because I've introduced `--instantiation=sync` in #254)
* `$init` uses `import` statements:

    ```js
    const $init = (async () => {
      let idx = 0;
      ({ greet } = await instantiate(n => idx++, {
        'foo:bar/baz': import0,
      }, (i, imports) => ({ exports: asmInit[i](imports) })));
    })();
    export { greet }

    await $init;
    ```

    where `import0` is defined as:

    ```js
    import * as import0 from 'foo:bar/baz';
    ```

I can fix that if you wish, it's important for me. I believe we need to clarify the behaviour of `--js`. As far as I understand it, it can be present without `--instantiation` right? But if `--instantiation` is missing and `--js` is present, then it implies `--instantiation=async`:

https://github.com/bytecodealliance/jco/blob/df1f3005cfcb3ea9fb7c89b8e756c18c53b65abc/src/cmd/transpile.js#L109-L113

So `--instantiation` does not imply `--js`, but `--js` always imply `--instantiation`. Alright. However, `--js` should check the instantiation mode, either `async` or `sync` in order to generate a valid `$init` function.

But if `--js` always imply `--instantiation`, why do we have an `import` statement then?

#### Proposal 1

Maybe we should revamp `--js`? Actually, let's remove it…!

1. `--instantiation` would have 3 values: `async`, `sync` and `js` (🆕)
2. `--instantiation=js` will generate a non-async `instantiate` function:
    * No `compileCore` argument,
    * No `instantiateCore` argument,
    * Just a single `imports` import object argument.
3. Because this is an `--instantiation`, we don't have `import` statements.

I know it's a breaking change. But `--js` is broken as is, even with `--instantiation[=async] --js` today.

#### Proposal 2

[...]

PR #273, opened 2023-11-23
Use `--js` with `--instantiation`

Hywan:
This patch aims at supporting `--js` with `--instantiation`.

`--js` needs an `instantiate` function to work, so it might look like
`--instantiation` is always implied, but actually no. It is correct
that when `--js` is present, an `instantiate` function _is_ generated,
but it doesn't mean that we expect the function to be used, it's simply
not exported, plus `instantiate` is automatically called (if `--tla-compat`
is `false`). When `--instantiation` is missing, functions are exported
with the `export` directive, and imports are imported with the `import`
directive. When `--instantiation` is present, there is no `export` and no
`import`: only a single exported `instantiate` function.

Basically, we get this:

* `--js` only:
  * `instantiate` is renamed to `_instantiate`,
  * A new `instantiate` function is created, that calls `_instantiate` with
    the correct imports (which are ASM.js code) and returns the exports,
  * A new `$init` function is created, that calls `instantiate` and maps
    the returned exports to their respective trampolines,
  * Trampolines are exported,
  * `$init` is called automatically.
  * ```js
    import * as import0 from '../flavorful.js';
    import * as import1 from '../helpers.js';

    // …

    // The “old” `instantiate` function.
    async function _instantiate() { … }

    // The “export trampolines”.
    let _testImports;

    function testImports() {
      return _testImports.apply(this, arguments);
    }

    // The new `instantiate` function (not exported).
    async function instantiate(imports) {
      const wasm_file_to_asm_index = {
        'flavorful.core.wasm': 0,
        'flavorful.core2.wasm': 1,
        'flavorful.core3.wasm': 2,
        'flavorful.core4.wasm': 3
      };

      return await _instantiate(
          module_name => wasm_file_to_asm_index[module_name],
          imports,
          (module_index, imports) => ({ exports: asmInit[module_index](imports) })
      );
    }

    // Exporting the trampolines.
    export {
      test,
      test_flavorful_test as 'test:flavorful/test',
      testImports,
    }
     
    // The `$init` function.
    async function $init() {
      ( {
        test,
        'test:flavorful/test': test_flavorful_test,
        'testImports': _testImports,
      } = await instantiate(
        {
          '../flavorful.js': import0,
          '../helpers.js': import1,
        }
      ) )
    }
     
    // Calling `$init`.
    await $init();
    ```

* `--js` with `--tla-compat`:
  * Same as with `--js` only, except that `$init` is exported instead of
    being called immediately.
  * ```js
    // …

    export {
      // Because of `--tla-compat`, `$init` is exported!
      $init,
      test,
      test_flavorful_test as 'test:flavorful/test',
      testImports,
    }

    async function $init() {
      ( {
        test,
        'test:flavorful/test': test_flavorful_test,
        'testImports': _testImports,
      } = await instantiate(
        {
          '../flavorful.js': import0,
          '../helpers.js': import1,
        }
      ) )
    }

    // `$init` isn't called.
    ```

* `--js` with `--instantiation[=async]`:
  * `instantiate` is renamed to `_instantiate`,
  * A new `instantiate` function is created, that calls `_instantiate` with
    the correct imports (which are ASM.js code) and returns the exports,
  * `instantiate` is exported.

* `--js` with `--instantiation=sync`:
  * Same as `--js` with `--instantiation[=async]`, except that
    `_instantiate` and `instantiate` are non-async:
  * ```js
    // …

    // The “old” `instantiate` function.
    function _instantiate(…) { … }

    // The new `instantiate` function.
    export function instantiate(imports) {
      const wasm_file_to_asm_index = {
        'strings.sync+js.core.wasm': 0,
        'strings.sync+js.core2.wasm': 1,
        'strings.sync+js.core3.wasm': 2,
        'strings.sync+js.core4.wasm': 3
      };

      return _instantiate(
          module_name => wasm_file_to_asm_index[module_name],
          imports,
          (module_index, imports) => ({ exports: asmInit[module_index](imports) })
      );
    }
    ```

This patch adds the relevant tests too. The code has been revamped a little
bit for the sake of clarity. Hope you like it.

Also please, note the new `wasm_file_to_asm_index` variable inside `instantiate`. Previously, the code was:

```js
    { … } = await instantiate(n => idx++, …, …)
```

Now it is:

```js
    { … } = await instantiate(module_name => wasm_file_to_asm_index[module_name], …, …)
```

The Wasm module name was never taken into account: only the order in which the `getCoreModule` function argument was called was relevant. It was fragile. I prefered to add a layer of safety by having a mapping from Wasm file names to ASM.js function indices.

---

* Fix https://github.com/bytecodealliance/jco/issues/269

### DIFF
--- a/crates/js-component-bindgen/src/intrinsics.rs
+++ b/crates/js-component-bindgen/src/intrinsics.rs
@@ -171,18 +171,10 @@ pub fn render_intrinsics(
                 const i64ToF64 = i => (i64ToF64I[0] = i, i64ToF64F[0]);
             "),
 
-            Intrinsic::InstantiateCore => match instantiation {
-                Some(InstantiationMode::Async) | None =>  {
-                    output.push_str("
-                        const instantiateCore = WebAssembly.instantiate;
-                    ")
-                }
-
-                Some(InstantiationMode::Sync) =>  {
-                    output.push_str("
-                        const instantiateCore = WebAssembly.Instance;
-                    ")
-                }
+            Intrinsic::InstantiateCore => if instantiation.is_none() {
+                output.push_str("
+                    const instantiateCore = WebAssembly.instantiate;
+                ")
             },
 
             Intrinsic::IsLE => output.push_str("
--- a/src/cmd/transpile.js
+++ b/src/cmd/transpile.js
@@ -180,7 +180,7 @@ export async function transpileComponent (component, opts = {}) {
 
       const outSource = `${
         imports.map((impt, i) => `import * as import${i} from '${impt}';`).join('\n')}
-${source.replace('export async function instantiate', 'async function instantiate')}
+${source.replace(/export (async )?function instantiate/, '$1function instantiate')}
 
 let ${exports.filter(([, ty]) => ty === 'function').map(([name]) => '_' + name).join(', ')};
 
--- a/src/cmd/transpile.js
+++ b/src/cmd/transpile.js
@@ -106,9 +106,12 @@ export async function transpileComponent (component, opts = {}) {
 
   let instantiation = null;
 
+  // Let's define `instantiation` from `--instantiation` if it's present.
   if (opts.instantiation) {
     instantiation = { tag: opts.instantiation };
-  } else if (opts.js) {
+  }
+  // Otherwise, if `--js` is present, an `instantiate` function is required.
+  else if (opts.js) {
     instantiation = { tag: 'async' };
   }
 
--- a/src/cmd/transpile.js
+++ b/src/cmd/transpile.js
@@ -46,7 +46,7 @@ try {
 }
 
 /**
- * @param {Uint8Array} source 
+ * @param {Uint8Array} source
  * @returns {Promise<Uint8Array>}
  */
 async function wasm2Js (source) {
@@ -60,8 +60,8 @@ async function wasm2Js (source) {
 }
 
 /**
- * 
- * @param {Uint8Array} component 
+ *
+ * @param {Uint8Array} component
  * @param {{
  *   name: string,
  *   instantiation?: 'async' | 'sync',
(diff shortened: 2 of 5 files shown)


guybedford:
Per discussion in the issue, we need to clearly separate instantiation as an implementation detail of `--js` for non-instantiation, versus actually supporting `--instantiation` alongside `--js`.

Hywan:
The PR is now ready for a review. I've updated the PR description too.

guybedford (approved):

crates/js-component-bindgen/src/intrinsics.rs
```diff
@@ -171,18 +171,10 @@ pub fn render_intrinsics(
...
-            Intrinsic::InstantiateCore => match instantiation {
-                Some(InstantiationMode::Async) | None =>  {
-                    output.push_str("
-                        const instantiateCore = WebAssembly.instantiate;
-                    ")
-                }
-
-                Some(InstantiationMode::Sync) =>  {
-                    output.push_str("
-                        const instantiateCore = WebAssembly.Instance;
```
guybedford, line 183: Is this removal fixing a bug in the previous instantiation sync work? If so, should we also change https://github.com/bytecodealliance/jco/blob/main/crates/js-component-bindgen/src/transpile_bindgen.rs#L248?

Hywan: Do you mean the default value for `instantiateCore` should be removed?

guybedford: Oh no, I mean that `instantiateCore` is still being initialized to `WebAssembly.Instance` when it was removed here? So it's a consistency point for the usage.

guybedford: Specifically shouldn't we also change that definition to be `instantiateCore = WebAssembly.instantiate` instead of `instantiateCore = WebAssembly.Instance`?

guybedford: Okay thanks for clarifying - right, so it simply doesn't apply to this case.

PR #275, opened 2023-11-24
feat: lazily generate `instanceFlags`

bnjbvr:
This lazily generates the `instanceFlags`, instead of emitting them all at once. The rationale is that if they're not required (e.g. in use cases involving only leaf components, if I understand correctly), then we can avoid allocating them and we can remove a dependency on the WebAssembly object being present. A future change could be to *not* use WebAssembly globals to represent the instance flags, but instead to use some global JS state, eliminating the dependency on the WebAssembly object more generally.


bnjbvr:
Got hit by the temporal dead zone, aka use-before-def of a let binding. So I had to add a `Source::prepend` method for the instance flags to be defined before the imports would reference them, of course.

crates/js-component-bindgen/src/transpile_bindgen.rs
```diff
@@ -599,6 +600,20 @@ impl<'a> Instantiator<'a, '_> {
         }
     }
 
+    fn ensure_instance_flags(&mut self) {
```
guybedford, line 603: We can just call this `instance_flags` I think, since the `ensure` naming is usually for cases where we need to check it against an already-written list.

crates/js-component-bindgen/src/transpile_bindgen.rs
```diff
@@ -1435,7 +1448,11 @@ impl<'a> Instantiator<'a, '_> {
         match def {
             CoreDef::Export(e) => self.core_export(e),
             CoreDef::Trampoline(i) => format!("trampoline{}", i.as_u32()),
-            CoreDef::InstanceFlags(i) => format!("instanceFlags{}", i.as_u32()),
+            CoreDef::InstanceFlags(i) => {
+                // SAFETY: short-lived borrow-mut.
+                self.used_instance_flags.borrow_mut().insert(*i);
```
guybedford, line 1453: These are quite a few hoops just to deal with a `u32`. I'd suggest rather just cloning it into the set directly to avoid all the other borrow complexity. (it's underlying datatype is just a u32)

guybedford (approved):
Looks good thank you. I suppose this fixes your issue in the simple component cases where instance flags are unused.

crates/js-component-bindgen/src/transpile_bindgen.rs
bnjbvr: oh it's because I'm mutating the `used_instance_flags` that I need the inner mutability with `RefCell`. The alternatives that I've discarded were the following:

- have `core_def` take `&mut sef`: didn't work with existing code where this would cause mutable vs immutable conflicts
- pass a `instance_flags: &mut BTreeSet` parameter
  - if that variable lived in `self` and is mutably borrowed at call sites, we get back to the previous bullet item
  - if not, then I found it a bit invasive to pass it down to all the callers, from the top-level.

Hence the inner mutability. Do you have a preference over how to achieve this?

guybedford: Ahh I see, makes sense, thanks for explaining the reasoning here!

PR #293, opened 2023-12-06
chore(tcp): migrate to node:net (wip)

manekinekko:
Explore a possible migration to node:net


guybedford:
As discussed today, we finally verified for certain this approach is untenable due to the bind being unambiguous between the server and client cases in the Node.js implementation. In addition we have verified Deno implements TCP Wrap in its compatibility layer. Sorry @manekinekko for the diversion on this, but very glad to have a solid reasoning and justification here.

PR #295, opened 2023-12-06
feat: experimental deno run

guybedford:
Leaving this one up here to track that there is an approach to test and follow this compat.

Currently it's stalling on the worker interactions, but it isn't that far off of a working Deno support for `jco run`. It would be good to pick this up for Deno compat in due course.


guybedford:
With the recent fixes by @mash-graz in https://github.com/bytecodealliance/jco/pull/387, this PR now works for some very simple preview2 test cases running in Deno.

This is now ready to land, to extend to more preview2 testing once Deno resolves https://github.com/denoland/deno/issues/22629.

PR #301, opened 2023-12-07
fix: make getChildEntry recursively traverse the path

whitequark:
I'm testing JCO together with [YoWASP](https://yowasp.org). I componentized a regular build of YoWASP Yosys using `jco new --wasi-command` and `jco transpile`, and am running it with this script:

```js
import { run } from './yosys.command.js';
import { _setArgs } from '@bytecodealliance/preview2-shim/cli';
import { _setFileData } from '@bytecodealliance/preview2-shim/filesystem';

var fileData = {
    dir: {
        "inv.v": {
            source: "module inv(input i, output o); assign o = ~i; endmodule"
        }
    }
};
_setFileData(fileData)
_setArgs(["yosys", "-Q", "./inv.v", "-o", "inv.il"])
run.run()

console.log(new TextDecoder().decode(fileData.dir["inv.il"].source))
```

Without this PR, Yosys attempting to open `./inv.v` opens the directory (and then crashes because it's not a file). I'm not entirely sure what the intent for the `getChildEntry` function is, but it seems wrong as-is, and the script completes with my changes.

### DIFF
--- a/packages/preview2-shim/lib/browser/filesystem.js
+++ b/packages/preview2-shim/lib/browser/filesystem.js
@@ -37,15 +37,18 @@ function getChildEntry (parentEntry, subpath, openFlags) {
   let segmentIdx;
   do {
     if (!entry || !entry.dir) throw 'not-directory';
-    segmentIdx = subpath.indexOf('/');
+    segmentIdx = subpath.indexOf('/', segmentIdx);
     const segment = segmentIdx === -1 ? subpath : subpath.slice(0, segmentIdx);
-    if (segment === '.' || segment === '') return entry;
-    if (segment === '..') throw 'no-entry';
-    if (!entry.dir[segment] && openFlags.create)
+    if (segment === '.' || segment === '')
+      entry = entry;
+    else if (segment === '..')
+      throw 'no-entry';
+    else if (!entry.dir[segment] && openFlags.create)
       entry = entry.dir[segment] = openFlags.directory ? { dir: {} } : { source: new Uint8Array([]) };
     else
       entry = entry.dir[segment];
-  } while (segmentIdx !== -1)
+    subpath = subpath.substring(segmentIdx + 1);
+  } while (segmentIdx !== -1);
   if (!entry) throw 'no-entry';
   return entry;
 }


packages/preview2-shim/lib/browser/filesystem.js
```diff
@@ -37,15 +37,18 @@ function getChildEntry (parentEntry, subpath, openFlags) {
   let segmentIdx;
   do {
     if (!entry || !entry.dir) throw 'not-directory';
-    segmentIdx = subpath.indexOf('/');
+    segmentIdx = subpath.indexOf('/', segmentIdx);
```
bakkot, line 40: Drive by: this will search starting at the previous `segmentIdx`, but that value was an offset into `subpath` as it was _before_ the later `subpath = subpath.substring(segmentIdx + 1)` line. Probably you don't want to have both that line and this one? 

Consider running this code on `longDirName/short/x` - the second time through the loop it will start searching the string "short/x" at index 11 (which is past the end of the string), and so miss the `/`.

whitequark: For posterity, here's the `traverse` function I eventually ended up writing, which I think is a lot more correct:

```js
  traverse(path, flags = { create: false, remove: false }) {
    let entry = this;
    let separatorAt = -1;
    do {
      if (entry instanceof File)
        throw 'not-directory';
      const files = entry.files;
      separatorAt = path.indexOf('/');
      const segment = separatorAt === -1 ? path : path.substring(0, separatorAt);
      if (separatorAt === -1 && flags.remove)
        delete files[segment];
      else if (segment === '' || segment === '.')
        /* disregard */;
      else if (segment === '..')
        /* hack to make scandir() work */;
      else if (Object.hasOwn(files, segment))
        entry = files[segment];
      else if (flags.create === 'directory' || flags.create === 'file' && separatorAt !== -1)
        entry = files[segment] = new Directory({});
      else if (flags.create === 'file')
        entry = files[segment] = new File(new Uint8Array());
      else
        throw 'no-entry';
      path = path.substring(separatorAt + 1);
    } while (separatorAt !== -1);
    return entry;
  }
```

It's not quite compatible with this specific shim but I think the algorithm is sound.

guybedford: Thanks for noticing this one @bakkot, the browser implementation is considered experimental and is kinda on the verge of either deprecation or heavy refactoring depending on which way we go here. I do like the idea of refining this implementation into something more like what was discussed in https://github.com/bytecodealliance/jco/issues/303#issuecomment-1847599114.

For now, it's kinda still placeholder code and lots of code paths are entirely untested, a decision will be made on the deprecation / refactoring soon though so it won't be staying in this state.

For now I posted a follow-up in https://github.com/bytecodealliance/jco/pull/309. Any review there would be very welcome. Always fun to have the opportunity to use a bodiless conditional...

PR #330, opened 2024-01-04
sockets: tcp refactoring

guybedford:
TCP refactoring, including:

* Unifying on a singular state integer
* Separating socket options from the other private state
* Implementing connect, accept and listen through the net handler for straightforward streams support

All TCP tests passing.


packages/preview2-shim/lib/io/worker-socket-tcp.js
```diff
@@ -1,132 +1,256 @@
...
+  const { handle } = getTcpSocketOrThrow(id);
+  const code = handle.setKeepAlive(enable);
+  if (code !== 0) throw convertSocketErrorCode(-code);
 }
 
 export function socketTcpDispose(id) {
-  const socket = getSocketOrThrow(id);
-  socket.close();
-  return 0;
+  openTcpSockets.delete(id);
```
manekinekko, line 255: should not we close the socket before deleting its reference?

PR #354, opened 2024-01-23
ci: Windows Preview2 Conformance

guybedford:
Adds Windows to the Preview2 Conformance CI

Work-in-progress, pending the following test cases:

- [ ] api_read_only
- [x] cli_file_dir_sync
- [x] preview1_interesting_paths
- [x] preview1_nofollow_errors
- [x] preview1_nofollow_errors
- [x] preview1_path_open_preopen
- [x] preview1_unlink_file_trailing_slashes

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -70,7 +70,6 @@ jobs:
       run: |
         cargo xtask generate tests
         cargo test
-      if: matrix.os != 'windows-latest'
 
   test-workspaces:
     name: Test Workspaces
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ package-lock.json
 /tests/rundir
 /tests/generated
 /tests/mod.rs
+/tests/output
--- a/packages/preview2-shim/lib/io/worker-sockets.js
+++ b/packages/preview2-shim/lib/io/worker-sockets.js
@@ -192,7 +192,7 @@ export function convertSocketErrorCode(code) {
     // TODO: return "name-unresolvable";
     // TODO: return "temporary-resolver-failure";
     default:
-      process._rawDebug('unknown error code', code);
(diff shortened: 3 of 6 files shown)


pchickey:
Just to call out, for reference- on windows, the wasmtime test runner defines the following environment variables, so that the test programs know to expect windows-like behavior:
* `ERRNO_MODE_WINDOWS`
* `NO_DANGLING_FILESYSTEM`
* `NO_RENAME_DIR_TO_EMPTY_DIR`
from https://github.com/bytecodealliance/wasmtime/blob/main/crates/test-programs/artifacts/src/lib.rs#L12-L20

The one about NO_FDFLAGS_SYNC is dead code now and should be garbage collected, ill make a quick PR for that.

guybedford:
Thanks @pchickey we do set those env vars on Window here - https://github.com/bytecodealliance/jco/pull/354/files#diff-cba25cb6f7bf4c37844e14bcaaea7be9d905e13d9fa80ef14eacbc41d9431e52R14. Note that we do actually set `NO_DANGLING_FILESYSTEM` for all tests because of the path-based implementation for Node.js FS.

These are now passing, except for one test which I've set as skipped - the `api_read_only` test. The case here is that when creating a file in a readonly directory, Wasmtime captures the directory permissions on the descriptor when it is first created, while we don't carry that information with the original open. Supporting this could be straightforward it's but a lot of wiring that doesn't fit our current model so cleanly so I've "skipped" it for now. 104/105 doesn't feel so bad to say...

pchickey:
Thanks, couldnt remember if we had those config variables all sorted but it makes sense that we do.

> These are now passing, except for one test which I've set as skipped - the api_read_only test. 
Sounds good to me, I agree we don't need to care about api_read_only on windows right now - it can be added later.

PR #364, opened 2024-01-30
fix: generated default arg WebAssembly.Instance needs to be called as constructor

zoren:
The current default parameter does not work. When running the generated code one gets:
```
    new WebAssembly.Instance,
    ^

TypeError: WebAssembly.Instance(): Argument 0 must be a WebAssembly.Module
```


guybedford:
Thanks for sharing the fix here.

PR #382, opened 2024-02-23
feat: optimized host bindings

guybedford:
With the work on [preview2-shim](https://github.com/bytecodealliance/jco/tree/main/packages/preview2-shim), we've implemented high-level bindgen for Node.js.

This proposal aims to flesh out the concept of low-level and hybrid import bindings a little more as a way to get to native binding performance and as an initial pointer for work in this direction.

An early version of this `--import-bindings` flag is provided supporting these host bindings, with some simple examples implemented for `getRandomBytes`, `now` and `resolution`.


alexcrichton (commented):
This looks quite nice and I can see this being quite useful in JS!

At a high level this looks good to me, I think it's the right approach to have a function-returning-a-function to capture the canonical abi options for any lowering. 

At a lower level though there's two concerns I would have with this exact approach, and some slight modifications I might recommend are:

* There may want to be some sort of generic "context" passed in as an argument as well. I'm mainly thinking about types like resources where indices need to be validated independently to lookup other objects. 
* Another thing to worry about here is all the "bad" or "misbehaved" cases of the canonical ABI. For example the optimized lowering of `wallClock.now` doesn't validate that the return pointer is aligned, but the component model should trap in that case. This also relates to the above case where there's a lot of trapping cases when it comes to resources and optimized lowerings need to be sure to handle all that.

The context object might be able to have some helper methods on it along the lines of all the `trap_if` calls in the canonical ABI document perhaps?

guybedford:
Thanks for the review, this is great feedback.

I think the assumption of low-level bindgen must be that the implementation of that low level bindgen is valid according to the component model spec.

So the onus is on that low-level bindgen to handle the correct traps. I've gone ahead and added the retptr alignment traps to the implementation. I think an inline conditional throw is better than an explicit function call, given that performance is the primary goal here to avoid that extra function call.

Handling all the various cases is definitely a big implementation question, and my hope would be that we could flesh out the preview2 API in this way, while running benchmarks to verify the wins, to also include all the correct behaviours and traps.

If we can do that, then this preview2 shim implementation can be a good example low-level bindgen to work from, and also for lowering into C++ eventually in Node.js.

In terms of the structure of the API, completely agreed resources will have a bunch of needs we need to ensure are met here, and I don't fully understand how we need to handle all those interactions - having a context argument seems like it may well be necessary to support various resource operations.

Let me try to work through a simple resource integration here as part of this proposal before landing, and see what shape of context might work for that. Seems like that might be a good way to validate the model further to me.

Any suggestions or guidance with respect to how to support resources very much welcome.

alexcrichton:
> So the onus is on that low-level bindgen to handle the correct traps.

Makes sense yeah, and also makes sense to not use extra function calls to keep this fast. I think it's basically worth highlighting this as a risk of this approach. Getting all the fiddly bits of the ABI right is not easy, even in Wasmtime in the few places we've had to do this we've had differences by accident. I suspect there's already differences between jco and Wasmtime today, too. 

That being said, I don't see this as a reason to abandon this work by any means. It's an important risk to highlight and is something to be aware of when implementing the low-level functions, but there is regardless a standard which can be pointed to whenever there's a disagreement between two implementations. This to me means that while this is a risk it's not an existential risk.

One thing we could do much better perhaps is to make it easier to test an implementation against an API. For example some sort of like fuzz-generated component or something like that. No idea how to do this concretely, but in the future it's something we could invest in perhaps.

> Any suggestions or guidance with respect to how to support resources very much welcome.

My best suggestion here would be to start as minimal as possible. For example don't allow `cabiLower` optimizations on necessarily all functions. For example the resource-related methods in WASI are all pretty simple, especially the ones that look like methods. If you don't have to handle things like `list<borrow<T>>` it'll probably make things at least initially much simpler.

I suspect though that this is still going to be somewhat tricky. You sort of want to limit "groups" of functions as one, for example whomever defines a resource would need to own the low-level `cabiLower` implementation of all functions that create or operate on that resource. How to model that I'm not sure.

guybedford:
I've put together a very rough low-level resource bindgen concept in https://github.com/bytecodealliance/jco/pull/382/files#diff-d089478d43999a7e0cb8c9e6124ab131485200dea2e1380c0bf77978f3bfc436R57.

We can assume that imports are at the bottom of the graph, and so the import side of the membrane doesn't need to know anything about the export side of the membrane, therefore the host component that is defining the high-level JS bindgen is the one with the universal picture, calling into the imported bindgen to manage the lifecycles. Therefore, this is very much a "polyfill model", and not what we would do for native support, rather it remains a cowpath to native support performance.

There is a new problem for resources being lowered though, and that is that if an optimized imported resource is exposed as an export of the component, then we potentially would still need to use the high-level bindgen for that resource.

Some options:
1. Only support low-level bindgen for resources which are not exposed on the exported membrane. This seems like it would work pretty much immediately with the constraints of the scheme as defined so far.
2. Define the concept of resource transfer to exist between the high level and low-level bindgen for the same imported resource, effectively as separate resource tables. We would need to split the resource table index for the low-level bindgen resources into two cases at every point, with exported functions using the one and local functions using the other. Transfer and other trampolines would all need to somehow be aware of this so I'm not sure how straightforward that would be.

This then also has me thinking about what it means to have separately transpiled components interacting through optimized bindgen. I'm tempted to say low-level bindgen optimizations should be reserved from transpiled component -> host imports only for now, instead of expecting this to magically work between transpiled components. Rather that's the benefit of component composition. So let's just focus on the direct transpiled code running against optimized host imports for now I think?

alexcrichton:
I think that the intrinsics you've sketched will probably need some more context, but I don't know exactly what it would look like. Anything that works with a `handle: i32` needs to be relative to a table, and there's not just a single table that can be baked into the class. For example if you lower a resource constructor into two different components (within the same outer component so it's still just one "thing" being bound by jco) then those two components have different resource tables. They might both have a handle 0 but that handle 0 would point to entirely different resources. This is more-or-less the `TypeResourceTableIndex` in Wasmtime.

So given that you'd sort of need something like:

* `ResourceClass[Symbol.for('cabiResourceNew')](table: i32, rep: i32) -> handle: i32`
* `ResourceClass[Symbol.for('cabiResourceRep')](table: i32, handle: i32) -> rep: i32`

(etc)

I'm also not certain about `cabiResourceDeref`, `cabiResourceRefOwn`, and `cabiResourceRefBorrow` here. Transferring a resource is pretty subtle and if table indices are already being passed (which I think they need to above) it might be worth just having a `transferBetween(tableA: i32, tableB: i32, handleInA: i32) -> handleInB: i32` or something like that

Issue #376, opened 2024-02-15
Generated ES modules has .js extension


morigs:
When using `jco transpile` it generates es modules with `.js` extension which causes error when importing from es module in nodejs.

```
***/index.mjs:1
import { example } from './bindings/example.component.js';
         ^^^^^^^^^
SyntaxError: Named export 'example' not found. The requested module './bindings/example.component.js' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from './bindings/example.component.js';
const { example } = pkg;

    at ModuleJob._instantiate (node:internal/modules/esm/module_job:132:21)
    at async ModuleJob.run (node:internal/modules/esm/module_job:214:5)
    at async ModuleLoader.import (node:internal/modules/esm/loader:323:24)
    at async loadESM (node:internal/process/esm_loader:28:7)
    at async handleMainPromise (node:internal/modules/run_main:120:12)

Node.js v21.6.1
```

Running Node.js with this flag helps, but looks unconventional...

```
node --experimental-default-type module index.mjs
```

guybedford:
This is correct - add `"type": "module"` to your package.json file.

morigs:
ooh, didn't think of that since I'm making a SPA, not a library. But it sounds logical, thanks!
If this is the recommended way, it would be cool to document it (including in the example).

PR #386, opened 2024-02-28
docs: fix end-to-end example

### LINKED ISSUE #278: Current example provided in docs isn't very beginner-friendly
The current example listed as part of the docs is not very beginner-friendly, primarily because people don't tend to have a cowsay.wasm lying around.

One of the things we could do to improve UX is to:

- Teach them how to install the jco project
- Create a simple hello-world scaffold in Node JS
- Run scaffold in NodeJS
- Teach them to componentize it using the componentize function to wasm
- Transpile said function back again & compare it with the scaffold we created in step 2 to prove that it'd be identical-ish.

guybedford:
This makes the existing example use ComponentizeJS to create the demo component so that it is a fully runnable end to end example.

Resolves #278.

### DIFF
--- a/docs/src/example.md
+++ b/docs/src/example.md
@@ -54,7 +54,7 @@ return `${text}
 To turn this into a component run:
 
 ```shell
-jco componentize cowsay.js --wit cowsay.wit -o cowsay.component.wasm
+jco componentize cowsay.js --wit cowsay.wit -o cowsay.wasm
 
 OK Successfully written cowsay.component.wasm with imports ().
 ```
--- a/docs/src/example.md
+++ b/docs/src/example.md
@@ -6,6 +6,43 @@ Jco also provides an experimental feature for generating components from JavaScr
 
 To demonstrate a full end-to-end component, we can create a JavaScript component embedding Spidermnokey then run it in JavaScript.
 
+### Installing Jco
+
+Either install Jco globally:
+
+```shell
+$ npm install -g @bytecodealliance/jco
+$ jco --version
+1.0.3
+```
+
+Or install it locally and use `npx` to run it:
+
+```shell
+$ npm install @bytecodealliance/jco
+$ npx jco --version
+1.0.3
+```
+
+Local usage can be preferable to ensure the project is reproducible and self-contained, but requires
+replacing all `jco` shell calls in the following example with either `./node_modules/.bin/jco` or `npx jco`.
+
+### Installing ComponentizeJS
+
+To use ComponentizeJS, it must be separately installed, locally or globally depending on whether Jco was installed locally or globally:
+
+```shell
+$ npm install -g @bytecodealliance/componentize-js
+```
+
+Or locally:
+
+```shell
+$ npm install @bytecodealliance/componentize-js
+```
+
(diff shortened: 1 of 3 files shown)


bakkot:
Nice! Went through it. Ran into the following problems:

- Following the "install" instructions in the main readme will not put `jco` on your path, so the commands which say to run `jco` will not work. Users either need to `install -g`, or prefix all `jco` commands with `npx`, as in `npx jco componentize cowsay.js --wit cowsay.wit -o cowsay.wasm`.
- The first time you run `componentize` it will fail and tell you that you need to run `npm install @bytecodealliance/componentize-js`.
- Running the test file will fail because it's trying to `import` from a file (`./cowsay/cowsay.js`) which ends in `.js`, which npm treats as cjs by default. Users either need to be told to create a `package.json` (as in [the ComponentizeJS example](https://github.com/bytecodealliance/ComponentizeJS/blob/54a7a0d341bf7f70bfa50899288931a61b4c1364/EXAMPLE.md#running-the-component-in-nodejs)), or (my preference) the code [needs to generate `.mjs` files](https://github.com/bytecodealliance/jco/issues/376).

guybedford:
Thanks for taking a look, added some install instructions and `type: module` instructions.

docs/src/example.md
```diff
@@ -1,88 +1,193 @@
...
+```
 
-As a first step, we might like to look instead this binary black box of a Component and see what it actually does.
+Local usage can be preferable to ensure the project is reproducible and self-contained, but requires
+replacing all `jco` shell calls in the following example with either `./node_modules/.bin/jco` or `npx jco`.
+
+### Installing ComponentizeJS
 
-To do this, we can use `jco wit` to extract the "WIT world" of the Component ([WIT](https://github.com/WebAssembly/component-model/blob/main/design/mvp/WIT.md) is the typing language used for defining Components).
+To use ComponentizeJS, it must be separately installed, locally or globally depending on whether Jco was installed locally or globally:
```
bakkot, line 32: Nit: the order here suggests that "npm install -g" is the "locally" option. Also you say "or locally" below but there's no first part of that "or".

```suggestion
To use ComponentizeJS, it must be separately installed, globally or locally depending on whether Jco was installed globally or locally. Globally:
```

docs/src/example.md
```diff
@@ -1,88 +1,193 @@
...
+  \   ^__^
+    \  (oo)\______
+      (__)\      )/\
+          ||----w |
+          ||     ||
+```
+
+Passing in the optional second parameter, we can change the cow:
+
+test.mjs
```
bakkot, line 170:
```suggestion
test.js
```

docs/src/example.md
```diff
@@ -1,88 +1,193 @@
...
 
-Alternatively `jco print cowsay.wasm -o out.wat` would output the full concrete Wasm WAT to inspect the Component, with all the implementation details (don't forget the `-o` flag...).
+```shell
+$ jco wit cowsay.wasm
+package root:component;
+
+world root {
+  export cow: interface {
+    enum cows {
+      cat,
```
bakkot, line 112:
```suggestion
```

Issue #385, opened 2024-02-27
`deno` compatibility (use of  `node:process._rawDebug()`)


mash-graz:
It would be nice, if `jco` and its transpilation capabilities could be used/executed also by `deno` and other popular JS runtimes.

Right now we get an error on every execution attempt:

```bash
deno run -A npm:@bytecodealliance/jco --help
error: Uncaught SyntaxError: The requested module 'node:process' does not provide an export named '_rawDebug'
    at <anonymous> (file:///home/mash/.cache/deno/npm/registry.npmjs.org/@bytecodealliance/preview2-shim/0.15.1/lib/io/worker-io.js:36:10)
```

guybedford:
Yes we are interested in this, but it's not currently a development priority.

Any PRs to improve Deno compatibility will be gladly merged though.

mash-graz:
The use of `_rawDebug()` seems to be in place since #334

guybedford:
Note that the main blocking issue for Deno support is related worker threads behaviours I believe.

mash-graz:
This particular issue of `_rawDebug()` should be easily solvable by using a more common logging mechanism, but it just couldn't get further. There may be other much more significant obstacles waiting behind this more or less trivial flaw. But in general it would be very appreciable, if `deno` compatibility could solved and tested in your CI pipeline. 

`Jco` transpilation seems to be the only option to utilize WASM components on `deno deploy`. That's why a working execution of this tool on this particular runtime would be really helpful.

mash-graz:
On a quick search I found three incompatibilities:

1. `_rawDebug` -- can be replaced by `console.error`
2.  `lutimesSync` -- that's indeed still missing in `deno` (#18218) but there are workarounds available in external sources.
3. `worker_threads` in `deno` are missing the `unref` function. But this shouldn't be a significant obstacle as long as all connected message handlers and timers are removed properly.

Although I could quickly figure out temporal workarounds for all of these conflicts, they still should be patched in more reliable manner by someone more familiar with this code base.

guybedford:
@mash-graz thanks for summarizing your findings here. With those changes were you able to get the implementation to work in Deno?

mash-graz:
yes -- after bypassing these three obstacles I could run `jco` and its `traspilation` command.

But I did it just as a very quick and incomplete sketch to look how much efforts are needed.
I'm still looking how to realize it in a more adequate fashion.

questions: 

[...]

PR #387, opened 2024-02-28
fixes for `deno` compatibility

mash-graz:
Just a first draft PR to feed the CI pipeline, because local testing doesn't work on my machine.

Fixes: #385

Warning! -- there are important parts missing/untested!

* `validateHeaderName` and `validateHeaderValue` are temporary replaced by dummy entries.

  This should be better fixed on the `deno` side (see: https://github.com/denoland/deno/issues/22614).

 * The `worker_thread.unref()` case is still uttery unsolved and just temporary removed.

### DIFF
--- a/packages/preview2-shim/lib/nodejs/http.js
+++ b/packages/preview2-shim/lib/nodejs/http.js
@@ -21,8 +21,11 @@ import {
   registerDispose,
   registerIncomingHttpHandler,
 } from "../io/worker-io.js";
-import { validateHeaderName, validateHeaderValue } from "node:http";
+// import { validateHeaderName, validateHeaderValue } from "node:http";
 import { HTTP } from "../io/calls.js";
+
+function validateHeaderName(_name){}
+function validateHeaderValue(_name, _value){}
 
 const symbolDispose = Symbol.dispose || Symbol.for("dispose");
 export const _forbiddenHeaders = new Set(["connection", "keep-alive"]);
--- a/packages/preview2-shim/lib/nodejs/http.js
+++ b/packages/preview2-shim/lib/nodejs/http.js
@@ -21,11 +21,10 @@ import {
   registerDispose,
   registerIncomingHttpHandler,
 } from "../io/worker-io.js";
-// import { validateHeaderName, validateHeaderValue } from "node:http";
 import { HTTP } from "../io/calls.js";
 
-function validateHeaderName(_name){}
-function validateHeaderValue(_name, _value){}
+import * as http from "node:http";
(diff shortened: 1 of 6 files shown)


packages/preview2-shim/lib/synckit/index.js
```diff
@@ -87,7 +87,7 @@ export function createSyncFn(workerPath, debug, callbackHandler) {
     }
     return result;
   };
-  worker.unref();
+  // worker.unref();
```
guybedford, line 90:
```suggestion
  if (worker.unref) worker.unref();
```

packages/preview2-shim/lib/nodejs/http.js
```diff
@@ -21,9 +21,12 @@ import {
   registerDispose,
   registerIncomingHttpHandler,
 } from "../io/worker-io.js";
-import { validateHeaderName, validateHeaderValue } from "node:http";
+// import { validateHeaderName, validateHeaderValue } from "node:http";
```
guybedford, line 24:
```suggestion
import * as http from "node:http";
const { validateHeaderName = () => {}, validateHeaderValue = () => {} } = http;
```

guybedford (approved):
I'd be happy to land this with the suggested changes.

packages/preview2-shim/lib/synckit/index.js
mash-graz: yes -- that's a much nicer solution! :)

Nevertheless, I think, it's not enough!

The timeouts in the CI pipeline are most likely caused by hanging processes affected by this particular change resp. broken termination of workers. :(

We have to check for active event listeners or timers...

packages/preview2-shim/lib/nodejs/http.js
mash-graz: That's also a very nice improvement!

I just committed both changes, but it still needs a lot more of additional testing and rework.

guybedford:
Thanks for working on this. Let's at least land for now, we can always improve later.

mash-graz:
> Thanks for working on this. Let's at least land for now, we can always improve later.

Yes -- that's a good interim solution, because the merged changes seem to do no harm to already existing use cases and test routines.

I just waited, because I could isolate the non-termination issue as another `deno` flaw (see:  https://github.com/denoland/deno/issues/22629) and I'm still searching a working solution to reuse the existing `mocha` tests also for `deno` unit checks (That's rather tricky, because it's very hard to create working `import_map`-redirections in mixed `node`/`deno`-workflows). But it's perfectly fine for me to send you these additional improvements in new PRs.

Thanks for all your help! It's really a pleasure to collaborate with you!

PR #393, opened 2024-03-02
feat: slab implementation for resource tables

guybedford:
This updates the resource tables to use a fixed sized array of numbers with a free list as a slab implementation for more optimal resource operations.

The table data structure is also designed to be used in the new low-level bindgen for optimized table operations..

Previously, we were using a non-numeric `rep` for imported resources, where instead of using a number we would just reference the JS class instance directly in this position. Now that `rep` is only numeric, this also implements a separate `captureTable` to do rep assignment for captured imported resources, while retaining the same semantics.

### DIFF
--- a/crates/js-component-bindgen/src/function_bindgen.rs
+++ b/crates/js-component-bindgen/src/function_bindgen.rs
@@ -45,7 +45,8 @@ pub enum ResourceData {
 ///
 /// For a given resource id {x}, the local variables are assumed:
 /// - handleTable{x}
-/// - handleCnt{x}
+/// - captureTable{x}
+/// - captureCnt{x}
 ///
 /// For component-defined resources:
 /// - finalizationRegistry{x}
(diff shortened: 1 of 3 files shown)


alexcrichton (commented):
Seems fine to me yeah, but if possible I'd highly recommend getting tests for this as that's what it took in Wasmtime to flesh out some bugs here and there

guybedford:
Thanks, at the very least this already fills in the new handling of scope checking for borrows, which was previously not implemented. We still do not have own handle ref counting for verifying own drops don't have outstanding borrows.

I believe will be one of the last pieces required to consider out implementation complete, and padding out a test suite as part of that would be amazing. Supporting resources has very much been a step by step process getting it wrong at first, then adding correctness later at each of those steps.

Having a test suite that could match that incremental adoption strategy in terms of turning on or off layers of expectations would be amazing, but that's effectively what we've done here in practise.

At least now that we are nearly at the final layer we can move in this direction though - any tips with regards to how to get the test suite wired in would be very welcome.

Issue #394, opened 2024-03-06
Incorrect codegen for `wasi:http` resource `RequestOptions`


landonxjames:
Using the following versions of `jco`:
```
"@bytecodealliance/jco": "^1.0.2",
"@bytecodealliance/preview2-shim": "^0.15.1",
```

Looking at the code for the `http` modules in the `preview2-shim` each method ends with `*TimeoutMs`:
https://github.com/bytecodealliance/jco/blob/32cc6123edb2cc032c00079bccdef1a1fdc33f87/packages/preview2-shim/lib/nodejs/http.js#L178-L200

But in the code generated by `jco transpile` all of these methods are created without the `Ms` at the end:
```js
const ret = RequestOptions.prototype.firstByteTimeout.call(rsc0);
...
const ret = RequestOptions.prototype.connectTimeout.call(rsc0);
```

These functions fail when called since `RequestOptions.prototype.connectTimeout` is `undefined`. Doing a naive find and replace `*Timeout.call -> *TimeoutMs.call` fixes the issue and allows me to complete the http call.

Changing the names of the methods in `RequestOptions` would probably be the quickest way to fix this, but it doesn't really account for why they had the `*Ms` appended  in the first place? They do seem different than the standard WASI `set_*_timeout` methods since those take `Option<Duration>` as input and the WASI `Duration` type is based on [`u64` nanoseconds](https://docs.rs/wasi/latest/wasi/clocks/monotonic_clock/type.Duration.html) instead of the milliseconds these presumably take.

Throwing some logging in the generated code, the value passed to the (renamed) `setConnectTimeoutMs` shows the initial nanosecond value passed in by the Rust code (`3.1s` or `3100000000ns`), so I think a `nanosecond -> millisecond` conversion might be missing somewhere. 

```js
console.log("setConnectTimeoutMs INPUT:", variant2);
    console.log("variant2 type:", typeof variant2);
    ret = {
      tag: "ok",
      val: RequestOptions.prototype.setConnectTimeoutMs.call(rsc0, variant2),
    };
```

outputs:
```
  console.log
    setConnectTimeoutMs INPUT: 3100000000n

      at trampoline15 (src/generated-js-runtime-bindings/handoff.js:191:13)

  console.log
    variant2 type: bigint

      at trampoline15 (src/generated-js-runtime-bindings/handoff.js:192:13)
```

And finally, I get a warning output from node:
```
[...]

PR #395, opened 2024-03-06
Updating `wasi:http` `RequestOptions` to  properly handle nano/milliseconds

landonxjames:
See #394 for a more detailed discussion. 

In short the current implementation of `wasi:http` `RequestOptions` incorrectly handles nano/milliseconds and it also did not export the methods expected by the WASI spec.


guybedford:
Thanks for posting a PR for this.

PR #404, opened 2024-03-14
Adding default timeout values for HTTP requests

landonxjames:
Adding a default timeout value of `600_000`Ms to each of the HTTP timeouts to align with the [defaults set by Wasmtime](https://github.com/bytecodealliance/wasmtime/blob/36fb62ca3bea7dd1456f12ead03084e2ceb48cda/crates/wasi-http/src/http_impl.rs#L23-L33).


packages/preview2-shim/lib/nodejs/http.js
```diff
@@ -179,32 +179,59 @@ class RequestOptions {
   #connectTimeout;
   #firstByteTimeout;
   #betweenBytesTimeout;
+  #defaultTimeoutMs = 600_000n;
+  #nanoToMilliConversion = 1_000_000n;
```
guybedford, line 183: These don't need to be a private field, instead they can just be top-level constants like `const defaultTimeoutMs = ...; const nanoTomilliConversion = ...`.

packages/preview2-shim/lib/nodejs/http.js
```diff
@@ -179,32 +179,59 @@ class RequestOptions {
   #connectTimeout;
   #firstByteTimeout;
   #betweenBytesTimeout;
```
guybedford, line 181: Making the value a top-level constant, we can then set the default value here - `#connectTimeout = defaultTimeoutMs`

guybedford (commented):
Happy to do this, with a few changes.

packages/preview2-shim/lib/nodejs/http.js
landonxjames: Updated in https://github.com/bytecodealliance/jco/pull/404/commits/dcf0f31934b9b81b560a7190bc8c0fae8608223f

packages/preview2-shim/lib/nodejs/http.js
landonxjames: Updated in https://github.com/bytecodealliance/jco/pull/404/commits/dcf0f31934b9b81b560a7190bc8c0fae8608223f

landonxjames:
I'll try to look into these failing tests in a bit, but it looks like they are all Deno related and none seem to be testing any HTTP functionality so probably not hitting the changed code. Are the Deno tests known to be a bit flakey?

```
failures:
    gen::deno_api_reactor::api_reactor
    gen::deno_api_time::api_time
    gen::deno_cli_args::cli_args
    gen::deno_cli_env::cli_env
    gen::deno_cli_exit_default::cli_exit_default
    gen::deno_cli_exit_success::cli_exit_success
    gen::deno_preview1_big_random_buf::preview1_big_random_buf
    gen::deno_preview2_random::preview2_random
```

guybedford (approved):
Thanks I'll land this once we've fixed the Deno case. This is likely due to a Deno upgrade.

Issue #405, opened 2024-03-19
Non-conformant exceptions should unwrap as traps


guybedford:
Rather than having a coercing `getErrorPayload`, if the error is a real JS error it should be rethrown.

guybedford:
I guess the issue here is what if you want to call an arbitrary JS library from a component, that might throw for a given function - we want there to exist a way to type the JS error.

Instead what I think we can do is treat errors of the form `result<_, string>`, specifically, as coercible from JS errors, but otherwise JS errors without a payload are non-coercible and trap.

guybedford:
This was fixed in #408.

PR #408, opened 2024-03-22
Toolchain deps update

guybedford:
Updates to the latest toolchain dependencies including Wasmtime 19.

Also includes some features and fixes:
* Refined error handling resolving https://github.com/bytecodealliance/jco/issues/405. JS errors from imported JS functions now only coerce into results when the top-level result accepts a string error type, otherwise the JS error will trap.
* Fixes a regression with the timeout API
* Fixes the connection timeout support to pass the new Wasmtime test for that


packages/preview2-shim/lib/io/worker-http.js
```diff
@@ -127,7 +124,7 @@ export async function createHttpRequest(
           host: authority.split(":")[0],
           port: authority.split(":")[1],
           path: pathWithQuery,
-          timeout: connectTimeout && Number(connectTimeout),
+          timeout: connectTimeout && Number(connectTimeout / 1_000_000n),
```
landonxjames, line 127: I think the `firstByteTimeout` and `betweenBytesTimeout` need this division step as well?

https://github.com/bytecodealliance/jco/blob/961feb73c3208cf1933db3e6b75ccb6f641b85ae/packages/preview2-shim/lib/io/worker-http.js#L150-L154

guybedford: Great catch, thanks, added https://github.com/bytecodealliance/jco/pull/409/.

Issue #375, opened 2024-02-13
Selecting instantiation mode at compile time is limiting


whitequark:
I've just encountered this error message in Chrome:

```
bundle.js:487 Uncaught RangeError: WebAssembly.Instance is disallowed on the main thread, if the buffer size is larger than 8MB. Use WebAssembly.instantiate, or use the flag `--enable-features=WebAssemblyUnlimitedSyncCompilation`.
    at bundle.js:487:28
    at Application.instantiate (bundle.js:4439:28)
    at Application.execute (bundle.js:483:30)
    at Application.run (bundle.js:513:17)
    at <anonymous>:1:7
```

That is ... a problem. I want to provide an option to do everything synchronously (except for the necessary network requests that can be run in advance), provided that you are running in some context where blocking is OK (like a web worker). But I think that completely removing the option to run *outside* of a web worker, even if it's generally unwise, is a regresstion, if for no other reason that it makes it impossible to quickly load something in DevTools and experiment.

I think I want JCO to allow me to make this choice at runtime, not at compile-time.

guybedford:
I guess we could create an `--instantiation hybrid` mode that calls the function, and if it returns a promise conditionally awaits it. Would gladly take a PR along those lines, and if it's not too big of a diff we could even make it the default instantiation mode.

whitequark:
@guybedford So while working on this I bumped into a question, namely, why do we have separate `getCoreModule` and `instantiateCore = <some default behavior>`? This seems to serve no purpose and creates a lot of complexity because there are three possible cases, all of which may be desirable in some circumstances:

1. `getCoreModule` returns a `Promise<Response>` and `instantiateCore` is `WebAssembly.instantiateStreaming` (which is actually slower than it has to be with `--instantiation async` because you first await all `getCoreModule` promises and only then start instantiation, which leaves parallelism opportunities on the table--that can get tangible with how many core modules jco outputs!)
2. `getCoreModule` returns an `ArrayBuffer` or `Promise<ArrayBuffer>` and `instantiateCore` is `WebAssembly.instantiate`. Works well if you don't necessarily have a `Response` around (although you can always construct a `Response` out of an `ArrayBuffer` if you want), but still means async instantiation, so not always desirable or viable.
[...]

PR #413, opened 2024-03-23
Generator-based initialization function

guybedford:
Based on discussion in https://github.com/bytecodealliance/jco/issues/375.

### DIFF
--- a/crates/js-component-bindgen/src/transpile_bindgen.rs
+++ b/crates/js-component-bindgen/src/transpile_bindgen.rs
@@ -256,38 +256,24 @@ impl<'a> JsBindgen<'a> {
             self.opts.instantiation.is_some(),
         );
 
-        match self.opts.instantiation {
-            Some(InstantiationMode::Async) => {
-                uwrite!(
-                    output,
-                    "\
-                        export async function instantiate(getCoreModule, imports, instantiateCore = WebAssembly.instantiate) {{
-                            {}
-                            {}
-                            {}
-                    ",
-                    &js_intrinsics as &str,
-                    &intrinsic_definitions as &str,
-                    &compilation_promises as &str,
-                )
-            }
-
-            Some(InstantiationMode::Sync) => {
-                uwrite!(
-                    output,
-                    "\
-                        export function instantiate(getCoreModule, imports, instantiateCore = (module, importObject) => new WebAssembly.Instance(module, importObject)) {{
-                            {}
-                            {}
-                            {}
-                    ",
-                    &js_intrinsics as &str,
-                    &intrinsic_definitions as &str,
-                    &compilation_promises as &str,
-                )
-            }
-
-            None => {}
+        if let Some(instantiation) = &self.opts.instantiation {
+            uwrite!(
+                output,
+                "\
+                    {}
+                    export function instantiate(getCoreModule, imports, instantiateCore = {}) {{
+                        {}
+                        {}
+                ",
+                &js_intrinsics as &str,
+                match instantiation {
+                    InstantiationMode::Async => "WebAssembly.instantiate",
+                    InstantiationMode::Sync =>
+                        "(module, importObject) => new WebAssembly.Instance(module, importObject)",
+                },
+                &intrinsic_definitions as &str,
+                &compilation_promises as &str,
+            );
         }
 
         let imports_object = if self.opts.instantiation.is_some() {
@@ -308,8 +294,31 @@ impl<'a> JsBindgen<'a> {
             uwrite!(
                 output,
                 "\
-                        {}\
-                        {};
+                        let gen = (function* init () {{
+                            {}\
+                            {};
+                        }})();
+                        let promise, resolve, reject;
+                        function runNext (value) {{
+                            try {{
+                                let done;
+                                do {{
+                                    ({{ value, done }} = gen.next(value));
+                                }} while (!(value instanceof Promise) && !done);
+                                if (done) {{
+                                    if (resolve) resolve(value);
+                                    else return value;
+                                }}
+                                if (!promise) promise = new Promise((_resolve, _reject) => (resolve = _resolve, reject = _reject));
+                                value.then(nextVal => done ? resolve() : runNext(nextVal), reject);
+                            }}
+                            catch (e) {{
+                                if (reject) reject(e);
+                                else throw e;
+                            }}
+                        }}
+                        const maybeSyncReturn = runNext(null);
+                        return promise || maybeSyncReturn;
                     }}
                 ",
                 &self.src.js_init as &str,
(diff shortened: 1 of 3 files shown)


crates/js-component-bindgen/src/transpile_bindgen.rs
```diff
@@ -326,8 +349,31 @@ impl<'a> JsBindgen<'a> {
                     {}
                     {}
                     {maybe_init_export}const $init = (async() => {{
```
bakkot, line 351: If you're in an async function you don't have to juggle promises manually. You can drive the generator with something like

```js
let done, value, threw;
while (true) {
  0, { done, value } = threw ? gen.throw(value) : gen.next(value);
  if (done) break;
  try {
    value = await value;
    threw = false;
  } catch (e) {
    value = e;
    threw = true;
  }
}
return value;
```

crates/js-component-bindgen/src/transpile_bindgen.rs
```diff
@@ -294,8 +294,31 @@ impl<'a> JsBindgen<'a> {
...
+                        let promise, resolve, reject;
+                        function runNext (value) {{
+                            try {{
+                                let done;
+                                do {{
+                                    ({{ value, done }} = gen.next(value));
+                                }} while (!(value instanceof Promise) && !done);
+                                if (done) {{
+                                    if (resolve) resolve(value);
+                                    else return value;
```
bakkot, line 310: This `return` doesn't go anywhere: one of the two calls to `runNext` only happens in the `if (resolve)` case, and the other is thrown away rather than being returned.

I can't tell if this block of code is meant to support both sync and async instantiation. If it's just sync, you don't need to worry about promises, and can drive the generator much more simply:

```js
let done, value;
while (!done) {
  0, { done, value } = gen.next(value);
}
return value;
```

crates/js-component-bindgen/src/transpile_bindgen.rs
guybedford: For now I've gone with the "hybrid bindgen" implementation, but specializing to sync and async better could make sense.

crates/js-component-bindgen/src/transpile_bindgen.rs
guybedford: That was an oversight, fixed.

guybedford:
Here's an example of the flavorful instantiation test output - https://gist.github.com/guybedford/06abf2ec8a17238e86413ce6af51487e.

whitequark:
> Here's an example of the flavorful instantiation test output - https://gist.github.com/guybedford/06abf2ec8a17238e86413ce6af51487e.

Sorry, I don't understand--this doesn't seem like it lets me pick instantiation style after the fact.

guybedford:
@whitequark it implements an initialization function that is sync and only returns a promise if one of the instantiate hooks returns a promise, so solves your use case. As I say I have no horse in this race, I just thought it might help you.

whitequark:
Oh I see--thanks, I got very confused due to the presence of asm.js code in the output and some of the stuff at the end. I think this will work for me--I can test that and report back.

guybedford:
Great, please do, this needs a few tweaks to land still, so that feedback will help a lot to know if it's worth doing so!

crates/js-component-bindgen/src/transpile_bindgen.rs
```diff
@@ -325,9 +336,32 @@ impl<'a> JsBindgen<'a> {
...
+                                }}
+                                if (!promise) promise = new Promise((_resolve, _reject) => (resolve = _resolve, reject = _reject));
+                                value.then(nextVal => done ? resolve() : runNext(nextVal), reject);
```
guybedford, line 356: Since the done path is handled above, this should probably just be:

```suggestion
                                value.then(runNext, reject);
```

whitequark:
How do I actually test this? I.e. what do I put in my `dependencies` dictionary in `package.json`?

whitequark:
I tried

    "@bytecodealliance/jco": "git://github.com/bytecodealliance/jco#generator"

But I get this error that I don't know how to deal with:

```
> @yowasp/runtime-test@0.0.0 transpile
> jco new ../copy.wasm --wasi-command --output copy.wasm && jco transpile copy.wasm --instantiation sync --no-typescript --no-namespaced-exports --map 'wasi:io/*=runtime#io' --map 'wasi:cli/*=runtime#cli' --map 'wasi:clocks/*=runtime#*' --map 'wasi:filesystem/*=runtime#fs' --out-dir gen/

node:internal/errors:496
    ErrorCaptureStackTrace(err);
    ^

Error [ERR_MODULE_NOT_FOUND]: Cannot find module '/home/whitequark/Projects/yowasp/runtime-js/test/yowasp_runtime_test/node_modules/@bytecodealliance/jco/obj/wasm-tools.js' imported from /home/whitequark/Projects/yowasp/runtime-js/test/yowasp_runtime_test/node_modules/@bytecodealliance/jco/src/cmd/opt.js
    at new NodeError (node:internal/errors:405:5)
    at finalizeResolution (node:internal/modules/esm/resolve:327:11)
    at moduleResolve (node:internal/modules/esm/resolve:980:10)
    at defaultResolve (node:internal/modules/esm/resolve:1193:11)
    at ModuleLoader.defaultResolve (node:internal/modules/esm/loader:403:12)
    at ModuleLoader.resolve (node:internal/modules/esm/loader:372:25)
    at ModuleLoader.getModuleJob (node:internal/modules/esm/loader:249:38)
    at ModuleWrap.<anonymous> (node:internal/modules/esm/module_job:76:39)
    at link (node:internal/modules/esm/module_job:75:36) {
  url: 'file:///home/whitequark/Projects/yowasp/runtime-js/test/yowasp_runtime_test/node_modules/@bytecodealliance/jco/obj/wasm-tools.js',
  code: 'ERR_MODULE_NOT_FOUND'
}
```

whitequark:
@guybedford That doesn't look like it works:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find module '/home/whitequark/Projects/yowasp/runtime-js/test/yowasp_runtime_test/node_modules/@bytecodealliance/jco/obj/wasm-tools.js' imported from /home/whitequark/Projects/yowasp/runtime-js/test/yowasp_runtime_test/node_modules/@bytecodealliance/jco/src/cmd/opt.js
```

whitequark:
Do you think you could just test YoWASP/runtime-js with this branch? There is clearly some sort of significant issue with how jco development is done and I can't really spend the effort right now to figure it out on top of the things I actually want to do...

whitequark:
It looks like the install of `jco` that's inside of `./node_modules/@bytecodealliance/jco` is missing `obj/`, but what's worse is that it's also missing all of the stuff needed to generate it:

```
yowasp/runtime-js/test/yowasp_runtime_test/node_modules/@bytecodealliance/jco (develop)$ npm run build

> @bytecodealliance/jco@1.1.1 build
> cargo xtask build debug

error: no such command: `xtask`

        Did you mean `flash`?

        View all installed commands with `cargo --list`
        Find a package to install `xtask` with `cargo search cargo-xtask`
```

whitequark:
I'm happy with my implementation of YoWASP/runtime-js on top of JCO right now--let me know when you've made a release and I'll flip the switch on all YoWASP projects to use it.

guybedford:
@whitequark thanks for the feedback here. I'm wondering if perhaps we can make this part of the next major and then unify on a new simpler hybrid instantiation interface that supports both async and sync. I've also created a tracking issue in https://github.com/bytecodealliance/jco/issues/441 for ideas around better instantiation APIs. Feedback is very welcome.

whitequark:
> I'm wondering if perhaps we can make this part of the next major and then unify on a new simpler hybrid instantiation interface that supports both async and sync.

That would work for me; I lock JCO at a specific version so the API breakage wouldn't cause a fallout. Thanks!

> I've also created a tracking issue in #441 for ideas around better instantiation APIs. Feedback is very welcome.

I'll have to think about ways to provide custom WASI instantiations.

whitequark:
> I'm wondering if perhaps we can make this part of the next major

Do you know when this could happen? I'd *really* like to use this in most or all of the YoWASP applications (it's a blocker for some important use cases) and unfortunately due to the cargo being a build dependency for jco I would probably have to publish a fork of jco under one of my namespaces in order to use it in the builds of downstream packages (as a git URL would not work).

guybedford:
Sure, let me try and include it in the next release cycle later this month, and if it turns out tricky for some reason I'll share an update here again.

whitequark:
Thanks so much!

PR #422, opened 2024-04-21
fix: clocks type definitions

kateinoigakukun:
The type definitions for the clocks were swapped


guybedford:
Wow, since these are auto-generated this might be a bug in the interface definitions unless there was a manual edit.

PR #423, opened 2024-04-24
Add support for generating types from a `wit` directory

kateinoigakukun:
We use jco to generate TypeScript interface from WIT definitions by using `obj/js-component-bindgen-component.js`. Exposing the ability to generate types from directory would allow us to manage our WIT in multiple files.


guybedford:
Thanks it would be great to get this merged.

We could alternatively handle the directory case when providing a path, we do this for ComponentizeJS already - https://github.com/bytecodealliance/ComponentizeJS/blob/main/crates/spidermonkey-embedding-splicer/src/lib.rs#L79

guybedford:
Thanks again for this - I ended up implementing the suggestion along with exposing the API and CLI usage with tests in https://github.com/bytecodealliance/jco/pull/426.

Further feedback very welcome, do let me know if that works for you.

kateinoigakukun:
Ah, I missed your message, sorry. And thanks for revising this!

PR #430, opened 2024-04-29
fix: support uninitialized resource transfer

guybedford:
This adds support for uninitialized resource table transfer. In rare edge cases where the target table is not otherwise referenced by any bindgen code and the resource is never dropped, it is possible for the target table in transfer to not otherwise be initialized by the component bindgen.

I don't want to land this without creating a simple test, and fully understanding what mechanisms can lead to this though, because it seems very odd to have an undroppable resource.


guybedford:
On further thought, strictly speaking components under these semantics are valid and execute under Wasmtime, therefore we should ensure the same for Jco.

PR #451, opened 2024-06-18
Generate Typescript Declaration Modules and Stubs

nicoburniske:


guybedford:
Lovely to see the testing workflow here. If you'd ever like to consider upstreaming any of this do feel free to as well.

PR #454, opened 2024-06-26
translate errno=-4094 errors to "no-such-device" on Windows

dicej:
When an application attempts to open a non-existent path such as `//foo/bar/baz` on Windows, Node generates an error of the form:

```
{
  errno: -4094,
  syscall: 'open',
  code: 'UNKNOWN',
  path: '//foo/bar/baz'
}
```

Prior to this commit, that error would be re-thrown, causing Wasm execution to trap.  Now we convert it to a "no-such-device" error instead.

### DIFF
--- a/packages/preview2-shim/lib/nodejs/filesystem.js
+++ b/packages/preview2-shim/lib/nodejs/filesystem.js
@@ -700,6 +700,13 @@ function convertFsError(e) {
       return "text-file-busy";
     case "EXDEV":
       return "cross-device";
+    case "UNKNOWN":
+      switch (e.errno) {
+        case -4094:
+          return "no-such-device";
+        default:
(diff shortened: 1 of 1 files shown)


dicej:
Thanks for merging this, @guybedford.  Is there a chance we could publish a new release of the shim with this fix soonish?  That would help unblock https://github.com/dotnet/runtimelab/pull/2614, which is where I encountered this issue.

guybedford:
Sure, I've pushed a 0.16.3.

oovm:
This fix was a bit confusing to me, and it took me a while to find the real reason.

On Windows, the path emitted by std::fs::read will be prefixed with `\\\\C:\`, but node cannot recognize such a path, node open only supports `C:\`

PR #459, opened 2024-07-01
feat: support semver aware exclusions

vados-cosmonic:
This PR updates `jco` to support the `@since`, `@after` and `@unstable` annotations as implemented by the upstream Rust tooling.

Currently there are a few issues left to address:

- [x] ~~We likely need an upstream release & alignment of the new code in `wit-parser` that supports the annotations:~~
  - ~~https://github.com/bytecodealliance/wasm-tools/commit/01bec9c8adb491a591447b52ebf353babcd8b8f1~~
  - ~~https://github.com/bytecodealliance/wasm-tools/commit/10d2e21b2171cf2a2227f9de04ac89985042a5f0~~
- [x] Finish implementing the test that was added (and fill out the checks for present/missing impl)

### DIFF
--- a/crates/js-component-bindgen/src/lib.rs
+++ b/crates/js-component-bindgen/src/lib.rs
@@ -13,14 +13,14 @@ pub use transpile_bindgen::{BindingsMode, InstantiationMode, TranspileOpts};
 use anyhow::Result;
 use transpile_bindgen::transpile_bindgen;
 
-use anyhow::{bail, Context};
+use anyhow::{bail, ensure, Context};
 use wasmtime_environ::component::{ComponentTypesBuilder, Export, StaticModuleIndex};
 use wasmtime_environ::wasmparser::Validator;
 use wasmtime_environ::{PrimaryMap, ScopeVec, Tunables};
 use wit_component::DecodedWasm;
 
 use ts_bindgen::ts_bindgen;
-use wit_parser::{Resolve, Type, TypeDefKind, TypeId, WorldId};
+use wit_parser::{Package, Resolve, Stability, Type, TypeDefKind, TypeId, WorldId};
 
 /// Calls [`write!`] with the passed arguments and unwraps the result.
 ///
@@ -171,4 +171,36 @@ pub fn dealias(resolve: &Resolve, mut id: TypeId) -> TypeId {
             _ => break id,
         }
     }
+}
+
+/// Check if an item (usually some form of [`WorldItem`]) should be allowed through the feature gate
+/// of a given package.
+fn feature_gate_allowed(
+    resolve: &Resolve,
+    package: &Package,
+    stability: &Stability,
+) -> Result<bool> {
+    Ok(match stability {
+        Stability::Unknown => true,
+        Stability::Stable { since, .. } => {
+            let Some(package_version) = package.name.version.as_ref() else {
+                // If the package version is missing (we're likely dealing with an unresolved package)
+                // and we can't really check much.
+                return Ok(true);
+            };
+
+            ensure!(
+                package_version >= since,
+                "feature gates cannot refer to unreleased (future) package versions"
+            );
+
+            // Stabilization (@since annotation) overrides features and deprecation
+            true
+        }
+        Stability::Unstable { feature } => {
+            // If a @unstable feature is present but the related feature was not enabled
+            // or all features was not selected, exclude
+            resolve.all_features || resolve.features.contains(feature)
+        }
+    })
 }
--- a/crates/js-component-bindgen/src/lib.rs
+++ b/crates/js-component-bindgen/src/lib.rs
@@ -67,7 +67,8 @@ pub fn generate_types(
 ) -> Result<Vec<(String, Vec<u8>)>, anyhow::Error> {
     let mut files = files::Files::default();
 
-    ts_bindgen(&name, &resolve, world_id, &opts, &mut files);
+    ts_bindgen(&name, &resolve, world_id, &opts, &mut files)
+        .context("failed to generate Typescript bindings")?;
 
     let mut files_out: Vec<(String, Vec<u8>)> = Vec::new();
     for (name, source) in files.iter() {
@@ -137,7 +138,8 @@ pub fn transpile(component: &[u8], opts: TranspileOpts) -> Result<Transpiled, an
     }
 
     if !opts.no_typescript {
-        ts_bindgen(&name, &resolve, world_id, &opts, &mut files);
+        ts_bindgen(&name, &resolve, world_id, &opts, &mut files)
+            .context("failed to generate Typescript bindings")?;
     }
 
     let (imports, exports) = transpile_bindgen(
@@ -179,6 +181,7 @@ fn feature_gate_allowed(
     resolve: &Resolve,
     package: &Package,
     stability: &Stability,
+    item_name: &str,
 ) -> Result<bool> {
     Ok(match stability {
         Stability::Unknown => true,
@@ -191,7 +194,7 @@ fn feature_gate_allowed(
 
             ensure!(
                 package_version >= since,
-                "feature gates cannot refer to unreleased (future) package versions"
+                "feature gate on [{item_name}] refers to an unreleased (future) package version [{since}] (current package version is [{package_version}])"
             );
 
             // Stabilization (@since annotation) overrides features and deprecation
--- a/crates/js-component-bindgen/src/ts_bindgen.rs
+++ b/crates/js-component-bindgen/src/ts_bindgen.rs
@@ -4,6 +4,7 @@ use crate::names::{is_js_identifier, maybe_quote_id, LocalNames, RESERVED_KEYWOR
 use crate::source::Source;
 use crate::transpile_bindgen::{parse_world_key, InstantiationMode, TranspileOpts};
 use crate::{dealias, feature_gate_allowed, uwrite, uwriteln};
+use anyhow::{Context as _, Result};
 use heck::*;
 use log::debug;
 use std::collections::btree_map::Entry;
@@ -51,7 +52,7 @@ pub fn ts_bindgen(
     id: WorldId,
     opts: &TranspileOpts,
     files: &mut Files,
-) {
+) -> Result<()> {
     let mut bindgen = TsBindgen {
         src: Source::default(),
         interface_names: LocalNames::default(),
@@ -66,9 +67,9 @@ pub fn ts_bindgen(
         .get(
             world
                 .package
-                .expect("unexpectedly missing package in world"),
+                .context("unexpectedly missing package in world")?,
         )
-        .expect("unexpectedly missing package in world for ID");
+        .context("unexpectedly missing package in world for ID")?;
 
     {
         let mut funcs = Vec::new();
@@ -76,8 +77,8 @@ pub fn ts_bindgen(
         for (name, import) in world.imports.iter() {
             match import {
                 WorldItem::Function(f) => {
-                    if !feature_gate_allowed(resolve, package, &f.stability)
-                        .expect("failed to check feature gate for imported function")
+                    if !feature_gate_allowed(resolve, package, &f.stability, &f.name)
+                        .context("failed to check feature gate for imported function")?
                     {
                         debug!("skipping imported function [{}] feature gate due to feature gate visibility", f.name);
                         continue;
@@ -89,8 +90,13 @@ pub fn ts_bindgen(
                     }
                 }
                 WorldItem::Interface { id, stability } => {
-                    if !feature_gate_allowed(resolve, package, &stability)
-                        .expect("failed to check feature gate for imported interface")
+                    let iface_name = &resolve.interfaces[*id]
+                        .name
+                        .as_ref()
+                        .map(String::as_str)
+                        .unwrap_or("<unnamed>");
+                    if !feature_gate_allowed(resolve, package, &stability, iface_name)
+                        .context("failed to check feature gate for imported interface")?
                     {
                         let import_specifier = resolve.id_of(*id).unwrap();
                         let (_, _, iface) = parse_world_key(&import_specifier).unwrap();
@@ -124,8 +130,8 @@ pub fn ts_bindgen(
                     let ty = &resolve.types[*tid];
                     let name = ty.name.as_ref().unwrap();
 
-                    if !feature_gate_allowed(resolve, package, &ty.stability)
-                        .expect("failed to check feature gate for imported type")
+                    if !feature_gate_allowed(resolve, package, &ty.stability, name)
+                        .context("failed to check feature gate for imported type")?
                     {
                         debug!("skipping imported type [{name}] feature gate due to feature gate visibility");
                         continue;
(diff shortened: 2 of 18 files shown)


yoshuawuyts:
@vados-cosmonic incredible; thank you for working on this!

> We likely need an upstream release & alignment of the new code in wit-parser that supports the annotations

There should already be a working upstream implementation pulled in. I checked with wasmtime to make sure that a working `wasm-tools` version was part of the latest release, specifically so that that `jco` could pull it in. If that's not working somehow, maybe @guybedford will know more?

vados-cosmonic:
Hey @yoshuawuyts thanks for chiming in, so it turns out I was using the git dep for `wasmtime-tools` which was my problem! 

`wasmtime-environ` (the last git dep) *does* need a release, but `wasm-tools` doesn't, I was able to remove that git dep!

yoshuawuyts:
Oh, rightt - yeah that's the "semver-aware exports" feature wasmtime was missing and still needed to implement. My assumption was (and this might be wrong) that both `jco` and `wasmtime` would resolve this issue in separate ways. But if `jco` uses `wasmtime-environ` for its resolution, it makes sense that we'd want to reuse it.

I'm not sure what the exact right path here will be. What I do know is that Wasmtime 0.23 will be released in two weeks - which is about a week before we're set to release WASI 0.2.1. So one option might be to have a patch ready which we can merge the moment a new wasmtime version is released, making it part of the jco release a few days later.

That does cut it a little close for my taste, so ideally we could land + test the patch well before then. I don't know what our options beyond that would be? - Floating patches? Pulling a non-release version of wasmtime? Ask wasmtime to pretty please push a patch version of `wasmtime-environ` with these changes? Implement the resolver logic ourselves on top? I don't know - but I think this is probably a @guybedford question.

vados-cosmonic:
Ahhh thanks for laying this out -- this helped me understand a lot more of the high level plan -- happy to hear how best to move forward here. IIRC @ricochet is also capable of helping with the upstream release if necessary here.

Right now with the code in this branch `wit-parser` still doesn't seem to recognize `@since`, which has me a bit confused. Maybe I'm missing something even after setting [`Resolve`'s `all_features`](https://docs.rs/wit-parser/latest/wit_parser/struct.Resolve.html#structfield.features)...

test/wit.js
```diff
@@ -0,0 +1,82 @@
...
+          runtimeName,
+          "-o",
+          outDir
+        );
+        strictEqual(stderr, "");
+      }
+
+      console.log(`OUTPUT to [test/output/${runtimeName}.js]`);
+
+      // TODO(remove): probably don't need to execute it!
```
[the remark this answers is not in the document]
vados-cosmonic: Removed -- Thanks for the input here!

test/wit.js
```diff
@@ -0,0 +1,82 @@
...
+        "test/fixtures/wit",
+        "--world-name",
+        "test:jco-feature-gates/imported",
+        "-o",
+        outFile
+      );
+      strictEqual(stderr, "");
+
+      // TODO: check the generated code to ensure that
+      // no featuregate related functions are present
```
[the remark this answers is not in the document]
vados-cosmonic: Thanks for pointing this out -- I switched to doing this and it's much easier to test.

I've checked the exports, but they seem to be lists of interfaces, so I moved to checking the file output (in particular the interfaces)

test/wit.js
```diff
@@ -0,0 +1,82 @@
...
+     * Ensure feature gates work properly
+     * 
+     * see: https://github.com/WebAssembly/component-model/blob/main/design/mvp/WIT.md#feature-gates
+     */
+    test("feature-gates", async () => {
+      const runtimeName = "semver";
+
+      const { stderr } = await exec(
+        jcoPath,
+        "componentize",
```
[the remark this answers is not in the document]
vados-cosmonic: Thanks! Switched to using this `componentEmbed` and `componentNew` instead!

crates/js-component-bindgen/src/transpile_bindgen.rs
```diff
@@ -458,10 +458,24 @@ impl<'a> Instantiator<'a, '_> {
...
+            .packages
+            .get(
+                self.resolve.worlds[self.world]
+                    .package
+                    .expect("unexpectedly missing package in world"),
+            )
+            .expect("unexpectedly missing package in world for ID");
+        let package_version = package.name.version.as_ref();
+
+        // Iterate & generate exports
```
guybedford, line 473: This code isn't doing export generation, it is doing a pre-pass to populate resource data structures.

crates/js-component-bindgen/src/transpile_bindgen.rs
```diff
@@ -511,16 +552,45 @@ impl<'a> Instantiator<'a, '_> {
                 },
             }
         }
+
         self.exports_resource_types = self.imports_resource_types.clone();
-        for (key, item) in &self.resolve.worlds[self.world].exports {
+
+        // Iterate & generate exports
```
guybedford, line 558: Same again here, both this and the above are not the right place to insert the stability gating.

guybedford (commented):
The feature gating needs to be moved into the bindgen functions for:

* `exports` in https://github.com/bytecodealliance/jco/blob/main/crates/js-component-bindgen/src/transpile_bindgen.rs#L1740
* `imports` in https://github.com/bytecodealliance/jco/blob/main/crates/js-component-bindgen/src/transpile_bindgen.rs#L1057

For the TypeScript generation, it looks like you've got per-interface filtering of functions and also the filtering of exports and imports. While all interfaces may still be written, that seems like it should be roughly correct? Would be good to clearly enumerate what cases are missing.

crates/js-component-bindgen/src/transpile_bindgen.rs
vados-cosmonic: Ahh! Thank you -- I actually totally forgot my own `TODO` in the that other bit of the file!

crates/js-component-bindgen/src/transpile_bindgen.rs
vados-cosmonic: Moved -- thanks!

vados-cosmonic:
So the upstream PR is still ongoing, *but* the tests here now reflect at least some of the intended flow. 

Since it looks like feature gated items with a version *newer* than the package itself won't be allowed in the future, we actually don't have to test `@since(version = v, feature = f)` with a missing feature `f` (it won't matter).

vados-cosmonic:
Ah looks like there are some [other errors due to the updated code](https://github.com/bytecodealliance/jco/actions/runs/10063010856/job/27816914733?pr=459) -- will look into these.

crates/js-component-bindgen/src/lib.rs
```diff
@@ -172,3 +172,35 @@ pub fn dealias(resolve: &Resolve, mut id: TypeId) -> TypeId {
...
+) -> Result<bool> {
+    let Some(package_version) = package.name.version.as_ref() else {
+        // If the package version is missing (we're likely dealing with an unresolved package)
+        // and we can't really check much.
+        return Ok(true);
+    };
+
+    Ok(match stability {
+        Stability::Unknown => true,
+        Stability::Stable { since, .. } => {
```
guybedford, line 191: We need to also check the `feature` here as well.

guybedford (commented):
This looks great. Can we just add two extra tests here:

1. Replacing `package test:feature-gates@0.2.1` in the package with `package test:feature-gates@0.2.0`, and verifying that `b` and `c` are not output, even with `features: ['gated']` set (which I think requires the change I've commented on in the code).
2. Running with an explicit feature enabling `features: ['fancier-foo']` and verifying it turns on `d` (currently just `features: ['all']` is tested)

It seems to me like these should be relatively straightforward to add, but would be good to verify!

crates/js-component-bindgen/src/lib.rs
vados-cosmonic: If I'm understanding this right you mean to use that to determine visibility right? Originally I thought so too -- but after the [discussion upstream](https://github.com/WebAssembly/component-model/issues/382) the consensus was that once a feature is *stable*, the written "feature" is actually just descriptive. 

Ideally I'd just reuse the function from [the `wasm-tools` PR](https://github.com/bytecodealliance/wasm-tools/pull/1689) in `witparser` here but it's not released just yet.

vados-cosmonic: Also, with regards to #1, trying to refer to a future package (i.e. `@since( version = 0.2.1)`, when the package is `0.2.0` is actually a bug -- wasmparser wouldn't output that, so I could add a test for the failure but not that the specific items were missing.

vados-cosmonic:
Thanks for review @guybedford ! I did #2, but please see https://github.com/bytecodealliance/jco/pull/459#discussion_r1697436985, I don't think I can actually do #1 (or at least it would be a different kind of test)

test/wit.js
```diff
@@ -0,0 +1,158 @@
...
+      ]);
+      ok(files['component.js'], "component js was generated");
+      ok(files['component.d.ts'], "component typings were generated");
+      ok(files['interfaces/test-feature-gates-foo.d.ts'], "interface typings were generated");
+
+      // Check the interfaces file for the right exports
+      const interfaces = Buffer.from(files['interfaces/test-feature-gates-foo.d.ts']).toString('utf8');
+      ok(interfaces.includes("export function a(): void;"), "unconstrained export foo/a is present");
+      ok(interfaces.includes("export function b(): void;"), "@since(0.2.1) export foo/b is present (version matches)");
+      ok(interfaces.includes("export function c(): void;"), "@since(0.2.1) export foo/c is present (feature is ignored)");
```
guybedford, line 78: This one feels wrong to me - when a feature is not enabled, even if the version matches it should surely still be a requirement that the feature is explicitly enabled?

test/wit.js
```diff
@@ -0,0 +1,158 @@
...
```
guybedford, line 153: Again this one seems like it shouldn't be included?

guybedford:
Also strictly speaking, according to the interpretation provided for since gates, we surely don't need any semver matching at all if all since gates will always be of versions less than the version we are checking?

That is, if for a package version `v`, all since gates are guaranteed to have version `s` where `s < v`, then we don't need to ever check `s < v`.

crates/js-component-bindgen/src/lib.rs
vados-cosmonic: Ah yep, so basically I think the *intended* progression is that once a feature goes from `@unstable(feature = ...)` to `@since(version = X.X.X)` it is "stable" and is basically always included -- the "feature" annotation doesn't really do anything at that point.

guybedford:
Okay this is making sense now, and I've resolved the outstanding questions.

So the only capability provided by this PR is the ability to disable features in the bindgen.

Last question then - if a component was generated assuming that a feature is enabled, is it actually valid to bindgen with that same feature disabled? That is, shouldn't the features that the component was built with be something we extract from the component itself as opposed to being an input to just the bindgen?

test/wit.js
vados-cosmonic: Yeah `c()` was `feature = gated`, right? But due to `@since(version = 0.2.1)` it's "stabilized" which means it should be visible... 

AFAIK this is working as intended, with the main level for visibility being the `@unstable` -> `@since` transition

I'll bring this up during the CM meeting tomorrow

test/wit.js
vados-cosmonic: Oh and just to note here -- I'll bring this up in the upcoming CM meeting

vados-cosmonic:
Sorry super late but forgot to properly respond to this:

> Also strictly speaking, according to the interpretation provided for since gates, we surely don't need any semver matching at all if all since gates will always be of versions less than the version we are checking?
>
> That is, if for a package version v, all since gates are guaranteed to have version s where s < v, then we don't need to ever check s < v.

This is... definitely true. The `version` specifier might actually be strictly informational since `wit-parser` would have blown up before hand.

And then

> So the only capability provided by this PR is the ability to disable features in the bindgen.

Yeah, basically support ensuring things marked `@unstable` actually don't show up.

> Last question then - if a component was generated assuming that a feature is enabled, is it actually valid to bindgen with that same feature disabled? That is, shouldn't the features that the component was built with be something we extract from the component itself as opposed to being an input to just the bindgen?

Hmnnn that's an interesting question -- I'm defaulting to the user needing to supply the feature(s) on every related tool, but that's obviously not a great answer.

Would you mind laying out a scenario? This is probably *also* worth discussing (kind of like the packaging implications too...)

guybedford:
Say a component was built with an imported feature enabled, then that package will be written and authored assuming it can access that feature function or interface import.

To just not then bindgen that import, would result in a Wasm core error that the import is not defined by the bindgen.

This makes me think, perhaps the right interpretation here is that imports should always be based on whatever the component itself _uses_? Rather than having to explicitly select features through an option? Because it feels like it will be a bad UX for imports if you have to guess what the component author intended, rather than just going with that to start.

Then maybe the feature filtering per this PR is only an exports filtering?

guybedford:
Of course for `jco types` we definitely do want type generation to support feature stripping, since this is done without a specific component implementation.

guybedford:
@vados-cosmonic agreed, it seems like if any of these gates ever fail it's effectively an invalid component. And that this validation might belong further up the stack.

There might be a use case for when a dummy component is generated to provide features for that dummy component. Although perhaps that is something that belongs in wasm-tools again.

Overall I'm also tending towards saying this should just be a features integration for `jco types`, although let me know if you have any further counter arguments to share too.

vados-cosmonic:
Definitely all for making this a `jco types` feature -- `ts_bindgen.rs` is now the only file with explicit feature gate checking

Thanks again for the patience and the reviews!

crates/js-component-bindgen/src/lib.rs
```diff
@@ -172,3 +172,35 @@ pub fn dealias(resolve: &Resolve, mut id: TypeId) -> TypeId {
...
+            let Some(package_version) = package.name.version.as_ref() else {
+                // If the package version is missing (we're likely dealing with an unresolved package)
+                // and we can't really check much.
+                return Ok(true);
+            };
+
+            ensure!(
+                package_version >= since,
+                "feature gates cannot refer to unreleased (future) package versions"
+            );
```
guybedford, line 195: For this error, in case it happens to someone for whatever reason, can we include the name of the function that it applies to, by adding a `debug_name` or otherwise to this function?

crates/js-component-bindgen/src/ts_bindgen.rs
```diff
@@ -154,6 +199,14 @@ pub fn ts_bindgen(
                         (iface_id.as_ref(), iface)
                     }
                 };
+
+                if !feature_gate_allowed(resolve, package, &stability)
+                    .expect("failed to check feature gate for export")
```
guybedford, line 204: Instead of a panic, could this be an error?

guybedford (changes requested):
Could we also add the `--features` flag option to the `jco types` command, should be relatively straightforward I think?

Then otherwise just a couple of last questions on the error reporting.

crates/js-component-bindgen/src/lib.rs
vados-cosmonic: Absolutely -- that would improve ease of debugging -- I'll add the versions as well.

- [ ] Add item being checked to stability fn, show versions that conflict

crates/js-component-bindgen/src/ts_bindgen.rs
vados-cosmonic: Ah so the reason this is an `expect` is because `ts_bindgen()` doesn't return a result -- I'll change the signature here and update the rest.

- [ ] change `ts_bindgen` to return a `Result`

vados-cosmonic:
Hey @guybedford thanks again for the careful review -- I made the changes and added some tests around `jco types` for feature use!

Finally got all the tests passing :sweat_smile:

guybedford (approved):
Perfect, thank you for carrying this one over the line!

PR #461, opened 2024-07-03
allow port numbers up to 65535 in `OutgoingRequest.setAuthority`

### LINKED ISSUE #460: [preview2-shim] OutgoingRequest.setAuthority rejects ports >9999
I'm using `jco` to run tests for .NET's new WASIp2 support, and have found that calling `set-authority` on an `outgoing-request` with a port number greater than 9999 fails: https://github.com/bytecodealliance/jco/blob/d7ec53f02ce12753dcc6f7f66ecd242259b99912/packages/preview2-shim/lib/nodejs/http.js#L265.  That's a problem for my tests, because I'm using Node's `http.Server.listen` with a port of zero (i.e. let the system select an unused port, which normally results in a port higher than 9999).

What's the rationale for that restriction?  The docs say `set-authority` should only return an error if the specified string is syntactically invalid: https://github.com/WebAssembly/wasi-http/blob/a81c61fc584278f801393b22289840a439e51f50/wit/types.wit#L308-L312, which doesn't seem to apply here.

dicej:
Fixes #460

### DIFF
--- a/packages/preview2-shim/lib/nodejs/http.js
+++ b/packages/preview2-shim/lib/nodejs/http.js
@@ -262,7 +262,7 @@ class OutgoingRequest {
       if (
         extra.length ||
         (port !== undefined &&
-          (portNum.toString() !== port || portNum > 9999)) ||
+          (portNum.toString() !== port || portNum > 65535)) ||
         !host.match(/^[a-zA-Z0-9-.]+$/)
       )
         throw undefined;


dicej:
Thanks for merging this @guybedford!  If it's not a pain, would you mind publishing a new release?  That would unblock some of my tests for .NET the WASIp2 port.

guybedford:
Sure no problem @dicej I've posted a 0.16.4.

Issue #472, opened 2024-07-22
Wrong function type for `instantiate` with `InstantiationMode::Async`


wooden-worm:
The type for `getCoreModule` should be `getCoreModule: (path: string) => WebAssembly.Module |  Promise<WebAssembly.Module>` instead of just `getCoreModule: (path: string) => WebAssembly.Module`

https://github.com/bytecodealliance/jco/blob/7f1f11e00a2a61f8bfa80ed1440722e3289a48a4/crates/js-component-bindgen/src/ts_bindgen.rs#L256

guybedford:
Thanks, yes this seems correct. Would you like to post a PR?

PR #473, opened 2024-07-22
Fix: `getCoreModule` type for async `instantiate`

wooden-worm:
Fixes https://github.com/bytecodealliance/jco/issues/472

In async mode, `getCoreModule` should have `Promise<WebAssembly.Module>` return type.

### DIFF
--- a/crates/js-component-bindgen/src/ts_bindgen.rs
+++ b/crates/js-component-bindgen/src/ts_bindgen.rs
@@ -254,7 +254,7 @@ pub fn ts_bindgen(
                         instantiateCore?: (module: WebAssembly.Module, imports: Record<string, any>) => WebAssembly.Instance
                     ): {camel};
                     export function instantiate(
-                        getCoreModule: (path: string) => WebAssembly.Module,
+                        getCoreModule: (path: string) => WebAssembly.Module | Promise<WebAssembly.Module>,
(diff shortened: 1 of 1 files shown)


crates/js-component-bindgen/src/ts_bindgen.rs
```diff
@@ -254,7 +254,7 @@ pub fn ts_bindgen(
                         instantiateCore?: (module: WebAssembly.Module, imports: Record<string, any>) => WebAssembly.Instance
                     ): {camel};
                     export function instantiate(
```
wooden-worm, line 256: Not an expert in TS, I'm not sure why we need 2 function signatures here (this one and the above). Should one be enough? And should just `Promise` be enough?

```typescript
                    export function instantiate(
                        getCoreModule: (path: string) => Promise<WebAssembly.Module>,
                        imports: ImportObject,
                        instantiateCore?: (module: WebAssembly.Module, imports: Record<string, any>) =>  Promise<WebAssembly.Instance>
                    ): Promise<{camel}>;
```

guybedford: It's an overload of the function in the fully sync case to note that if both functions are sync the return value is always known synchronously.

So I think this PR is correct but thanks for checking on this.

PR #479, opened 2024-08-06
upgraded wasm-tools

thesuhas:
Updating `wasm-tools` dependencies as downstream tools such as [ComponentizeJS](https://github.com/bytecodealliance/ComponentizeJS/tree/main) have issues when you update `wasm-tools` due to upstream tools having different variants.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 .vscode
+.idea
 node_modules
 /target
 /obj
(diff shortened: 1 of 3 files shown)


guybedford:
We update wasm-tools monthly with Wasmtime releases, so the next one will be in two weeks. We can't really race the Wasmtime releases unforutnately. In the mean time feel free to use a fork if you need to upgrade ComponentizeJS.

PR #481, opened 2024-08-08
fix: Wrap lists in Array to handle optional types

### LINKED ISSUE #480: Wrong typescript generated for `list<option<something>>`
Currently, the wrong typescript types are generated when trying to use a WIT type like `list<option<something>>` because it outputs `Something | undefined[]`.

Here's the WIT that I wrote for my reproduction:
```wit
package test:issue;

interface types {
  variant value {
    foo,
    bar,
    baz
  }

  foobarbaz: func() -> list<option<value>>;
}

world issue {
  import types;
}
```

I think that the generated code should just use `Array<Something | undefined>` which avoids the need to wrap everything in parens.

I'll be submitting the PR to fix this in a minute but I wanted an issue number to name the test fixtures.

blaine-arcjet:
Closes #480 

This changes the types generated for a WIT `list` to use `Array<>` syntax, which allows for something like `option<foo>` to be expanded to `foo | undefined` without generating the wrong type signature. Using `Array<>` also avoids the need to wrapping too much stuff in parens.

### DIFF
--- a/crates/js-component-bindgen/src/ts_bindgen.rs
+++ b/crates/js-component-bindgen/src/ts_bindgen.rs
@@ -638,8 +638,9 @@ impl<'a> TsInterface<'a> {
             Some("Uint8Array") => self.src.push_str("Uint8Array"),
             Some(ty) => self.src.push_str(ty),
             None => {
+                self.src.push_str("Array<");
                 self.print_ty(ty);
-                self.src.push_str("[]");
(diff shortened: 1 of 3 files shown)


phated:
@guybedford I need this at work. Can we get this merged and released 🙏❤️

guybedford:
Thanks for the fix, will post out a release soon.

Issue #497, opened 2024-09-05
preview2-shim: `TypeError: Cannot read properties of null (reading 'startsWith')` on browser


kateinoigakukun:
## Minimum repro

```rust
fn main() {
    use std::fs;
    let metadata = fs::metadata(".");
    println!("{:?}", metadata);
}
```

```console
$ rustc check.rs --target wasm32-wasip1 -o check.core.wasm
$ npx jco new --wasi-command ./check.core.wasm -o check.wasm
$ npx jco transpile check.wasm -o component
```

```html
<html>
  <script type="importmap">
    {
      "imports": {
        "@bytecodealliance/preview2-shim/filesystem": "https://cdn.jsdelivr.net/npm/@bytecodealliance/preview2-shim@0.16.5/lib/browser/filesystem.js",
        "@bytecodealliance/preview2-shim/io": "https://cdn.jsdelivr.net/npm/@bytecodealliance/preview2-shim@0.16.5/lib/browser/io.js",
        "@bytecodealliance/preview2-shim/cli": "https://cdn.jsdelivr.net/npm/@bytecodealliance/preview2-shim@0.16.5/lib/browser/cli.js"
      }
    }
  </script>
  <script type="module">
    import { run } from './component/check.js';
    run.run();
  </script>
</html>
```

Then we get the following:

```
TypeError: Cannot read properties of null (reading 'startsWith')
    at getChildEntry (filesystem.js:33:17)
    at Descriptor.statAt (filesystem.js:201:19)
    at trampoline13 (check.js:1063:57)
    at wit-component:shim.indirect-wasi:filesystem/types@0.2.0-rc-2023-11-10-[method]descriptor.stat-at (wit-component:shim-f0b2f95a:0x11f)
    at wit-component:adapter:wasi_snapshot_preview1._ZN22wasi_snapshot_preview18bindings4wasi10filesystem5types10Descriptor7stat_at17hba52cb8364bf76aaE (check.core2.wasm:0x291e)
    at wit-component:adapter:wasi_snapshot_preview1.path_filestat_get (check.core2.wasm:0x21f8)
    at wit-component:shim.adapt-wasi_snapshot_preview1-path_filestat_get (wit-component:shim-f0b2f95a:0x1a1)
    at wasi::lib_generated::path_filestat_get::h0652677c656d456f (lib_generated.rs:1739)
[...]

PR #499, opened 2024-09-08
Set `/` as the default initial working directory for browser p2 shim

kateinoigakukun:
Close https://github.com/bytecodealliance/jco/issues/497

I'll try to make time to set up more test suites for the browser shim, but for now, just make it work with ruby.


guybedford:
Thanks for posting, I'll post out the patch shortly.

PR #503, opened 2024-09-20
Typescript Stubgen Command

nicoburniske:
> Finally upstreaming what was originally done here https://github.com/bytecodealliance/jco/pull/451

Introduces new command `stubgen` that will create all typescript declaration files from the corresponding wit types. There are rust tests for some of the complex cases

- WIT exports get a stub interface type
- Each WIT import interface gets its own `{namespace}{package}{interface}.d.ts` file

## WIT Exports
This simple wit, will generate the following "stub" type for the user to then implement and export from their component

```wit
package test:basic;

interface greet {
    say-hello: func(name: string) -> string;
}

world test {
    export greet;
}
```

#### test.d.ts

```ts
export interface Greet {
    sayHello(name: string): string,
}

export interface TestWorld {
    greet: Greet,
}
```

So then developers can import the types, and implement them. 

```ts

import { TestWorld, Greet } from './generated/test';

export const greet: Greet = {
  sayHello: (name: string) => {
    return `Hello, ${name}!`;
  }
};

```

## WIT Imports

```wit
package test:types;

interface types {
    type dimension = u32;
    record point {
        x: dimension,
        y: dimension,
    }
}
```

```wit
package test:canvas;

interface canvas {
    use test:types/types.{dimension, point};
    type canvas-id = u64;
    draw-line: func(canvas: canvas-id, origin: point, target: point, thickness: dimension);
}
```

This would generate two typescript files, one for each interface
#### interfaces/test-canvas-canvas.d.ts

```ts
declare module "test:canvas/canvas" {
    import type { Dimension } from "test:types/types";
    import type { Point } from "test:types/types";
    export type CanvasId = bigint;
    export function drawLine(canvas: CanvasId, origin: Point, target: Point, thickness: Dimension): void;
} 
```
#### interfaces/test-types-types.d.ts

```ts
declare module "test:types/types" {
    export type Dimension = number;
    export interface Point {
        x: Dimension,
        y: Dimension,
    }
}


guybedford (commented):
This is fantastic thanks for working on upstreaming this and providing such comprehensive tests as well. It would be great to provide for ComponentizeJS / StarlingMonkey workflows.

Some initial design space questions:

* Would it make sense to extend this to full JS boilerplate creation, ie creating the JS files as well for stub implementations to fill in, possibly with // TODO comments or otherwise? It seems the missing piece remains importing the types and setting up the boilerplate and perhaps as we think about increasing automation in dev workflows this is actually what users want?
* Instead of `interfaces/namespace-package-interface.d.ts` perhaps we should consider the structure `interfaces/namespace/package/interface.d.ts` since we can do folder nesting?

I will aim to do a full review soon, thanks for your patience!

guybedford:
We now have a `jco guest-types` that does this.

PR #518, opened 2024-10-28
DRAFT: Adding JSPI and Asyncify

calvinrp:
Adds options to `jco transpile` and `jco types`:

```javascript
.addOption(new Option('--async-mode [mode]', 'use async imports and exports').choices(['sync', 'jspi', 'asyncify']).preset('sync'))
.option('--default-async-imports', 'default async component imports from WASI interfaces')
.option('--default-async-exports', 'default async component exports from WASI interfaces')
.option('--async-imports <imports...>', 'async component imports (examples: "wasi:io/poll@0.2.0#poll", "wasi:io/poll#[method]pollable.block")')
.option('--async-exports <exports...>', 'async component exports (examples: "wasi:cli/run@#run", "handle")')
```

Which treats the specified imports and exports as `async` using either `JSPI` or `Asyncify`.

JSPI is behind a feature flag in Chrome, Deno and NodeJS. Can run via:
```bash
DENO_V8_FLAGS="--experimental-wasm-jspi" deno
```
```bash
node --experimental-wasm-jspi
```

From my understanding, `Asyncify` does not support reentrancy (so it throws an error) but JSPI does.


crates/js-component-bindgen-component/wit/js-component-bindgen.wit
```diff
@@ -62,6 +62,24 @@ world js-component-bindgen {
...
+
+    /// Configure whether to use `async` imports or exports with
+    /// JavaScript Promise Integration (JSPI) or Asyncify.
+    async-mode: option<async-mode>,
+  }
+
+  record async-imports-exports {
+    imports: list<string>,
+    exports: list<string>,
+  }
```
guybedford, line 74: Since components can contain multiple sub-components, this is currently joining imports and exports across all sub-components.

If we only consider the outer component import and exports, then we should restrict to that, but we can't do that either since internal components must be async if the outer component is!

**Therefore I think we should probably restrict this to outer component imports only**

And we should make exports all async as soon as we go into asyncify mode. This also will then help avoid the easy footgun of inadvertantly not making an export async, and then finding it calls an async import down the line which would trap.

crates/js-component-bindgen/src/function_bindgen.rs
```diff
@@ -1048,7 +1050,25 @@ impl Bindgen for FunctionBindgen<'_> {
...
+                        uwriteln!(
+                            self.src,
+                            "await asyncifyWrapExport({})({});",
+                            self.callee,
+                            operands.join(", ")
+                        );
+                    } else {
+                        uwriteln!(
+                            self.src,
+                            "await WebAssembly.promising({})({});",
```
guybedford, line 1064: Ideally this wrapper and the one above should be applied before we even get to the `callee` so that the WebAssembly.promising wrapper is cached at initialization time. Basically where the callee is first constructed.

It might be possible to just change top-level function definitions like the following:

```js
function binding () { .... }
```

to

```js
const binding = WebAssembly.promising(function binding () { ... });
```

crates/js-component-bindgen/src/function_bindgen.rs
```diff
@@ -1066,15 +1086,16 @@ impl Bindgen for FunctionBindgen<'_> {
 
             Instruction::CallInterface { func } => {
                 let results_length = func.results.len();
+                let if_async_await = if self.is_async { "await " } else { "" };
```
guybedford, line 1089:
```suggestion
                let maybe_await = if self.is_async { "await " } else { "" };
```

crates/js-component-bindgen/src/transpile_bindgen.rs
```diff
@@ -546,6 +583,103 @@ impl<'a> Instantiator<'a, '_> {
...
     }
     fn instantiate(&mut self) {
+        if self.use_asyncify {
+            let (if_async, if_async_await) =
+                if let Some(InstantiationMode::Sync) = self.gen.opts.instantiation {
+                    ("", "")
+                } else {
+                    ("async ", "await ")
+                };
+            uwrite!(self.src.js, "
```
guybedford, line 593: All of the functions and values written here should be moved into the `src/intrinsics.rs` file, rather than here on instantiation. It should be possible to write these functions to take all arguments explicitly instead of relying on closures? Or let me know if I'm missing something.

Not urgent of course, but noting now.

crates/js-component-bindgen/src/transpile_bindgen.rs
```diff
@@ -1894,14 +2069,30 @@ impl<'a> Instantiator<'a, '_> {
...
+                .find('@')
+                .map(|i| {
+                    self.async_exports.contains(&format!(
+                        "{}#{}",
+                        export_name.get(0..i).unwrap(),
+                        func.name
+                    ))
+                })
+                .unwrap_or(false);
+        let if_async = if is_async { "async " } else { "" };
```
guybedford, line 2086:
```suggestion
        let maybe_async = if is_async { "async " } else { "" };
```

src/cmd/transpile.js
```diff
@@ -14,6 +14,22 @@ import { platform } from 'node:process';
...
+const DEFAULT_ASYNC_IMPORTS = [
+  "wasi:io/poll#poll",
+  "wasi:io/poll#[method]pollable.block",
+  "wasi:io/streams#[method]input-stream.blocking-read",
+  "wasi:io/streams#[method]input-stream.blocking-skip",
+  "wasi:io/streams#[method]output-stream.blocking-flush",
+  "wasi:io/streams#[method]output-stream.blocking-write-and-flush",
+  "wasi:io/streams#[method]output-stream.blocking-write-zeroes-and-flush",
+  "wasi:io/streams#[method]output-stream.blocking-splice",
+];
```
guybedford, line 26: It could be nice to automatically include a preview2 implementation designed for async now, and run the preview2 tests against that.

One issue here is how to get the async imports, but an easy option might be to just use a convention like:

```js
import ... from '@bytecodealliance/preview2-shims/async/io'`;
```

This would be amazing to explore further...

guybedford (commented):
Very exciting to see this, thank you.

Left some initial comments.

crates/js-component-bindgen/src/transpile_bindgen.rs
calvinrp: Fixed

crates/js-component-bindgen/src/function_bindgen.rs
calvinrp: fixed

src/cmd/transpile.js
calvinrp: Yep, I have a working draft of WASI P2 shims that I have been testing against. But was thinking of breaking into a separate PR.

crates/js-component-bindgen/src/transpile_bindgen.rs
calvinrp: Moved to the intrinsics

crates/js-component-bindgen/src/function_bindgen.rs
calvinrp: Fixed

calvinrp:
Other PRs landed or work in progress that replaces this one

PR #522, opened 2024-11-08
feat: allow custom weval bin

vados-cosmonic:
This commit enables using a custom `weval` binary from `jco` (see also https://github.com/bytecodealliance/ComponentizeJS/pull/156) -- the idea is to hopefully download the weval binary separately from the actual test (i.e. where `jco componentize` gets run), and be able to use it w/ jco specifically via arg & env var from CI.

[Example failure](https://github.com/bytecodealliance/jco/actions/runs/11729397309/job/32675027786#step:7:134)


.nvmrc
```diff
@@ -0,0 +1 @@
+v20.18.0
```
guybedford, line 1: Again, not clear on the use case here?

src/jco.js
```diff
@@ -34,6 +34,7 @@ program.command('componentize')
   .requiredOption('-w, --wit <path>', 'WIT path to build with')
   .option('-n, --world-name <name>', 'WIT world to build')
   .option('--aot', 'Enable Weval AOT compilation of JS')
+  .option('--weval-bin <path>', 'Specify a custom weval binary to use')
```
guybedford, line 37: I would have expected we'd need to set this somewhere in this PR to use it?

.nvmrc
vados-cosmonic: Ah sorry -- this was just so I could use `nvm`/know the version of node that `jco` is expected to work with for development! I can remove if we'd rather not check something like this in!

src/jco.js
vados-cosmonic: Yes -- this PR actually isn't *quite* ready yet -- I need to make sure that installing `weval` with https://github.com/jaxxstorm/action-install-gh-release actually works. Will move this to draft while I work on the CI bits. 

The place this has to go is in the CLI test where we call `componentize` -- just haven't added it yet!

.nvmrc
guybedford: Yeah Node majors happen every 6 months so it's important to get early testing in, and something like this might inhibit noticing changes.

vados-cosmonic: What do you think about bumping the version? We can bump it there and in CI as versions change/update etc.

guybedford: It's easy to forget to do it is the problem. That's why CI tests latest by default.

I guess my question is what is the benefit of locking the version? Node.js has strong backwards compat.

vados-cosmonic: Yeah this is a reasonable point -- I'm definitely open to taking this out and just having the `latest` be the implied default, with explicit testing for each in CI.

> I guess my question is what is the benefit of locking the version? Node.js has strong backwards compat.

My thought was to make it more consistent for people contributing, *BUT* I agree with your point -- if we're basically expecting that our backwards compat tracks/inherits Node's, it really shouldn't matter!

I'll back this out!

src/jco.js
vados-cosmonic: Added and in-use now w/ the `--aot` option!

guybedford:
I've merged https://github.com/bytecodealliance/jco/pull/527 which should resolve the CI error here, updating the branch and rerunning now.

.github/workflows/main.yml
```diff
@@ -45,6 +45,15 @@ jobs:
             macos-latest
         ]
         node: [18.x, 20.x, latest]
+        weval-version:
+          - 'v0.3.2'
```
guybedford, line 49: My concern here is the same as on the other CI change - this is a maintenance burden now, separately to our ComponentizeJS Weval version.

I'm not sure how to square that?

vados-cosmonic: I can get the latest weval version dynamically -- multiple ways I can do that but I'll do it with `curl` (turns out if you try to hit `https://github.com/bytecodealliance/weval/releases/latest` you get redirected to the latest release!)

vados-cosmonic: Done! Weval version is now being downloaded dynamically

Issue #508, opened 2024-10-14
support for top-level WIT resources


ctaggart:
I would really love to see support for WIT resources https://component-model.bytecodealliance.org/design/wit.html#resources . I tried to use them and received the "not yet implemented" error. My `cowsay.wit` just added a `resource`:

```wit
package local:cowsay;
world cowsay {
  export test: func() -> string;
  export ts-version: func() -> string;
  resource compiler-options {
    constructor();
  }
}
```

The error looks to be raised from this `todo`:
https://github.com/bytecodealliance/jco/blob/f0eabadcb90c51a9c532ade5dd8d19cb669f559e/crates/js-component-bindgen/src/ts_bindgen.rs#L159-L160 

```
$ bunx jco transpile cowsay.wasm -o dist
thread '<unnamed>' panicked at crates/js-component-bindgen/src/ts_bindgen.rs:159:50:
not yet implemented
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
(jco transpile) RuntimeError: unreachable
    at js_component_bindgen_component.wasm.__rust_start_panic (wasm://wasm/js_component_bindgen_component.wasm-19499d82:wasm-function[43527]:0x10844a8)
    at js_component_bindgen_component.wasm.rust_panic (wasm://wasm/js_component_bindgen_component.wasm-19499d82:wasm-function[43520]:0x108403a)
    at js_component_bindgen_component.wasm._ZN3std9panicking20rust_panic_with_hook17h002a9d757cec9266E (wasm://wasm/js_component_bindgen_component.wasm-19499d82:wasm-function[43518]:0x1083ea0)
    at js_component_bindgen_component.wasm._ZN3std9panicking19begin_panic_handler28_$u7b$$u7b$closure$u7d$$u7d$17h5e11e16f0cbe0dd6E (wasm://wasm/js_component_bindgen_component.wasm-19499d82:wasm-function[43503]:0x1083101)
    at js_component_bindgen_component.wasm._ZN3std3sys9backtrace26__rust_end_short_backtrace17h34d7c06dcd4fe40dE (wasm://wasm/js_component_bindgen_component.wasm-19499d82:wasm-function[43502]:0x1083040)
    at js_component_bindgen_component.wasm.rust_begin_unwind (wasm://wasm/js_component_bindgen_component.wasm-19499d82:wasm-function[43510]:0x10837f0)
[...]

PR #525, opened 2024-11-20
feat: support top-level resources

guybedford:
Adds support for running components with top-level resources.

In the process fixes up our test infrastructure to ensure we transpile all wit-bindgen test cases again, which would have caught this one sooner.


guybedford:
Resolves https://github.com/bytecodealliance/jco/issues/508.

PR #528, opened 2024-11-26
feat: add module declaration support

### LINKED ISSUE #439: Question: `jco types` generating namespaces instead of modules?
I am using `jco types` to create the typings to build an SDK based on the generated code, the generated types contain namespaces instead of modules which means IDE does not provide intellisense and also `tsc` fails to compile without `//@ts-ignore` the wit  imports .

I have the following `wit`
```
package local:hello;

interface logger {
  log: func(val: string);
}

world hello {
    import logger; 
    export say-hello: func(name: string);
}
```

Using the following command to generate the types `jco types -o types <path to wit>` provides the following directory structure
```
types
├── hello.d.ts
└── interfaces
    └── local-hello-logger.d.ts
 ```
 And the contents of the files are 
 ```js
// hello.d.ts
import { LocalHelloLogger } from './interfaces/local-hello-logger.js';
export function sayHello(name: string): void;

// local-hello-logger.d.ts
export namespace LocalHelloLogger {
  export function log(val: string): void;
} 
```

And the following is the guest code
```js
// hello.js
import * as logger from 'local:hello/logger';

export function sayHello(name) {
  logger.log(`Hello ${name}`);
}
```

I could get around this by manually editing the generated types to be the following 
```ts
declare module "local:hello/logger" {
  export function log(val: string): void;
}
```

Please let me know if I am missing something here. Thanks!

lachieh:
Adds a `--declare-imports` option to the `jco types` command.

Closes #439

### DIFF
--- a/crates/js-component-bindgen/src/ts_bindgen.rs
+++ b/crates/js-component-bindgen/src/ts_bindgen.rs
@@ -29,6 +29,9 @@ struct TsBindgen {
     import_object: Source,
     /// TypeScript definitions which will become the export object
     export_object: Source,
+
+    /// Whether or not the types should be generated for a guest module
+    guest: bool,
 }
 
 /// Used to generate a `*.d.ts` file for each imported and exported interface for
@@ -59,6 +62,7 @@ pub fn ts_bindgen(
         local_names: LocalNames::default(),
         import_object: Source::default(),
         export_object: Source::default(),
+        guest: opts.guest,
     };
 
     let world = &resolve.worlds[id];
@@ -520,9 +524,15 @@ impl TsBindgen {
             return local_name;
         }
 
+        let module_or_namespace = if self.guest {
+            format!("declare module '{id_name}' {{")
+        } else {
+            format!("export namespace {camel} {{")
+        };
+
         let mut gen = self.ts_interface(resolve, false);
 
-        uwriteln!(gen.src, "export namespace {camel} {{");
+        uwriteln!(gen.src, "{module_or_namespace}");
         for (_, func) in resolve.interfaces[id].functions.iter() {
             // Ensure that the function  the world item for stability guarantees and exclude if they do not match
             if !feature_gate_allowed(resolve, package, &func.stability, &func.name)
--- a/src/jco.js
+++ b/src/jco.js
@@ -1,7 +1,7 @@
 #!/usr/bin/env node
 import { program, Option } from 'commander';
 import { opt } from './cmd/opt.js';
-import { transpile, types } from './cmd/transpile.js';
+import { transpile, types, guestTypes } from './cmd/transpile.js';
 import { run as runCmd, serve as serveCmd } from './cmd/run.js';
 import { parse, print, componentNew, componentEmbed, metadataAdd, metadataShow, componentWit } from './cmd/wasm-tools.js';
 import { componentize } from './cmd/componentize.js';
@@ -80,6 +80,18 @@ program.command('types')
   .option('--feature <feature>', 'enable one specific WIT feature (repeatable)', collectOptions, [])
   .option('--all-features', 'enable all features')
   .action(asyncAction(types));
+
+program.command('guest-types')
+  .description('(experimental) Generate guest types for the given WIT')
+  .usage('<wit-path> -o <out-dir>')
+  .argument('<wit-path>', 'path to a WIT file or directory')
+  .option('--name <name>', 'custom output name')
+  .option('-n, --world-name <world>', 'WIT world to generate types for')
+  .requiredOption('-o, --out-dir <out-dir>', 'output directory')
+  .option('-q, --quiet', 'disable output summary')
+  .option('--feature <feature>', 'enable one specific WIT feature (repeatable)', collectOptions, [])
+  .option('--all-features', 'enable all features')
+  .action(asyncAction(guestTypes));
 
 program.command('run')
   .description('Run a WASI Command component')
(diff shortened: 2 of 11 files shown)


lachieh:
@guybedford updated to `--guest`. The windows tests seems to be failing for another reason as far as I can tell.

crates/js-component-bindgen/src/ts_bindgen.rs
```diff
@@ -472,6 +476,7 @@ impl TsBindgen {
         resolve: &Resolve,
         id: InterfaceId,
         files: &mut Files,
+        guest: bool,
```
guybedford, line 479: Can `guest` be a property of the bindgen struct itself?

src/jco.js
```diff
@@ -79,6 +79,7 @@ program.command('types')
   .option('-q, --quiet', 'disable output summary')
   .option('--feature <feature>', 'enable one specific WIT feature (repeatable)', collectOptions, [])
   .option('--all-features', 'enable all features')
+  .option('--guest', 'generate types to be consumed by a guest component (module declarations)')
```
guybedford, line 82:
```suggestion
  .option('--guest', 'generate types to be consumed by a componentize guest component (using module declarations)')
```

guybedford (commented):
This looks like a great start, thanks.

I'm okay for us to land this as experimental to start even if it's not completely comprehensive. But I'm still wondering if there are other cases we should test though, perhaps some multi-import / multi-interface / resource generation or WASI worlds or otherwise?

crates/js-component-bindgen/src/ts_bindgen.rs
lachieh: Done. I initially went to do this but decided against it because of imposter syndrome 🥲. Glad to see I wasn't too far off base!

vados-cosmonic:
Hey @lachieh thanks for contributing this -- I guess this is for both you and @guybedford at the same time but what do y'all  think about having a more explicit name for the feature? 

I can think of at least two ways this would work -- removing the namespace and exporting a module "instead" of namespaces. So concretely something like `--disable-namespace` or `--no-namespace-wrapper` or `--use-module-wrapper`. Then if we have some sort of "suite" of guest-related options we want to set we can re-introduce `--preset-guest` that kind of wraps a bunch of them.

Also, I was really hoping that this would turn into the default over time as we know it's not likely to break people/works with most things. The ease of use is dramatically better "out of the box" when exporting the module instead of the namespace for guests.

I'm happy to help add tests in a follow up (along with some more examples that need to be added to `jco` which I'm working on)

lachieh:
This aligns with the experience I [detailed in the issue thread](https://github.com/bytecodealliance/jco/issues/439#issuecomment-2504023219). 

As far as I understand it, the types now are used mostly used for transpiling. I haven't read that code yet so forgive if this is not the right question; would switching to namespaces affect how transpiling works? Should module declarations for guest components be a net-new feature?

guybedford:
We discussed this in the meeting today - and if there may be potential for wanting to configure whether to use module declarations or namespaces within guest generation then I think this should be guest-specific ie `jco types --guest --use-namespace`. Conversely if we did want to configure for host type generation to use module declarations for some reason then we should have a flag for that on the `jco types --module-declarations` or otherwise.

This does have me thinking though - since the flag options may end up being specific to the host or guest path, it might even be worth considering treating this as a new top-level command - `jco guest-types` or `jco types-componentize` or otherwise?

t3hmrman:
> This does have me thinking though - since the flag options may end up being specific to the host or guest path, it might even be worth considering treating this as a new top-level command - jco guest-types or jco types-componentize or otherwise?

A bit late responding here but yeah this actually does seem like a good idea... *BUT* with the documentation burden it would cause... maybe we hold of on that for now :)

lachieh:
I'm in agreement there. Since the behavior of `jco types --guest --use-namespace` would be the same as `jco types` can we add that if it is requested?

Spoke with @vados-cosmonic today and found a couple of reasons we may indeed want to split the command into a new one, though it will require a bit more discussion—mostly around what you envision the JS/TS experience look like long term. I'm synthesizing some notes on that and will likely post a couple of issues.

Are we able to get this one in as is to get some feedback and then continue the discussion on the next phase?

src/jco.js
```diff
@@ -79,6 +79,7 @@ program.command('types')
   .option('-q, --quiet', 'disable output summary')
   .option('--feature <feature>', 'enable one specific WIT feature (repeatable)', collectOptions, [])
   .option('--all-features', 'enable all features')
+  .option('--guest', 'generate types to be consumed by a componentize guest component (using module declarations)')
```
guybedford, line 82: Per discussion, should we rename this to `jco guest-types` or similar?

guybedford (commented):
Thanks for your patience, will aim to land this this week - just wondering if we should switch over to a new command here or in a follow up?

guybedford (changes requested):
I realise we skipped over the discussion on this one this morning - please do let me know if there is more to discuss here.

Marking as request changes for now based on the agreement on using the `jco guest-types` command name, and not to miss that for review further. Happy to discuss alternatives as well.

lachieh:
Thanks, @guybedford! Not a worry—not much to discuss. I'll get this updated to the new command tomorrow, most likely.

src/jco.js
```diff
@@ -81,6 +81,18 @@ program.command('types')
   .option('--all-features', 'enable all features')
   .action(asyncAction(types));
 
+program.command('guest-types')
+  .description('Generate guest types for the given WIT')
```
guybedford, line 85: Perhaps let's note that this is experimental for now?

```suggestion
  .description('Generate experimental guest types for the given WIT')
```

guybedford (approved):
This is a great start, thank you.

In the main README we usually keep the CLI information up to date in https://github.com/bytecodealliance/jco?tab=readme-ov-file#cli (pending better docs here).

Just would be good to have an experimental note while this feature is not yet stable.

PR #533, opened 2024-12-02
feat(examples): add HTTP hello world

vados-cosmonic:
This adds an example that builds a HTTP hello world example


examples/components/http-hello-world/README.md
```diff
@@ -0,0 +1,184 @@
...
+A WebAssembly binary will be written to `string-reverse.wasm`.
+
+## Serving web requests with the WebAssembly component
+
+To run the component and serve requests we can either use `jco` or `wasmtime`:
+
+```console
+$ jco serve http-hello-world.wasm
+Server listening on 8000...
+```
```
guybedford, line 49: I think we should rather not recommend `jco serve` or alternatively be very clear that this is for development only.

examples/components/http-hello-world/README.md
```diff
@@ -0,0 +1,184 @@
...
+As WebAssembly components are powered by a [WebAssembly Interface Types ("WIT")][wit]-first workflow, making
+a HTTP handler component in WebAssembly requires creating a WIT contract to represent that component.
+
+This folder contains a `wit` directory (by convention) with the following content in `component.wit`:
+
+```wit
+package example:http-hello-world;
+
+world component {
+    export wasi:http/incoming-handler@0.2.0;
```
guybedford, line 78: Should we be using WASI@0.2.1? Or 0.2.2?

Thinking about this - we likely will need a process to update the examples to newer WASI versions.

examples/components/http-hello-world/README.md
```diff
@@ -0,0 +1,184 @@
...
+
+This will add many WIT files to your local repository, but you can move/rename all the downloaded `*.wit` files
+by making a folder named `deps` under `wit` and dropping them there.
+
+```console
+mkdir wit/deps
+mv *.wit wit/deps
+```
+
+### Typescript support
```
guybedford, line 123: I don't think we should be documenting `jco types` for guest component construction since it is not guest component typing compatible, at least until we have an implementation of `jco guest-types` or similar.

Alternatively we could carefully nuance that the types output by `jco types` are the types used when consuming the component and not when authoring it.

guybedford (commented):
Thanks for your patience here. Just left a few comments.

examples/components/http-hello-world/README.md
vados-cosmonic: Ah yeah -- I'll add a note that this is for development only!

examples/components/http-hello-world/README.md
vados-cosmonic: Ah thanks! will update this. In general we should try to use the latest, but also I don't expect things to just break if we're one or two minor versions behind.

Will update to 0.2.2 here!

examples/components/http-hello-world/README.md
vados-cosmonic: Yeah let's remove it for now, and when the other PR lands that adds `jco guest-types` we'll use that!

vados-cosmonic:
Thanks for the additional review @guybedford ! Addressed the outstanding points

PR #540, opened 2024-12-25
feat(examples): add browser fetch() example

vados-cosmonic:
This example adds a component that uses browser-provided `fetch()` (and related shims) which returns the response for a GET request to a given URL.

This example should showcase how to use a transpiled component with fetch from the browser context.

Currently, the example builds and tranpsiles the component but *does not* test it, as we need code in the async support PR to land first.

Thanks @cytommi for finding, reporting and diagnosing this bug!

### :warning: This PR is blocked on code in the [Async support PR](https://github.com/bytecodealliance/jco/pull/518)

While fetch is being translated at the engine level, the lacking implementation of `wasi:http/types` is causing `Request.send` to not actually get called properly.

### DIFF
--- a/examples/components/browser-fetch/.gitignore
+++ b/examples/components/browser-fetch/.gitignore
@@ -1,0 +1,4 @@
+node_modules
+dist
+*.wasm
+pnpm-lock.yaml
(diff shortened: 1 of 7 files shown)


guybedford:
Great to see, although yes we shouldn't include examples that depend on unreleased features I think unfortunately.

For browser examples, we are likely stuck with sync WebIDL examples mostly for now, but would still be great to see something here further.

vados-cosmonic:
Since this isn't quite ready, as browser bindings are not close merged yet, I'm going to close this for now to avoid clogging up the PR list! Will hold on to the branch

PR #544, opened 2025-01-17
chore: add webidl example

vados-cosmonic:


guybedford (commented):
This is a great start! Will be so useful to have this.

I think `npm install` is not enough here due to the Rust step in getting `webidl2wit` running. Would be great to see the steps fleshed out on that.

Then to have a test JS file that can show how the imports get picked up and describing the two Jco conventions used for the WIT conversion per https://bytecodealliance.github.io/jco/transpiling.html#experimental-webidl-imports.

vados-cosmonic:
> I think npm install is not enough here due to the Rust step in getting webidl2wit running. Would be great to see the steps fleshed out on that.

Ah yes -- so there actually *wasn't* a `webidl2wit` that was installable :) Been spending some time in https://github.com/wasi-gfx/webidl2wit getting a usable CLI put together.

> Then to have a test JS file that can show how the imports get picked up and describing the two Jco conventions used for the WIT conversion per https://bytecodealliance.github.io/jco/transpiling.html#experimental-webidl-imports.

Yup -- will do! `book-library.js` will turn into this, thanks for pointing out the docs to link to

guybedford:
This would be great to land - can we get the test running on CI?

vados-cosmonic:
Hey @guybedford so I got this working -- it requires a few things though:

- `webidl2wit` needs to [merge a PR upstream](https://github.com/wasi-gfx/webidl2wit/pull/42) (I can merge it there and release but would love for Mendy to get a chance to take a look first) 
- `jco` [needed a fix](https://github.com/bytecodealliance/jco/pull/544/commits/2f086bfad20c9c63d3d6de0cb7eea4a145316a9d) -- Classes didn't seem to work properly when passed through the `globalThis` and proxied

I'm going to make a separate PR for the jco fix so it can be reviewed by itself, but wanted to note it here (it all works in this PR)

vados-cosmonic:
Hey @guybedford this PR is now *only* blocked on merge + release of https://github.com/bytecodealliance/jco/pull/584 (webidl2wit fixes were doable on this side!), since examples use the published `jco` toolchain

crates/js-component-bindgen/src/intrinsics.rs
```diff
@@ -255,11 +255,13 @@ pub fn render_intrinsics(
...
-                            return proxy(res, () => {});
-                        if (typeof res === 'object' && res !== null)
-                            return () => proxy(res);
+                    function maybeProxy(res, prop) {
+                        // Catch Class lookups
+                        if (typeof res === "function" && res.prototype?.constructor === res) return res;
+                        // Catch "regular" function calls
+                        if (typeof res === 'function') return proxy(res, () => {});
+                        // Catch all other objects
+                        if (typeof res === 'object' && res !== null) return () => proxy(res);
```
guybedford, line 264: This seems like it still needs a rebase?

vados-cosmonic: Ah thank you, yes, this needed a rebase!

guybedford:
With this working, were we going to add it to CI like the others?

t3hmrman:
Yeah so we can't actually test `demo.js` until a new `jco` version is released with https://github.com/bytecodealliance/jco/pull/584 -- I was going to do that as a follow up post-release. 

What I can do though is make sure the compilation works, so I've added that.

vados-cosmonic:
Yeah so we can't actually test demo.js until a new jco version is released with https://github.com/bytecodealliance/jco/pull/584 -- I was going to do that as a follow up post-release.

What I can do though is make sure the compilation works, so I've added that.

guybedford:
@vados-cosmonic we should be able to run all the examples against the latest Jco in the main workspace itself which should then link in the local preview2-shims from the workspace as well.

I think all you need to do is set the `@bytecodealliance/jco` package dependency to `file:../../` in the examples packages.

vados-cosmonic:
Ah good idea -- will use a relative dep in the WebIDL example!

guybedford:
Alternatively try adding the examples folders to the workspace itself.

guybedford:
Perhaps something like:

```
  "workspaces": [
    ".",
    "packages/preview2-shim",
    "examples/components/webidl-book-library"
  ]
```

in https://github.com/bytecodealliance/jco/blob/main/package.json#L77, then keeping the normal Jco dependency definition.

vados-cosmonic:
Yeah that's even better since we're trying to move to workspaces anyway -- thanks for the suggestion

vados-cosmonic:
Hey @guybedford actually I don't think this will work -- we don't currently build the local jco to run the examples (we run pretty quickly with the published version) -- `cargo` and other things for building the local `jco` package aren't actually present.

What I'm doing is refactoring a bit so we do one `build` and then re-use it w/ examples. Unfortunately this means `main.yml` gets longer but it's not outrageous yet.

guybedford:
I've posted https://github.com/bytecodealliance/jco/pull/592 demonstrating just the workspaces part here, rebasing to that might help.

vados-cosmonic:
Yeah, so I tried a bunch of stuff and was able to get it working -- the problems that are left *should* be trivially fixable, but aren't:

- Using relative path jco means doing the cargo build in every test area (this lead me to refactor the CI to pass the built package along)
- Tests (including WebIDL) itself doesn't pickup that version of jco even when I install the local package, and installing it globally (and removing the local devDep) isn't a good solution.
- Separately, when tests run and pass w/ workspaces, they seem to return a nonzero code (I got this working with the WebIDL example).

I just don't think npm workspaces work very well with how the repo is set up yet -- it's incredibly unintuitive/obvious things don't work on the first try. I'm not sure the juice is worth the squeeze in this particular PR.

Maybe the workspace integration should be handled separately from this ticket -- making it work is introducing a ton of noise, and if someone pulls the repo to run an example I don't think they'd *want* to do an `npm run build` + `npm install` just to run the example in the first place. I intended those examples to be run against the published version so someone could pull and get started easily (`jco` doesn't have a project-starter templating mechanism yet, but maybe someday)

guybedford:
I landed the layering here for workspaces in https://github.com/bytecodealliance/jco/pull/592 which should avoid the concerns mentioned.

vados-cosmonic:
Hey @guybedford this should be ready for review now, I've rebased to the workspace integration (thanks!) and tests are :green_circle:

PR #551, opened 2025-01-28
JSPI and Asyncify

calvinrp:
Aiming for a relatively minimal diff by separating the WASI P2 shims into a separate PR. More PRs to follow with additional fixes and tests.


crates/js-component-bindgen/src/intrinsics.rs
```diff
@@ -114,6 +119,117 @@ pub fn render_intrinsics(
 
     for i in intrinsics.iter() {
         match i {
+            Intrinsic::AsyncifyAsyncInstantiate => output.push_str("
+                const asyncifyModules = [];
+                let asyncifyPromise;
+                let asyncifyResolved;
+                async function asyncifyInstantiate(module, imports) {
+                  const instance = await instantiateCore(module, imports);
+                  const memory = instance.exports.memory || (imports && imports.env && imports.env.memory);
```
guybedford, line 128: I think we should be a bit smarter about this memory detection, and instead make the `memoryExportName` an argument into the `asyncifyInstantiate` function - we should be able to have the context at the Rust side to know this information?

Also flipping between the imports or the exports for the memory seems a bit too loose - can you explain the cases further? Could we make it a boolean argument from the Rust side whether the import or export memory should be used?

crates/js-component-bindgen/src/intrinsics.rs
```diff
@@ -114,6 +119,117 @@ pub fn render_intrinsics(
...
+                  const realloc = instance.exports.cabi_realloc || instance.exports.cabi_export_realloc;
```
guybedford, line 129: Same as for the memory, this should be an argument.

crates/js-component-bindgen/src/intrinsics.rs
```diff
@@ -114,6 +119,117 @@ pub fn render_intrinsics(
...
+                  if (instance.exports.asyncify_get_state && memory) {
```
guybedford, line 130: What if we don't have a memory here or asyncify_get_state? Does the fallthrough work, or should we throw an error?

crates/js-component-bindgen/src/intrinsics.rs
```diff
@@ -114,6 +119,117 @@ pub fn render_intrinsics(
...
+                    } else {
+                        address = 16;
+                        new Int32Array(memory.buffer, address).set([address + 8, address + 1024]);
+                    }
+                    asyncifyModules.push({ instance, memory, address });
+                  }
+                  return instance;
+                }
+                function asyncifyState() {
+                  return asyncifyModules[0]?.instance.exports.asyncify_get_state();
```
guybedford, line 144: Should this function take an asyncify module index?

crates/js-component-bindgen/src/intrinsics.rs
```diff
@@ -114,6 +119,117 @@ pub fn render_intrinsics(
...
+                }
+                function asyncifyAssertNoneState() {
+                  let state = asyncifyState();
```
guybedford, line 147: Should this assert across all asyncify module indices?

crates/js-component-bindgen/src/intrinsics.rs
```diff
@@ -114,6 +119,117 @@ pub fn render_intrinsics(
...
+                    throw new Error(`reentrancy not supported, expected asyncify state '0' but found '${state}'`);
+                  }
+                }
+            "),
+
+            Intrinsic::AsyncifyWrapExport => output.push_str("
+                function asyncifyWrapExport(fn) {
+                  return async (...args) => {
+                    if (asyncifyModules.length === 0) {
+                        throw new Error(`none of the Wasm modules were processed with wasm-opt asyncify`);
```
guybedford, line 190: If we didn't have the fall-through path on the asyncify instantiate, would we be able to turn this into a static instead of a runtime error?

crates/js-component-bindgen/src/intrinsics.rs
```diff
@@ -114,6 +119,117 @@ pub fn render_intrinsics(
...
+                    }
+                    asyncifyAssertNoneState();
+                    let result = fn(...args);
+                    while (asyncifyState() === 1) {
```
guybedford, line 194: It sounds like the logic here is that the first module to be instantiated is the top of the call graph.

What happens if the first module we instantiate is not the top of the asyncify call graph?

crates/js-component-bindgen/src/intrinsics.rs
```diff
@@ -114,6 +119,117 @@ pub fn render_intrinsics(
...
+                    asyncifyAssertNoneState();
+                    return result;
+                  };
+                }
+            "),
+
+            Intrinsic::AsyncifyWrapImport => output.push_str("
+                function asyncifyWrapImport(fn) {
+                  return (...args) => {
+                    if (asyncifyState() === 2) {
```
guybedford, line 215: Could this be something like `asyncifyState(instantiateIdx)` where `instantiateIdx` is the index of the imported function being executed when sorted by the instantiation order?

crates/js-component-bindgen/src/transpile_bindgen.rs
```diff
@@ -1930,14 +2051,50 @@ impl<'a> Instantiator<'a, '_> {
         export_name: &String,
         resource_map: &ResourceMap,
     ) {
+        let is_async = self.async_exports.contains(&func.name)
```
guybedford, line 2054: Because function names can clash between components, we should figure out a way of defining `async_exports` and `async_imports` so that they can be uniquely addressing what they refer to.

Options:
1. Only support the shorthand forms for the outer component export and import interfaces, ie make sure we don't match any internal components, only publicly exposed import and export functions
2. Update the import and export functions to include the iterface name / resource name / etc
3. Go all in on supporting the world name and version in the async_exports and async_imports definitions

They're effectively just a world ID like any other function address, would be good to make sure we don't write ourselves into a corner where we need to make a breaking change in future.

I think the simple function names can work though, but with the restrictions of (1).

guybedford (commented):
Let some initial comments. I understand the transitive workflows aren't fully worked out yet, but we can at least think about the edge cases now to avoid writing ourselves into a corner.

calvinrp:
Closing this PR. Will separate JSPI and Asyncify into their own PRs.

guybedford:
Reopening as stalled but informative for anyone else wanting to take a stab.

PR #552, opened 2025-01-29
chore(deps): use published version of webidl2wit

vados-cosmonic:
Now that [`webidl2wit`](https://crates.io/crates/webidl2wit) is a published crate, we can use the released version (and later use [`webidl2wit-cli`](https://crates.io/crates/webidl2wit-cli) in docs and tests!)


guybedford:
Great!

PR #561, opened 2025-01-30
Experimental JSPI async option for `jco transpile`

calvinrp:
Adds the following new `jco transpile` options, which are labeled "EXPERIMENTAL":
- `--async-mode [mode]`: EXPERIMENTAL: For the component imports and exports, functions and methods on resources can be specified as `async`. The only option is `jspi` (JavaScript Promise Integration).
- `--async-imports <imports...>`: EXPERIMENTAL: Specify the component imports as `async`. Used with `--async-mode`.
- `--async-exports <exports...>`: EXPERIMENTAL: Specify the component exports as `async`. Used with `--async-mode`.

Expected follow on PRs:
- WASI P2 shims that use `async` functions;
- Asyncify `wasm-opt` pass integration as an alternative to `jspi` option;


src/jco.js
```diff
@@ -52,6 +52,11 @@ program.command('transpile')
   .option('--no-typescript', 'do not output TypeScript .d.ts types')
   .option('--valid-lifting-optimization', 'optimize component binary validations assuming all lifted values are valid')
   .addOption(new Option('--import-bindings [mode]', 'bindings mode for imports').choices(['js', 'optimized', 'hybrid', 'direct-optimized']).preset('js'))
+  .addOption(new Option('--async-mode [mode]', 'EXPERIMENTAL: use async imports and exports').choices(['sync', 'jspi']).preset('sync'))
+  .option('--default-async-imports', 'EXPERIMENTAL: default async component imports from WASI interfaces')
+  .option('--default-async-exports', 'EXPERIMENTAL: default async component exports from WASI interfaces')
```
guybedford, line 57: Should we call these `--async-wasi-imports` and `--async-wasi-exports` perhaps?

crates/js-component-bindgen/src/intrinsics.rs
```diff
@@ -285,6 +286,8 @@ pub fn render_intrinsics(
                 const i64ToF64 = i => (i64ToF64I[0] = i, i64ToF64F[0]);
             "),
 
+            Intrinsic::Imports => {},
```
guybedford, line 289: This new intrinsic doesn't seem to be used anywhere?

crates/js-component-bindgen/src/transpile_bindgen.rs
```diff
@@ -226,6 +253,28 @@ impl<'a> JsBindgen<'a> {
...
+        }
+
+        // adds a default implementation of `getCoreModule`
+        if matches!(self.opts.instantiation, Some(InstantiationMode::Async)) {
+            uwriteln!(
+                compilation_promises,
+                "if (!getCoreModule) getCoreModule = (name) => {}(new URL(`./${{name}}`, import.meta.url));",
+                self.intrinsic(Intrinsic::FetchCompile)
+            );
+        }
```
guybedford, line 277: Rather than doing this here and just for async, we could do this for all instantiations in https://github.com/bytecodealliance/jco/blob/main/crates/js-component-bindgen/src/transpile_bindgen.rs#L277.

James-Mart:
Is there any reason to prefer `asyncify` to `jspi`, or vice versa?

guybedford (commented):
Looks great. Does this support composed async calls where an async function exported calls a dependency component exported function in turn, calling an import? Do we have a test case for the transitive case like this?

src/jco.js
calvinrp: Yeah, that's good idea

crates/js-component-bindgen/src/intrinsics.rs
calvinrp: I debated removing this in this PR diff. Unrelated to this PR. I think the `imports` arg needs to be recorded for name conflicts. It is used in the instantiation function parameters. Is this the place to handle that?

crates/js-component-bindgen/src/transpile_bindgen.rs
calvinrp: Updated

src/jco.js
calvinrp: Updated

crates/js-component-bindgen/src/intrinsics.rs
guybedford: You only need to include it in the reserved strings list then.

calvinrp:
> Is there any reason to prefer `asyncify` to `jspi`, or vice versa?

JSPI is behind a feature flag in V8 (not available in other JS engines yet). Expected to land in stable in about a month. Asyncify is a `wasm-opt` pass that rewrites the Wasm binary and instruments it to be able to save the stack to linear memory and pause/replay. Introduces overhead but is better supported.

This PR just includes the JSPI implementation.

calvinrp:
> Looks great. Does this support composed async calls where an async function exported calls a dependency component exported function in turn, calling an import? Do we have a test case for the transitive case like this?

Added this test, which is a deeply nested component that has a `hello-world() -> string` export and import. This transpile test makes both `async` and exercises. https://github.com/bytecodealliance/jco/pull/561/files#diff-1160243babc0ac123bd874b8bd4a131f802dfb1c1ba3d7f77b02452b999011e1R108-R133

src/cmd/transpile.js
```diff
@@ -141,11 +185,13 @@ async function wasm2Js (source) {
  *   js?: bool,
  *   minify?: bool,
  *   optimize?: bool,
+ *   preoptimized?: bool,
  *   namespacedExports?: bool,
  *   outDir?: string,
  *   multiMemory?: bool,
  *   experimentalIdlImports?: bool,
  *   optArgs?: string[],
+ *   wasmOptBinPath?: string,
```
guybedford, line 194: Did you want to add this through into opt or was this kept in by mistake?

src/cmd/transpile.js
```diff
@@ -141,11 +185,13 @@ async function wasm2Js (source) {
  *   js?: bool,
  *   minify?: bool,
  *   optimize?: bool,
+ *   preoptimized?: bool,
```
guybedford, line 188: This also seems like it's not being used?

crates/js-component-bindgen/src/transpile_bindgen.rs
guybedford: It should be possible to move this into the intrinsics.rs file as the first line inside the `instantiate ()` function declaration to then apply generally.

guybedford (approved):
The tests are looking really good now thanks. Will be great to get this landed asap.

src/cmd/transpile.js
calvinrp: Removing, left from a previous iteration.

src/cmd/transpile.js
calvinrp: Removing, left over from some of the Asyncify tests

crates/js-component-bindgen/src/transpile_bindgen.rs
guybedford: Actually that makes sense this is InstantiationMode Async only, thanks for talking it through.

Issue #418, opened 2024-04-10
JS/WIT Reference


guybedford:
It would be beneficial to have a complete reference documentation for how WIT types are represented in JS, covering all WIT generations.

While this will still evolve a little over time, it forms the public API of Jco's version contract as well, so we should document it and ensure that any breaking changes are reflected as major versions.

PR #563, opened 2025-02-05
feat(docs): add docs for object conversions

vados-cosmonic:
Resolves https://github.com/bytecodealliance/jco/issues/418.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -9,7 +9,7 @@ jobs:
   deploy:
     permissions:
       contents: write
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-24.04
     concurrency:
       group: ${{ github.workflow }}-${{ github.ref }}
     steps:
@@ -18,8 +18,13 @@ jobs:
       - name: Setup mdBook
         uses: peaceiris/actions-mdbook@v1
         with:
-          mdbook-version: '0.4.30'
-          # mdbook-version: 'latest'
+          mdbook-version: '0.4.43'
+
+      - uses: taiki-e/install-action@v2
+        with:
+          tool: >-
+            mdbook-alerts@0.7.0,
+            mdbook-linkcheck@0.7.7
 
       - run: mdbook build docs
 
--- a/docs/src/wit-type-representations.md
+++ b/docs/src/wit-type-representations.md
@@ -1,0 +1,411 @@
+# WIT Type Representations
+
+Similar to any other guest langauge, there are multiple type systems in play when dealing with JS WebAssembly components.
+
+Types represented in [WebAssembly Interface Types ("WIT")][wit] must be converted down to types that are familiar for Javascript,
+and [Typescript][ts] (if dealing with `jco types` or `jco guest-types` subcommands).
+
+This document details the type representations and usage for types that are defined in WIT and built into components.
+
+[wit]: https://github.com/WebAssembly/component-model/blob/main/design/mvp/WIT.md
+[ts]: https://www.typescriptlang.org/
+
+## Basic types
+
+Here is a basic table of conversions between WIT types and JS types:
+
+More complicated types that are built into WIT but require more work to translate are explained below.
+
+| WIT type | JS Type                   |
+|----------|---------------------------|
+| `u8`     | `number`                  |
+| `u16`    | `number`                  |
+| `u32`    | `number`                  |
+| `u64`    | [`BigInt`][mdn-js-bigint] |
+| `s8`     | `number`                  |
+| `s16`    | `number`                  |
+| `s32`    | `number`                  |
+| `s64`    | [`BigInt`][mdn-js-bigint] |
+| `f32`    | `number`                  |
+| `f64`    | `number`                  |
+| `bool`   | `boolean`                 |
+| `char`   | `string`                  |
+| `string` | `string`                  |
+
+[mdn-js-bigint]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt
+
+## Variants (`variant`)
+
+> [!NOTE]
+> See the [Variant section of the WIT IDL](https://github.com/WebAssembly/component-model/blob/main/design/mvp/WIT.md#item-variant-one-of-a-set-of-types) for more information on Variants
+
+Variants are like basic enums in most languages with one exception; members of the variant can hold a single data type.
+Alternative variant members may hold *different* types to represent different cases. For example:
+
+```wit
+variant exit-code {
+  success,
+  failure-code(u32),
+  failure-msg(string),
+}
+```
+
+### WIT syntax
+
+```wit
+variant filter {
+    all,
+    none,
+    some(list<string>),
+}
+```
+
+### Jco Representation
+
+Jco represents variants as objects with a `tag` that represents the variant, and `data` that represent the content:
+
+For example, pseudo Typescript for the of the above `filter` variant would look like the following:
+
+```ts
+// Filter with all
+{
+  tag: 'all';
+}
+
+// Filter with None
+{
+  tag: 'none';
+}
+
+// Filter with some and a list of strings
+{
+  tag: 'some';
+  data: string[];
+}
+```
+
+> [!NOTE]
+> WIT `variant`'s options may only contain *one* piece of data.
+>
+> You can work around this limitation of variants by having the contained type be a *tuple*,
+> (e.g. `tuple<string, u32, string>`), or using a named record as the related data.
+
+## Records (`record`)
+
+### WIT Syntax
+
+```wit
+record person {
+    name: string,
+    age: u32,
+    favorite-color: option<string>,
+}
+```
+
+### Jco Representation
+
+Jco represents records as the [Javascript Object basic data type][mdn-js-obj]:
+
+Given the WIT record above, you can expect to deal with an object similar to the following Typescript:
+
+```ts
+interface Person {
+  person: string;
+  age: number;
+  favoriteColor?: number;
+}
+```
+
+> [!NOTE]
+> If using `jco guest-types` or `jco types`, you will be able to use Typescript types that
+> properly constrain the Typescript code you write.
+
+[mdn-js-obj]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object
+
+## Options (`option`)
+
+### WIT Syntax
+
+```wit
+option<u32, u32>
+option<string, u32>
+```
+
+### Jco Representation
+
+Jco represents options as an optional value or undefined, so some examples:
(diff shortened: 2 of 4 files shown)


docs/src/wit-type-representations.md
```diff
@@ -0,0 +1,250 @@
...
+Here is a basic table of conversions between WIT types and JS types:
+
+More complicated types that are built into WIT but require more work to translate are explained below.
+
+| WIT type | JS Type   |
+|----------|-----------|
+| `u8`     | `number`  |
+| `u16`    | `number`  |
+| `u32`    | `number`  |
+| `u64`    | `number`  |
```
guybedford, line 24: We use `BigInt` for u64.

docs/src/wit-type-representations.md
```diff
@@ -0,0 +1,250 @@
...
+### WIT Syntax
+
+```
+list<u8>
+list<string>
+```
+
+### Jco Representation
+
+Jco represents lists with native Javscript Arrays, with some exceptions:
```
guybedford, line 46:
```suggestion
Jco represents lists with native Javscript Arrays, with the exception of a `list<u8>`:
```

docs/src/wit-type-representations.md
```diff
@@ -0,0 +1,250 @@
...
+```wit
+record person {
+    name: string,
+    age: u32,
+}
+```
+
+### Jco Representation
+
+Jco represents records as the [Javascript Object basic data type][mdn-js-obj]:
```
guybedford, line 84: The extra information for object types, is that optional types on objects are represented as optional properties of the object itself.

docs/src/wit-type-representations.md
```diff
@@ -0,0 +1,250 @@
...
+}
+```
+
+> [!NOTE]
+> WIT `variant`'s options may only contain *one* piece of data.
+>
+> You can work around this limitation of variants by having the contained type be a *tuple*,
+> (e.g. `tuple<string, u32, string>`), or using a named record as the related data.
+
+## Result (`result`)
```
guybedford, line 157: We're missing `optional` here as well, and I think we should put both result and optional before lists, tuples and objects and right after variants, because they have interesting chaining behaviours with the other types.

docs/src/wit-type-representations.md
```diff
@@ -0,0 +1,250 @@
...
+
+> [!NOTE]
+> See the [Basic types section of the WIT IDL](https://github.com/WebAssembly/component-model/blob/main/design/mvp/WIT.md#types) for more information on Resources
+
+Result represent a result that *may or may not* be present, due to a failure. A result value either contains
+a value that represents a completed computation, or some "error" that indicates a failure. You can think
+of the type of a result as:
```
guybedford, line 164: This doesn't feel quite right to me.

Rather the rule is something like:

> A result has two representations depending on whether it is the return type of a function or a nested type. When a result is returned directly from a function, any thrown error of the function is treated as the result error type, while any direct return value is treated as the result success type. On the other hand, a result stored inside a type otherwise returned from a function will be treated as a result type much like a variant type of the form `{ tag: 'ok', val: SuccessType } | { tag: 'err', val: ErrorType }`.

guybedford (commented):
Would be great to continue iterating on this.

docs/src/wit-type-representations.md
vados-cosmonic: Ah yes -- this should have been `BigInt`, will fix

docs/src/wit-type-representations.md
vados-cosmonic: Ah thanks for the note, I'll include an `option<string>` in there as an example to illustrate that

docs/src/wit-type-representations.md
vados-cosmonic: Ah yes -- will add optional and do a bit of a restructuring in something closer to simple type order.

docs/src/wit-type-representations.md
vados-cosmonic: That paragraph is actually a general description on the `Result` type as a general concept -- I don't think it's something everyone will have necessarily seen, especially in the JS world (as opposed to just `value | null` or something).

I included the rule you added -- IIRC it's actually configurable but I won't mention that here!

vados-cosmonic:
Thanks for the feedback! Wasn't able to get back to this, but just addressed the review points

docs/src/wit-type-representations.md
vados-cosmonic: Ah *one* change I did here was to put objects *before* tuples.

What I'm going for (insipired by the issues what you pointed out) is to include the listing this way:
- basic types
- sum/product types
- containers (fixed)
- containers (unbounded)
- resources

docs/src/wit-type-representations.md
```diff
@@ -0,0 +1,345 @@
...
+
+### Jco Representation
+
+Jco represents options as as an optional (arrays), so some examples:
+
+| Type          | Representation (TS)  | Example                               |
+|---------------|----------------------|---------------------------------------|
+| `option<u32>` | `number | undefined` | `option<u32>` -> `number | undefined` |
+
+When used in the context of a `record` (which becomes a [JS Object][mdn-js-obj]), optional values are represented as optional properties (i.e in TS a `propName?: value`).
```
guybedford, line 142: Another point to include for option here is that functions taking options as arguments have those arguments treated as optional arguments

docs/src/wit-type-representations.md
```diff
@@ -0,0 +1,345 @@
...
+
+function f(input: Result<string, string>): string {
+  switch (input.tag) {
+    case 'ok': return `SUCCESS, returned: [${input.val}]";
+    case 'err': return `ERROR, returned: [${input.val}]";
+    // We we should never reach the case below
+    default: throw Error("something has gone seriously wrong");
+  }
+}
+```
```
guybedford, line 199: There's even more to mention on the Result type actually, in particular the behaviour of `getErrorPayload`:

* When running components in JS hosts, host functions may want to throw real JS errors per JS idioms for other callers instead of throwing the exact type expected by Jco. For example when an error code number is expected, throwing a number for JS consumers would be a JS anti-pattern.
* To support this, Error objects with a `payload` property will have that `payload` property extracted as the result error type.
* For example, a host function may be written `function importedJcoFunctionThatThrows () { throw Object.assign(new Error('Error for JS users'), { payload: ERR_CODE_JCO_ERROR_CODE_PER_RESULT_ERR_TYPE }) }`

guybedford (commented):
Thanks, looking good. Yeah I know it's better to have simpler types first, but variant types end up being more fundamental weirdly.

Added some more of the important nuances to cover. Thanks again for your work here.

docs/src/wit-type-representations.md
```diff
@@ -0,0 +1,345 @@
...
+option<string, u32>
+```
+
+### Jco Representation
+
+Jco represents options as as an optional (arrays), so some examples:
+
+| Type          | Representation (TS)  | Example                               |
+|---------------|----------------------|---------------------------------------|
+| `option<u32>` | `number | undefined` | `option<u32>` -> `number | undefined` |
```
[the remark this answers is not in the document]
vados-cosmonic: AH thanks for pointing this out -- I see the case in `as_nulalble()`... Will add this

docs/src/wit-type-representations.md
vados-cosmonic: Yup, will add to this -- along with the nesting note

docs/src/wit-type-representations.md
vados-cosmonic: Hmnn OK I think we might want to make these sections a bit more explicit so they're easy to scan, will do some refactoring and add these notes w/ examples here.

vados-cosmonic: Hey I addressed this -- please take a look https://github.com/bytecodealliance/jco/pull/563/commits/c148e759ed694e3b9fdb671f11e7c0c32459283d
(unfortunately there's a little reorg in there, but I think I added an easy-to-understand example of your notes above).

.github/workflows/docs.yml
```diff
@@ -18,8 +18,13 @@ jobs:
       - name: Setup mdBook
         uses: peaceiris/actions-mdbook@v1
         with:
-          mdbook-version: '0.4.30'
-          # mdbook-version: 'latest'
+          mdbook-version: '0.4.43'
+
+      - uses: taiki-e/install-action@510b3ecd7915856b6909305605afa7a8a57c1b04 # v2.48.1
```
guybedford, line 23: I still prefer to use semver major versions in these as I will never update one of these things.

docs/src/wit-type-representations.md
```diff
@@ -0,0 +1,410 @@
...
+### WIT Syntax
+
+```wit
+option<u32, u32>
+option<string, u32>
+```
+
+### Jco Representation
+
+Jco represents options as as an optional (arrays), so some examples:
```
guybedford, line 136:
```suggestion
Jco represents options as an optional value or undefined, so some examples:
```

docs/src/wit-type-representations.md
```diff
@@ -0,0 +1,410 @@
...
+interface optional {
+    f: func(n: option<u32>) -> string;
+}
+```
+
+An implementation of the function `optional.f` would look like the following Typescript:
+
+```ts
+function f(n?: number): string {
+  if (typeof n === 'undefined') { return "no n provided"; }
```
guybedford, line 169: Undefined checking in JS doesn't need a typeof, that's only for objects and other types.

```suggestion
  if (n === undefined) { return "no n provided"; }
```

docs/src/wit-type-representations.md
```diff
@@ -0,0 +1,410 @@
...
+> How JS treats large numbers is not in focus here, but it is worth noting that
+> `Number.MAX_VALUE + Number.MAX_VALUE === Infinity`.
+
+### Typescript Schema
+
+```
+interface Result {
+  ok: null
+  data: null
+}
```
guybedford, line 229: Not clear what this is describing?

guybedford (approved):
Looking great, thank you.

.github/workflows/docs.yml
vados-cosmonic: Ah right -- sorry force-of-habit here! Normally dependabot is quite happy to update these as necessary, forget we're not using it here!

docs/src/wit-type-representations.md
vados-cosmonic: Ah -- tahnks!

docs/src/wit-type-representations.md
vados-cosmonic: Ah this was a really old artifact, the `type Result<T,E>` is much easier to parse, updating to that!

PR #564, opened 2025-02-05
chore(examples): update to use latest jco version, guest-types

vados-cosmonic:


guybedford:
Thanks, is there a good suggestion for how to keep these examples up-to-date? Or should we just include updating the examples as part of the release process "handbook" effectively?

vados-cosmonic:
Yeah I want to keep the release process as light as possible, and automate as much as we can...

I think the little refactoring that we talked about during the last call would help here -- if we could turn our current repo into essentially two workspace-based repo (one for rust, one for JS) then we could get some more automation going. 

What I'm thinking of is:

- `crates` & `packages` folders
- `justfile` for top level scripting that either calls down into rust or JS depending)

If we have the code in workspaces, then we can manage the deps all in one spot, and at least make it easier to update all the deps at once.

guybedford:
That framing certainly makes sense to me.

PR #575, opened 2025-02-24
fix: stdout/stderr pollable hang

### LINKED ISSUE #550: flush cause pollable to hang
## Description
Flush, then poll on the stdout, will cause the program to hang.

## Expected behavior
The program finishes, as wasmtime's behavior.

## Steps to reproduce:

- run `cargo component new flush --lib && cd flush`
- update `deps.toml` and run `wit-deps update`:
  ```toml
  cli = "https://github.com/Webassembly/wasi-cli/archive/refs/tags/v0.2.0.tar.gz"
  ```
- update `world.wit` (and `deps.toml`) to:
  ```wit
  package component:flush;

  /// An example world for the component to target.
  world example {
      include wasi:cli/command@0.2.0;
  }
  ```
- append to `Cargo.toml`
  ```toml
  [package.metadata.component.target.dependencies]
  "wasi:random" = { path = "wit/deps/random" }
  "wasi:cli" = { path = "wit/deps/cli" }
  "wasi:io" = { path = "wit/deps/io" }
  "wasi:clocks" = { path = "wit/deps/clocks" }
  "wasi:filesystem" = { path = "wit/deps/filesystem" }
  "wasi:sockets" = { path = "wit/deps/sockets" }
  ```
- Modify `src/lib.rs`
  ```rust
  #[allow(warnings)]
  mod bindings;
  
  use bindings::wasi;
  
  use crate::bindings::exports::wasi::cli::run::Guest;
  
  struct Component;
  
  impl Guest for Component {
      fn run() -> Result<(), ()> {
          let stdout = wasi::cli::stdout::get_stdout();
          stdout.flush().unwrap();
          let poll = stdout.subscribe();
          poll.block();
          Ok(())
      }
  }
  
  bindings::export!(Component with_types_in bindings);
  ```
- run `cargo component build`
- run `wasmtime run target/wasm32-wasip1/debug/flush.wasm` and `jco run target/wasm32-wasip1/debug/flush.wasm`

vados-cosmonic:
Note that to reproduce the fix, you need to use a local override for `@bytecodealliance/preview2-shim` (that's done in this PR), but that will have to be removed before the next public release of `jco` (and after `preview2-shim` is released)

Resolves #550


packages/preview2-shim/lib/io/worker-thread.js
```diff
@@ -646,34 +646,87 @@ function handle(call, id, payload) {
...
+          // stdout/stderr in a port that is only flushed on exit of the worker.
+          //
+          // In this case, we cannot attempt to wait for the promise.
+          //
+          // This code may have to be reworked for browsers.
+          //
+          // see: https://github.com/nodejs/node/blob/v22.12.0/lib/internal/worker/io.js#L288
+          // see: https://github.com/nodejs/node/blob/v22.12.0/lib/internal/worker.js#L303
+          // see: https://github.com/nodejs/node/blob/v22.12.0/typings/internalBinding/messaging.d.ts#L27
+          stream.stream.write(new Uint8Array([]));
```
guybedford, line 689: I would have thought that this write would be unnecessary in this case, since Node.js IO is always flushing.

guybedford (commented):
Thanks so much for digging into this - we just need to change that the "blocking write and flush" path must still do the write.

packages/preview2-shim/lib/io/worker-thread.js
guybedford: That is we only do an empty write in the other cases, to be able to get the event for when the flush has completed.

packages/preview2-shim/lib/io/worker-thread.js
```diff
@@ -646,34 +646,87 @@ function handle(call, id, payload) {
       }
       stream.pollState.ready = false;
       return (stream.flushPromise = new Promise((resolve, reject) => {
-        stream.stream.write(payload, (err) => {
+        if (stream.stream === stdout || stream.stream === stderr) {
```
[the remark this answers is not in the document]
vados-cosmonic: Ah good catch -- missed copying the original write to this branch. Thanks

packages/preview2-shim/lib/io/worker-thread.js
vados-cosmonic: Yup, I removed this and the last case too -- this write can safely be skipped.

vados-cosmonic:
Reduced the number of commits -- once the tests are :green_circle: this should be ready to go, thanks for the review @guybedford

PR #576, opened 2025-02-26
fix: usage of private member from output stream

### LINKED ISSUE #577: Failure to run `wstd` `http-client` example wasm binary
https://github.com/yoshuawuyts/wstd/blob/main/examples/http_client.rs
cargo 1.87.0-nightly (1d1d646c0 2025-02-21), Rust edition 2021, `cargo component build --release --target wasm32-wasip2`

```js
import { run } from "./transpiled/wstd-example.js"

console.log(run.run())
```

When running `node index.js https://example.com`:

```
> GET / HTTP/1.1
< HTTP/1.1 200 OK
< content-type: text/html
< etag: "84238dfc8092e5d9c0dac8ef93371a07:1736799080.121134"
< last-modified: Mon, 13 Jan 2025 20:11:20 GMT
< cache-control: max-age=697
< date: Wed, 26 Feb 2025 04:00:26 GMT
< alt-svc: h3=":443"; ma=93600,h3-29=":443"; ma=93600,quic=":443"; ma=93600; v="43"
< content-length: 1256
< connection: keep-alive
.../node_modules/@bytecodealliance/preview2-shim/lib/io/worker-io.js:323
      { src: src.#id, len }
                 ^

TypeError: Cannot read private member #id from an object whose class did not declare it
    at OutputStream.splice (.../node_modules/@bytecodealliance/preview2-shim/lib/io/worker-io.js:323:18)      
    at trampoline56 (.../transpiled/wstd-example.js:2917:34)
    at wit-component:shim.indirect-wasi:io/streams@0.2.3-[method]output-stream.splice (wasm://wasm/wit-component:shim-b13bbaba:wasm-function[16]:0x283)
    at wasm://wasm/001e552a:wasm-function[184]:0x7504
    at wasm://wasm/001e552a:wasm-function[281]:0x16dc8
    at wasm://wasm/001e552a:wasm-function[279]:0x103e0
    at wasm://wasm/001e552a:wasm-function[96]:0x3286
    at wasm://wasm/001e552a:wasm-function[329]:0x24c61
    at wasm://wasm/001e552a:wasm-function[61]:0x148e
    at wit-component:adapter:wasi_snapshot_preview1.wasi:cli/run@0.2.3#run (wasm://wasm/wit-component:adapter:wasi_snapshot_preview1-2fd92fea:wasm-function[29]:0x701)

Node.js v22.5.1
```

vados-cosmonic:
Resolves #577

### DIFF
--- a/packages/preview2-shim/lib/io/worker-io.js
+++ b/packages/preview2-shim/lib/io/worker-io.js
@@ -200,8 +200,9 @@ class InputStream {
       this
     );
   }
-  static _id(stream) {
-    return stream.#id;
+
+  id() {
+    return this.#id;
   }
   /**
    * @param {FILE | SOCKET_TCP | STDIN | HTTP} streamType
@@ -260,9 +261,6 @@ function httpInputStreamDispose(id) {
 export const inputStreamCreate = InputStream._create;
 delete InputStream._create;
 
-export const inputStreamId = InputStream._id;
-delete InputStream._id;
-
 class OutputStream {
   #id;
   #streamType;
@@ -320,14 +318,14 @@ class OutputStream {
     return streamIoErrorCall(
       OUTPUT_STREAM_SPLICE | this.#streamType,
       this.#id,
-      { src: outputStreamId(src), len }
+      { src: src.id(), len }
     );
   }
   blockingSplice(src, len) {
     return streamIoErrorCall(
       OUTPUT_STREAM_BLOCKING_SPLICE | this.#streamType,
       this.#id,
-      { src: inputStreamId(src), len }
+      { src: src.id(), len }
     );
   }
   subscribe() {
@@ -335,10 +333,10 @@ class OutputStream {
       ioCall(OUTPUT_STREAM_SUBSCRIBE | this.#streamType, this.#id)
     );
   }
-
-  static _id(outputStream) {
-    return outputStream.#id;
+  id() {
+    return this.#id;
   }
+
   /**
    * @param {OutputStreamType} streamType
    * @param {any} createPayload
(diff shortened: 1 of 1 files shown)


guybedford:
Splice should always be called with a valid instance of OutputStream I think? So if that's failing that's probably already an invariant breakdown?

vados-cosmonic:
So the error that we were seeing wasn't that the OutputStream was invalid -- just that the member couldn't be accessed. Am I misunderstanding how you inferred that the outputstream stream was invalid?

It's not actually clear to me why we were using the static methods (I assumed to prevent external mutation) to access the id rather than a getter function that copies the value. 

That said, with the change we actually get past this issue and there *another* much deeper issue... 

`wstd` gets the *same* Pollable resource as a previously used (and previously dropped/destroyed one on the jco side) one when they call `trailers.subscribe()`:

https://github.com/yoshuawuyts/wstd/blob/main/src/http/body.rs#L171

The upstream impl of `wasi:http` finish actually seems to effectively reuse (from what I can tell) an identical table entry:

https://github.com/bytecodealliance/wasmtime/blob/1302a0a4eddf2ddadfa4421a8d62b2b47c2c8be1/crates/wasi-http/src/types_impl.rs#L716

Which explains how we get a Pollable with an already-used ID out of `trailers.subscribe()`.

What I'm seeing is that the resource gets dropped on the Jco side, and then when the component comes back asking about the *same* resource ID, we can't find it  (because it's been cleaned up/removed from our internal Map).

guybedford:
The zero poll is supposed to be an immediately resolving poll. Trailer support should throw an error or panic that notes the real cause.

vados-cosmonic:
Yeah -- immediately resolving polls aren't *yet* implemented (we actually try and look it up), but also implementing trailers is probably the only right :tm: way to solve this :sweat_smile: , so I'll try to find time to work on this after making some progress on async stuff.

vados-cosmonic:
This isn't solved yet, but since this issue is basically a missing trailers impl, going to leave this for when someone gets to adding trailers support

PR #584, opened 2025-03-04
fix: allow objects through globalThis intrinsic (WebIDL)

vados-cosmonic:
This commit enables objects to be set on the global context and used via the WebIDL integration in Jco. Previously, objects would be detected as functions and then proxied into another function, which prevented Object.create(O.prototype) calls from succeeding.

With this change, we do a little more work to check for check for dealing wiht a class.

### DIFF
--- a/crates/js-component-bindgen/src/intrinsics.rs
+++ b/crates/js-component-bindgen/src/intrinsics.rs
@@ -235,7 +235,7 @@ pub fn render_intrinsics(
                                 if (res === undefined && prop[0].toUpperCase() === prop[0]) {
                                     return Object.getPrototypeOf(globalThis[`${prop[0].toLowerCase()}${prop.slice(1)}`]).constructor;
                                 }
-                                return maybeProxy(res);
+                                return maybeProxy(res, prop);
                             },
                             apply: (_, thisArg, args) => {
                                 if (args.length === 1 && Array.isArray(args[0]) && origTarget.length === 0) args = args[0];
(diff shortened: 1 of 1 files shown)


crates/js-component-bindgen/src/intrinsics.rs
```diff
@@ -255,11 +255,15 @@ pub fn render_intrinsics(
...
                         });
                     }
-                    function maybeProxy(res) {
-                        if (typeof res === 'function')
-                            return proxy(res, () => {});
-                        if (typeof res === 'object' && res !== null)
-                            return () => proxy(res);
+                    function maybeProxy(res, prop) {
+                        // Catch prototype lookups
+                        if (typeof res === "object" && prop === "prototype") return res;
```
guybedford, line 260: Not proxying here means that we don't be able to lookup things "past" the prototype right? But that definitely isn't needed?

crates/js-component-bindgen/src/intrinsics.rs
```diff
@@ -255,11 +255,15 @@ pub fn render_intrinsics(
...
+                        // Catch Class lookups
+                        if (typeof res === "function" && prop[0].toUpperCase() === prop[0] && res.name === prop && !!res.constructor) return res;
```
guybedford, line 262: The uppercasing convention here feels very ad-hoc. Could we do something more like:

```
obj.prototype?.constructor === obj
```

instead?

And again like above the short-circuiting is desired?

guybedford (commented):
Thanks for diving into this code! So great to see. Is there an easy way to test this?

crates/js-component-bindgen/src/intrinsics.rs
vados-cosmonic: It is -- I followed the convention of the existing code (the upper case check, which wasn't immediately clear) -- I tried it and this also works -- will update!

vados-cosmonic: oh and as far as the short circuiting being desired, yes -- it's what was stopping classes provided from JS from being instantiatable/working properly. Classes provided to satisfy WebIDL bindings from JS don't work otherwise -- the `Object.create(A.prototype)` ends up with `A.prototype` being `() => proxy(A.protoype)`, essentially.

crates/js-component-bindgen/src/intrinsics.rs
vados-cosmonic: It wasn't needed in any of the existing tests but it's hard to say it could never be needed... 

I did some checking and what I found is that the class lookup below actually is enough to "fix" this -- checking for class was the important case so I'll just leave that one.

PR #585, opened 2025-03-05
feat: wasi p3 async support for jco

vados-cosmonic:
This PR is almost certain to be long lived, and contains the ongoing p3 async support for 

To understand this PR, see the following other resources:

- https://github.com/WebAssembly/component-model/blob/main/design/mvp/CanonicalABI.md
- https://github.com/WebAssembly/component-model/blob/main/design/mvp/Async.md
- https://github.com/WebAssembly/component-model/blob/main/design/mvp/WIT.md

Current ongoing implementation in `wasmtime` is over here:

- https://github.com/bytecodealliance/wasip3-prototyping (temporary fork of `wasmtime`)
- https://github.com/bytecodealliance/wit-bindgen (async-ready changes are freely upstreamed)

Particularly useful from the upstream repos are the test programs that show usage of the new async APIs from the only language with support at present (Rust):

- https://github.com/bytecodealliance/wasip3-prototyping/tree/main/crates/test-programs/src/bin
  - Ex. [`async_error_context.rs`](https://github.com/bytecodealliance/wasip3-prototyping/blob/main/crates/test-programs/src/bin/async_error_context.rs) which is one of the fixtures we add and use.


vados-cosmonic:
Just a note -- I'm rolling the upgrade to tooling version 227.1 here -- the upgrade seems to have broken some things on the jco side, right now the remaining errors are:

- **unreachable resource trampolines**: It turns out that some trampolines that are *supposed* to be unreachable are actually needed, so I think upstream might have made our detection logic not work.
- **Deduping** The deduping logic seems to be broken with upstream changes -- in `PString$1` isn't being generated anymore.

guybedford:
@vados-cosmonic please create a separate PR for the 227.1 upgrade path for direct review and testing, that way I can also more easily assist.

vados-cosmonic:
Ah apologies for the noise, will pull out that update into a new PR, thanks for the help @guybedford :bow:

vados-cosmonic:
OK, so I had to make a couple concessions here:

- updating upstream libs broke the WebIDL testing (this seems to be purely a problem with resources that contain methods with the same exact name -- at the wasm-tools component new level)
- Actually running the updated async error-context component requires many more intrinsics than it did when this PR was first open, and to stop this PR from growing even more, the tests checks generation but stops JUST short of actually running the function.

PR #592, opened 2025-03-10
support workspaces for examples

guybedford:
This adds support for treating the examples as workspaces so they build against the local preview2-shims and Jco versions.


guybedford:
Due to an optional dependencies conflict between CI and local I had to disable the lockfile.

Not having a lockfile for a library is perfectly fine though since testing should always be testing against the latest versions of deps, unless there are dev dependencies which need to be locked of course.

Alternatively we can solve the issue where napi-rs was not upgrading its optional dependencies properly.

PR #594, opened 2025-03-12
feat: add utf8-utf16 transcoder

peter-jerry-ye:


vados-cosmonic:
Hey @peter-jerry-ye this looks great! Would you mind adding a test for this? I can help explaining the test infrastructure if necessary, but having some code that exercised this path would be excellent to make sure it works now and into the future.

peter-jerry-ye:
Hi @vados-cosmonic . Thank you for your proposal. It would be helpful if you could kindly explain how to find where this transcoder is used as this implementation doesn't work currently and I still have some doubt about the API (please see the discussion in the issue).

vados-cosmonic:
Hey so I think that we'd need to write a component that *needs* this transcoding implicitly -- so conceptually a component which lift/lowers a utf8 string into a utf16 string. 

There are probably two ways to do this "easily":
- Write WAT that performs a transcode using component builtins
- Write a [C# component](https://github.com/bytecodealliance/componentize-dotnet) (UTF16 by default IIRC) that interacts with a  non-C# component which uses UTF8. This should work *unless* `componentize-dotnet` actually just switches everything to utf8 internally before lifting values out.

I say "easy" but those are both pretty hard :sweat_smile: , I think we can [steal from the greats here](https://github.com/bytecodealliance/wasmtime/blob/main/tests/misc_testsuite/component-model/strings.wast) and edit one of those component WATs (and pull it onto `test/fixtures`) into one that would call the transcoding.

peter-jerry-ye:
@vados-cosmonic Hi, I've added a `lib.wasm`. If you run it with `wasmtime lib.wasm "asdf中文🀄️⏰"`, it should return `"asdf中文🀄️⏰asdf中文🀄️⏰\n"`. You can verify this with wasmtime.

Under the hood is C accepting input, calling a double function through WIT which is provided by MoonBit (using UTF16), and print that response out, so it has UTF8 -> UTF16 -> UTF8.

With that said, this example currently doesn't work with this PR, so could you please take a look at what might be the issue?

vados-cosmonic:
Hey @peter-jerry-ye thanks for the speedy addition of a new test case -- that looks fantastic. Would you mind contributing the WAT instead of the WASM? I'm a bit curious myself what this ends up looking like and think it would be instructive for me and future contributors.

That said, what error did you run into? Would you mind pasting that? It looks like we'll need to add some machinery to run all the transcoding tests in the new dir you added

peter-jerry-ye:
The file is quite large: it's 19774 lines with `wasm-tools print`, 1.2M in size, so I think it would be better to keep as it is?

If I run in the directory, i.e. `./src/jco.js run test/fixtures/transcoding/lib.wasm "asdf中文🀄️⏰"`, I get
```
RuntimeError: memory access out of bounds
    at wasm://wasm/97f8b602:wasm-function[16]:0x8a5
    at wasm://wasm/97f8b602:wasm-function[24]:0x8f0
    at wasm://wasm/97f8b602:wasm-function[33]:0x9ca
    at wasm://wasm/97f8b602:wasm-function[36]:0xa57
    at wasm://wasm/5df4192a:wasm-function[8]:0x207
    at indirect-peter-jerry-ye:experiment/foo@0.1.0-double (wasm://wasm/2a260dea:wasm-function[0]:0xf2)
    at peter_jerry_ye_experiment_foo_double (wasm://wasm/000e85f2:wasm-function[21]:0x637)
    at main (wasm://wasm/000e85f2:wasm-function[22]:0x6ff)
    at __main_void (wasm://wasm/000e85f2:wasm-function[32]:0x2fa7)
    at _start (wasm://wasm/000e85f2:wasm-function[17]:0x3f0)
```

If I run in another directory, I get 

```
TypeError: run020Run is not iterable (cannot read property Symbol(Symbol.iterator))
    at wasm://wasm/5df4192a:wasm-function[8]:0x297
    at wit-component:shim.indirect-peter-jerry-ye:experiment/foo@0.1.0-double (wasm://wasm/wit-component:shim-c50c8466:wasm-function[0]:0xf2)
```

The position may be different from the committed one as I'm adjusting around stuff.

vados-cosmonic:
> The file is quite large: it's 19774 lines with wasm-tools print, 1.2M in size, so I think it would be better to keep as it is?

Ah I see -- yeah you're right that's quite large, a minimal legible example would be preferable, but the binary is certainly better in this case!

> The memory access out of bounds happen on the MoonBit side, and I think it's because it got a wrong ptr.

Ah OK yeah do you think this is implementation related? If MoonBit is getting a wrong pointer there may be some code that's going wrong in the trampoline. Will take another look and see if I can find anything obvious.

> If I run in another directory, I get

This I'm a bit worried about -- the error certainly shouldn't be different from folder to folder. Did you run `./src/jco.js` from somewhere else, or just `jco` as installed on the system?

vados-cosmonic:
OK so I *think* you might have the wrong args for your trampoline: 

https://github.com/bytecodealliance/wasmtime/blob/9da52ede33a9f8996832b0210579ef15296addbd/crates/environ/src/component.rs#L164

You *should* be dealing with a function signature that looks more like this for utf16->utf8

```
            utf16_to_utf8(src: ptr_u16, src_len: size, dst: ptr_u8, dst_len: size, ret2: ptr_size) -> size;
```

You can peek at the wasmtime implementation for some hints on what the impl should look like:

https://github.com/bytecodealliance/wasmtime/blob/9da52ede33a9f8996832b0210579ef15296addbd/crates/wasmtime/src/runtime/vm/component/libcalls.rs#L441

(note the discrepancy in having ret ptr is [an usptream trampoline quirk](https://github.com/bytecodealliance/wasmtime/blob/9da52ede33a9f8996832b0210579ef15296addbd/crates/cranelift/src/compiler/component.rs#L1545))

peter-jerry-ye:
I would like to find out how the trampoline is used, but I can't tell from the generated code as it seems to be a Wasm using the function. I still hesitate about the API having `from_ptr, len, to_ptr` and not using `realloc`. 

> This I'm a bit worried about -- the error certainly shouldn't be different from folder to folder. Did you run ./src/jco.js from somewhere else, or just jco as installed on the system?

I ran the `./src/jco.js` from somewhere else. I'm not quite sure if this is the correct way to use the developing version?

vados-cosmonic:
> I would like to find out how the trampoline is used, but I can't tell from the generated code as it seems to be a Wasm using the function. I still hesitate about the API having `from_ptr, len, to_ptr` and not using `realloc`.

This should hopefully be answered by [my previous comment ](https://github.com/bytecodealliance/jco/pull/594#issuecomment-2720016485)

> I ran the `./src/jco.js` from somewhere else. I'm not quite sure if this is the correct way to use the developing version?

This is definitely a bug IMO -- that is definitely the correct way to use the developing version, and I wouldn't expect it to give you different errors based on where you ran it from! 

Would you mind creating a separate issue for that? It's worth tracking down separately.

peter-jerry-ye:
Thank you for your help @vados-cosmonic ! Now the PR should work, and I added the test to `cli.js`.

vados-cosmonic:
Excellent job @peter-jerry-ye really appreciate you doing the hard work here, glad to have some of those transcoding functions get filled out!

PR #599, opened 2025-03-13
feat(p2-shim): add easy to use p2 shim object

vados-cosmonic:
This commit introduces an import for `preview2-shim` that can be used to more easily fulfill requirements for `instantiate` calls that are exported from transpiled modules.

After creating a `WasiP2Shim` object, users should be able to call `instantiate(undefined, shim.importObject())` and make use of all WASI interfaces specified (or overriden) on the shim.


guybedford:
This is exactly the right approach, thanks.

Make sure to add this into the package.json file `"exports"` for the preview2-shims as a new `"./instantiation"` export or otherwise.

vados-cosmonic:
Thanks for taking a look -- will finish out the PR and make sure to update the exports!

guybedford:
Sure, just let me know when it is ready for review.

vados-cosmonic:
Hey @guybedford yeah this is ready for review now!

packages/preview2-shim/lib/common/instantiation.js
```diff
@@ -0,0 +1,49 @@
...
+*     }
+* });
+*
+* const component = instantiate(null, customImports.importObject())
+* ```
+*/
+export function imports()  {
+  const shim = new WasiP2Shim();
+  return shim.importObject();
+}
```
guybedford, line 49: Is this really worth the extra definition over documenting using `new WasiP2Shim().importObject()` directly?

packages/preview2-shim/lib/common/instantiation.js
```diff
@@ -0,0 +1,49 @@
+import { WasiP2Shim } from "./wasi-shim.js";
+
+export { WasiP2Shim } from "./wasi-shim.js";
```
guybedford, line 3: Note this can just be directly written `export { WasiP2Shim } from './wasi-shim.js'`.

But if that's all this module does, it seems an unnecessary redirection?

vados-cosmonic: > Note this can just be directly written export { WasiP2Shim } from './wasi-shim.js'.

So it *seems* like this would work but it actually doesn't -- it's bit me in the past. Something that is only `export`ed actually isn't available to the code in the current module.

That said, we can certainly move `WasiP2Shim` (and maybe just rename it `WasiShim`) to this file as well, I'm fine with that.

packages/preview2-shim/lib/common/instantiation.js
vados-cosmonic: Nope, happy to change this!

guybedford: Ah, I meant rather than separately defining this function to encourage users to use that expression instead.

guybedford:
Looks good, can we add a test for this using one of the instantiation outputs?

packages/preview2-shim/lib/common/instantiation.js
vados-cosmonic: Ah gotcha! sure

vados-cosmonic:
Added some tests for the shim usage and a basic overriding case!

packages/preview2-shim/lib/common/instantiation.js
```diff
@@ -0,0 +1,116 @@
...
+ *         }
+ *     }
+ * });
+ *
+ * const component = await instantiate(null, customShim.importObject())
+ * ```
+ *
+ * @class Shim
+ */
+export class Shim {
```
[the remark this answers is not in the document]
vados-cosmonic: absolutely -- can do that, and one thing I noticed is that we're *really* close to having an API similar to the [Node WASI construct](https://nodejs.org/api/wasi.html) so I'll rename `importObject` to `getImportObject` while I'm here.

Also, given that most ecosystem JS code uses caps when the thing is caps (ex. `URL`, etc) -- I'm going to go with `WASIShim`, what do you think?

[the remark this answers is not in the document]
guybedford: Sounds good!

guybedford:
STILL getting the Mac timeout unfortunately!

PR #601, opened 2025-03-18
chore: ensure package-lock.json isn't checked in

vados-cosmonic:
This commit ensures that package-lock.json isn't checked in, as npm version v7 and up uses "hidden" lockfiles in node_modules.


vados-cosmonic:
yeah -- I'm not sure why this test is *only* failing on MacOS -- I wonder if there's a problem with the server itself launching.

PR #606, opened 2025-03-19
feat(examples): add node fetch example

vados-cosmonic:


examples/components/node-fetch/README.md
```diff
@@ -0,0 +1,246 @@
+# WebAssembly Component with NodeJS `fetch()` shim
+
+This repository showcases how to use the Javascript WebAssembly Component toolchain (`jco`) to
+transpile a WebAssembly component and run it in a NodeJS, using [`fetch()`][mdn-fetch] via
+[StarlingMonkey][sm] and [WASI][wasi] HTTP interface ([`wasi:http`][wasi-http]) `outgoing-handler`.
```
guybedford, line 5: It would help to have a little more explanation about the case. It can be confusing to know what is being run and how sometimes especially for those not very familiar with the component model.

Perhaps something like "create a WebAssembly component from JavaScript that initiates a WASI HTTP request when using JavaScripts `fetch`, we can then run this component inside of Node.js, running a virtual JS VM on top of WASI on top of Node.js" or something like that.

guybedford (approved):
Much better, thanks.

PR #609, opened 2025-03-20
Fixing typo in http-hello-world example's package.json

marosset:


vados-cosmonic (approved):
LGTM :rocket: 

Thanks for fixing this @marosset -- a bit weird that this wasn't covered under CI, but I'll look into that!

PR #610, opened 2025-03-21
chore: add prettier config

vados-cosmonic:
It would be beneficial to have a consistent style across the codebase -- this commit adds prettier with some minimal default settings, and applies across the codebase.


.prettierignore
```diff
@@ -0,0 +1,2 @@
+*.wasm
+*.wat
```
guybedford, line 2: Seems like we will always want to ignore `test/runtime` unless we want to run prettier against our tests and have them pass code style by updating the bindgen...!?

guybedford (approved):
I wonder if it's worth aligning code style with the output code of tests.

vados-cosmonic:
I could definitely get behind that -- TS is 4 spaces no matter what [with no intent to change or make configurable](https://github.com/microsoft/TypeScript/issues/4042)

While we're here, what do you think about `jco` adopting TS?

PR #611, opened 2025-03-21
Fix ts stubgen result types

vigoo:


vigoo:
Sorry wanted to open it against a fork

PR #626, opened 2025-04-02
docs: fix rendering for | in code blocks in table

tanishiking:
In https://bytecodealliance.github.io/jco/wit-type-representations.html#options-option

before
Screenshot 2025-04-02 at 17 07 02

after
Screenshot 2025-04-02 at 17 08 11

The `|` was not rendering correctly inside code blocks within tables when processed by `mdbook`. https://github.com/rust-lang/mdBook/issues/637

Ideally, this should be addressed by the upstream. The root cause of this problem in `pulldown-cmark` has recently been fixed. https://github.com/pulldown-cmark/pulldown-cmark/issues/356 However, `mdbook` seems to expose the `pulldown-cmark` types as a public API, and is currently facing difficulties updating its dependencies. https://github.com/rust-lang/mdBook/pull/2386
Therefore, the upstream fix may not available in any time soon.

As a workaround for the time being, use `<code>` tag instead of code blocks in tables with `|`.

### DIFF
--- a/docs/src/wit-type-representations.md
+++ b/docs/src/wit-type-representations.md
@@ -137,8 +137,8 @@ Jco represents options as an optional value or undefined, so some examples:
 
 | Type                  | Representation (TS)                     | Example                               |
 |-----------------------|-----------------------------------------|---------------------------------------|
-| `option<u32>`         | `number | undefined`                    | `option<u32>` -> `number | undefined` |
-| `option<option<u32>>` | `{ tag: "some" | "none", val: number }` | `option<u32>` -> `number | undefined` |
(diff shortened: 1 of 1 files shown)


vados-cosmonic:
Thanks for this fix @tanishiking -- let's put the workaround in. Would you mind making an issue against the repo to update when the upstream PR lands? Not that you have to do it, but just so we keep track of it.

tanishiking:
Sure @vados-cosmonic I opened an issue https://github.com/bytecodealliance/jco/issues/630 (I hope I understand correctly).

vados-cosmonic:
Thanks, that's perfect!

PR #648, opened 2025-04-29
fix: appending to const value raises error

thomas9911:
Small fix, it did an append to a const variable. This crashes the error handling.

### DIFF
--- a/src/cmd/run.js
+++ b/src/cmd/run.js
@@ -117,7 +117,7 @@ async function runComponent(componentPath, args, opts, executor) {
                 '../../../'
             );
         } catch (err) {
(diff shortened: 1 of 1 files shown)


vados-cosmonic (approved):
LGTM :rocket: 

Great catch @thomas9911

vados-cosmonic:
Hey actually @thomas9911 could I get you to change the commit title to something like this:

```
fix(jco): appending to const value raises error
```

I want to make sure this gets attributed correctly to jco in the next outgoing release.

thomas9911:
I encountered the error :D updated the commit message

PR #698, opened 2025-05-14
refactor: project layout

vados-cosmonic:
This PR rearranges the jco project so that it's a bit easier to do more work in either language (Rust/JS) and expain the structure of the jco repository itself.


vados-cosmonic:
This is a massive refactor so to avoid any breaking changes on other PRs, I'm going to merge this fast. 

At present everything looks great, though release automation *is likely to be affected the most* by things moving around. I will handle that in a follow up!

Issue #466, opened 2024-07-08
[preview2-shim] Consider giving guest no preopens by default


dicej:
I was surprised to find that `preview2-shim` gives guests access to the entire host filesystem by default: https://github.com/bytecodealliance/jco/blob/304aa9dc7157e1f31b1695ea26e0c3dec3ecb6b6/packages/preview2-shim/lib/nodejs/filesystem.js#L600.  That's the opposite of most other WASI host implementations, which provide no access to the host filesystem by default.

Is there some specific reason for that?  If not, I'd recommend changing the default to no access and let the user or embedder specify the set of preopens using `_setPreopens` as desired.

More generally, I wonder if other host resources (network access, environment variables) should also be restricted by default and granted explicitly, similar to how e.g. `wasmtime run` and the `wasmtime-wasi` handles such things.

guybedford:
I guess the reasoning is that this is the same default policy a Node.js application has so in the JS ecosystem it's kind of the expectation.

That is, if components are published to npm and loaded as libraries in JS, they should have the same capabilities as JS libraries, which in Node.js means everything.

But, if you are running in instantiation mode, where you're constructing a virtualization, then yes you definitely want all the sandboxing options.

So I'd suggest here that we:

* Have a better way of supporting instantiation mode for preview2-shim
* In that new workflow, support all the various sandboxing options as inputs

How does that sound to you?

dicej:
That sounds fine to me.  The main thing I'm looking for is an easy way to say: "Fully sandbox this component, giving it access only to what I've explicitly granted."  Sounds like instantiation mode would fit the bill.

guybedford:
Great, I'll mark this as an enhancement, happy to try and hack on some workflows here further.

PR #724, opened 2025-06-02
feat: implement wasi preview3 nodejs filesystem API

andreiltd:


.github/workflows/main.yml
```diff
@@ -302,9 +302,12 @@ jobs:
...
         run: cargo test node_
 
-  test-preview2-shims:
-    name: Test Preview2 Shims
+  test-shims:
+    name: Test ${{ matrix.workspace }} Shims
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        workspace: [preview2-shim, preview3-shim]
```
vados-cosmonic:
```suggestion
        workspace: 
          - preview2-shim
          - preview3-shim
```

meganit

packages/preview3-shim/eslint.config.js
```diff
@@ -0,0 +1,11 @@
+import globals from "globals";
```
vados-cosmonic: Forgot to note this earlier, but we should use the top level eslint config here!

packages/preview3-shim/lib/nodejs/filesystem/error.js
```diff
@@ -0,0 +1,85 @@
...
+}
+
+/**
+ * Custom error for File operations with JCO compatible payload.
+ */
+export class FsError extends Error {
+    /**
+     * @param {string} tag        – machine‐readable error tag
+     * @param {string} [message]  – human‐readable message
+     * @param {any}    [val]      – optional extra data
```
vados-cosmonic, line 62:
```suggestion
     * @param {any}    [val]      – optional extra data
     * @param {object} [opts]
     * @param {Error} [opts.cause]
```

Maybe the cause should be an `any`?

packages/preview3-shim/lib/nodejs/filesystem/error.js
```diff
@@ -0,0 +1,85 @@
...
+     */
+    constructor(tag, message, val, opts = {}) {
+        super(message ?? `Error: ${tag}`, { cause: opts.cause });
+        this.name = 'FsError';
+        this.payload = val !== undefined ? { tag, val } : { tag };
+    }
+
+    /**
+     * Create or rewrap an error into SocketError
```
vados-cosmonic:
```suggestion
     * Create or rewrap an error into FSError
```

Oh *also*, in JS people generally keep acronyms in upper case (ex. `HTTPServer`)

packages/preview3-shim/lib/nodejs/filesystem/error.js
```diff
@@ -0,0 +1,85 @@
...
+     * @param {any} err – number, string, Error, or SocketError
```
vados-cosmonic:
```suggestion
     * @param {any} err – number, string, Error, or FSError
```

packages/preview3-shim/lib/nodejs/filesystem/error.js
```diff
@@ -0,0 +1,85 @@
...
+    ETXTBSY: 'text-file-busy',
+    EXDEV: 'cross-device',
+};
+
+function mapError(e) {
+    if (e.code in ERROR_MAP) {
+        return ERROR_MAP[e.code];
+    }
+
+    if (e.code === -4094 || (e.code === 'UNKNOWN' && e.errno === -4094)) {
```
vados-cosmonic, line 48: Would be nice to have a constant + note about this magic number, and when it might reasonably be this

packages/preview3-shim/lib/nodejs/filesystem/error.js
```diff
@@ -0,0 +1,85 @@
...
+
+        const tag = mapError(err);
+        const message = err?.message;
+        return new FsError(tag, message, undefined, { cause: err });
+    }
+}
+
+export const types = {
+    filesystemErrorCode(err) {
+        return mapError(err.payload);
```
vados-cosmonic: Should this be named `errorCode`s? or `errorType`s? `types` is a very general export name

packages/preview3-shim/lib/nodejs/finalization.js
```diff
@@ -0,0 +1,23 @@
+const registry = new FinalizationRegistry((dispose) => dispose());
+const dummySymbol = Symbol();
```
vados-cosmonic: NIT: Could we make this if not present inside the function? Also a bit late but constants usually get the `ALL_CAPS` treatment.

packages/preview3-shim/lib/nodejs/future.js
```diff
@@ -0,0 +1,156 @@
...
+    #consumed = false;
+
+    /**
+     * Constructs a FutureReader.
+     *
+     * @param {Promise} promise - The promise to be read.
+     * @throws {Error} If the provided argument is not a promise.
+     */
+    constructor(promise) {
+        if (!(promise instanceof Promise)) {
```
vados-cosmonic: I think the definitive way to check for this is whether it has a `.then` on a object or fn

packages/preview3-shim/lib/nodejs/future.js
```diff
@@ -0,0 +1,156 @@
...
+            throw new Error('Provided future must be a Promise');
+        }
+        this.#promise = promise;
+    }
+
+    /**
+     * Reads the value from the promise. Can only be called once.
+     *
+     * @returns {Promise<*>} Resolves with the value of the promise or null if already consumed.
```
vados-cosmonic:
```suggestion
     * @returns {Promise<any>} Resolves with the value of the promise or null if already consumed.
```

consistenly `any` here? let's stick close to TS syntax

packages/preview3-shim/lib/nodejs/future.js
```diff
@@ -0,0 +1,156 @@
...
+    async read() {
+        if (this.#consumed) return null;
+        try {
+            const value = await this.#promise;
+            this.#consumed = true;
+            return value;
+        } catch (err) {
+            this.#consumed = true;
+            throw err;
+        }
```
vados-cosmonic:
```suggestion
        try {
            return await this.#promise;
        } catch (err) {
            throw err;
        } finally {
            this.#consumed = true;
        }
```
nit

packages/preview3-shim/lib/nodejs/future.js
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    /**
+     * Cancels the future. Not supported for FutureReader.
+     */
+    async cancel() {
+        // FutureReader does not support cancel
```
vados-cosmonic:
```suggestion
        // FutureReader does not support cancel
```

What do you think about throwing here instead of having things seem like they work?

packages/preview3-shim/lib/nodejs/future.js
```diff
@@ -0,0 +1,156 @@
...
+    close() {
+        // Nothing to release.
+    }
+
+    /**
+     * Converts the reader back into a promise.
+     *
+     * @returns {Promise} The original promise.
+     */
+    intoFuture() {
```
vados-cosmonic:
```suggestion
    intoPromise() {
```

Also I wonder if we should set `this.#promise` to null afterwards to hollow out this object?

packages/preview3-shim/lib/nodejs/future.js
```diff
@@ -0,0 +1,156 @@
...
+        this.close();
+        return this.#promise;
+    }
+}
+
+export class FutureWriter {
+    #deferred;
+    #written = false;
```
vados-cosmonic: comments and conditions on which these are set/what they are would be great

packages/preview3-shim/lib/nodejs/future.js
```diff
@@ -0,0 +1,156 @@
...
+     *
+     * @param {Object} deferred - A deferred object with `resolve` and `reject` methods.
+     * @throws {Error} If the deferred object is invalid.
+     */
+    constructor(deferred) {
+        if (!deferred || typeof deferred.resolve !== 'function') {
+            throw new Error(
+                'Provided deferred object must have a resolve method'
+            );
+        }
```
vados-cosmonic, line 120: can we make this for the user if it's not present?

packages/preview3-shim/lib/nodejs/future.js
```diff
@@ -0,0 +1,156 @@
...
+     *
+     * @returns {Promise} The original promise.
+     */
+    intoFuture() {
+        this.close();
+        return this.#promise;
+    }
+}
+
+export class FutureWriter {
```
vados-cosmonic, line 103:
```suggestion
class FutureWriter {
```

What do you think about not even making this public?

packages/preview3-shim/lib/nodejs/future.js
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    /**
+     * Resolves the promise with the given value.
+     *
+     * @param {*} value - The value to resolve the promise with.
+     * @throws {Error} If the writer is already closed.
+     */
+    async write(value) {
+        if (this.#written) throw new Error('FutureWriter is closed');
```
vados-cosmonic:
```suggestion
        if (this.#written) throw new Error('FutureWriter is closed (already written)');
```

packages/preview3-shim/lib/nodejs/future.js
```diff
@@ -0,0 +1,156 @@
...
+                'Provided deferred object must have a resolve method'
+            );
+        }
+        this.#deferred = deferred;
+    }
+
+    /**
+     * Resolves the promise with the given value.
+     *
+     * @param {*} value - The value to resolve the promise with.
```
vados-cosmonic:
```suggestion
     * @param {any} value - The value to resolve the promise with.
```

NIT: How do you feel about consistently using `any`?

packages/preview3-shim/lib/nodejs/future.js
```diff
@@ -0,0 +1,156 @@
...
+     * @throws {Error} If the writer is already closed.
+     */
+    async write(value) {
+        if (this.#written) throw new Error('FutureWriter is closed');
+        this.#written = true;
+        this.#deferred.resolve(value);
```
vados-cosmonic, line 133: Don't know if I have a concrete suggestion here, but I *wonder* if it's possible try to re-resolve `this.#deferred` here. Clearly it wouldn't take this path, but if I gave the constructor an already resolved promise, this would fail I think.

I think this is fixed by hiding the class itself (making it private), since then people are forced to use `future()` and it won't happen but this seems  little optimistic to me.

packages/preview3-shim/lib/nodejs/future.js
```diff
@@ -0,0 +1,156 @@
...
+        this.#deferred.resolve(null);
+    }
+
+    /**
+     * Rejects the promise with the given error and closes the writer.
+     *
+     * @param {Error} error - The error to reject the promise with.
+     */
+    async closeWithError(error) {
+        if (this.#written) return;
```
vados-cosmonic: Wonder if we should throw an error here -- someone should not attempt to close with (or without) errors if it's already been written to. 

Also, it would be nice to be able to tell the difference between a close with a write and without one.

packages/preview3-shim/lib/nodejs/future.js
```diff
@@ -0,0 +1,156 @@
...
+        this.#written = true;
+        this.#deferred.reject(error);
+    }
+
+    /**
+     * Converts the writer back into a promise.
+     *
+     * @returns {Promise} The original promise.
+     */
+    intoFuture() {
```
vados-cosmonic:
```suggestion
    intoPromise() {
```

May want to also hollow out this object somehow/ensure we don't have the reference in multiple places as we we don't have the usual Rust protections here.

packages/preview3-shim/lib/nodejs/future.js
```diff
@@ -0,0 +1,156 @@
...
+        this.close();
+        return this.#deferred.promise;
+    }
+}
+
+/**
```
vados-cosmonic: nit: this should be at the top, as it's just about the first thing any implementer will want to see/understand.

An example would also *really* help people IMO -- a quick example of how to use a future to do something useful (ex. process a `fetch` or something)

packages/preview3-shim/lib/nodejs/index.js
```diff
@@ -0,0 +1,5 @@
+import * as filesystem from './filesystem.js';
+import * as stream from './stream.js';
+import * as future from './future.js';
+
+export { filesystem, future, stream };
```
vados-cosmonic:
```suggestion
export * as filesystem from './filesystem.js';
export * as stream from './stream.js';
export * as future from './future.js';
```

I think this should work

packages/preview3-shim/lib/nodejs/stream.js
```diff
@@ -0,0 +1,185 @@
+export class StreamReader {
```
vados-cosmonic, line 44:
```suggestion
class StreamReader {
```

Same suggestion as in writer

packages/preview3-shim/lib/nodejs/stream.js
```diff
@@ -0,0 +1,185 @@
...
+            this.#reader = null;
+        }
+    }
+
+    /**
+     * Converts the reader back into a readable stream.
+     *
+     * @returns {ReadableStream} The original readable stream.
+     */
+    intoStream() {
```
vados-cosmonic:
```suggestion
    intoReadableStream() {
```

Also, we kind of check for anything that's *shaped like* a readable stream, so maybe that's worth calling out in the docs -- the thing you get back could actually not be a `ReadableStream`, but just something that had `getReader()` (which is a useful feature IMO)

packages/preview3-shim/lib/nodejs/stream.js
```diff
@@ -0,0 +1,185 @@
...
+    }
+
+    /**
+     * Closes the reader and releases the lock on the stream.
+     */
+    close() {
+        if (this.#reader) {
+            this.#reader.releaseLock();
+            this.#reader = null;
+        }
```
vados-cosmonic:
```suggestion
        if (!this.#reader) { return; }
        this.#reader.releaseLock();
        this.#reader = null;
```

vados-cosmonic: Also I *think* we should check for the `releaseLock` function before assuming it's there -- it's actually not technically present on `ReadableStream` but instead on the officially provided (default) subclass. 

IMO this is less important if people can't make readers for themselves of course!

packages/preview3-shim/lib/nodejs/stream.js
```diff
@@ -0,0 +1,185 @@
...
+        } finally {
+            this.close();
+        }
+    }
+
+    /**
+     * Cancels the stream with the given reason.
+     *
+     * @param {*} reason - The reason for canceling the stream.
+     * @returns {Promise<void>}
```
vados-cosmonic: SUPER MEGA NIT: `cancel` should return `undefined`

packages/preview3-shim/lib/nodejs/stream.js
```diff
@@ -0,0 +1,185 @@
...
+        return value;
+    }
+
+    /**
+     * Reads all chunks from the stream and concatenates them into a buffer.
+     *
+     * @returns {Promise<Buffer>} Resolves with the concatenated buffer.
+     */
+    async readAll() {
+        const chunks = [];
```
vados-cosmonic, line 90: Should we check/throw if the stream is closed?


NIT: Also I'm not 100% sure this will work, but [utility consumers in node](https://nodejs.org/api/webstreams.html#utility-consumers) ? IMO the way you have it right now is actually just fine because it would work well for the browser. BUT we're using stuff like `releaseLock` which is node specific...

*If* utility consumers make sense I'm worried about the following cases:
- partial read (what if someone calls `read()` then `readAll()`?)
- double close (not sure if the stream consumer helper will close)

packages/preview3-shim/lib/nodejs/stream.js
```diff
@@ -0,0 +1,185 @@
...
+
+    /**
+     * Constructs a StreamReader.
+     *
+     * @param {ReadableStream} readable - The readable stream to consume.
+     * @throws {Error} If the provided stream does not implement `getReader`.
+     */
+    constructor(readable) {
+        if (!readable || typeof readable.getReader !== 'function') {
+            throw new Error('Provided readable must implement getReader()');
```
vados-cosmonic:
```suggestion
            throw new Error('failed to create StreamReader: provided readable must implement getReader()');
```

What do you think about some sort of indication of where this comes from? `StreamReader.constructor <actual message>` ? I just want to see this in the logs and instantly went wrong (Feels like I might ask "readable provided to what?").

packages/preview3-shim/lib/nodejs/stream.js
```diff
@@ -0,0 +1,185 @@
...
+    }
+
+    #ensureWriter() {
+        if (this.#writer === null) {
+            throw new Error('StreamWriter is closed');
+        }
+    }
+}
+
+export function stream() {
```
vados-cosmonic: comments similar to future would be great here

packages/preview3-shim/lib/nodejs/stream.js
```diff
@@ -0,0 +1,185 @@
...
+    const transform = new TransformStream(
+        {},
+        { highwatermark: 64 * 1024 },
+        { highWaterMark: 64 * 1024 }
```
vados-cosmonic: Would users ever want to configure this? It might be nice if they could pass their own streams in

packages/preview3-shim/lib/nodejs/stream.js
```diff
@@ -0,0 +1,185 @@
...
+            this.#writer = null;
+        }
+    }
+
+    /**
+     * Converts the writer back into a writable stream.
+     *
+     * @returns {WritableStream} The original writable stream.
+     */
+    async intoStream() {
```
vados-cosmonic:
```suggestion
    async intoWritableStream() {
```

Same note about possibly hollowing out this object/ensuring there are fewer references to the underlying stream.

packages/preview3-shim/lib/nodejs/stream.js
```diff
@@ -0,0 +1,185 @@
...
+     * Closes the writer with an error.
+     *
+     * @param {Error} error - The error to close the writer with.
+     * @returns {Promise<void>}
+     */
+    async closeWithError(error) {
+        if (this.#writer) {
+            await this.#writer.abort(error);
+            this.#writer = null;
+        }
```
vados-cosmonic:
```suggestion
        if (!this.#writer) { return; }
        await this.#writer.abort(error);
        this.#writer = null;
```

Also, I wonder if this should error on an attempt to close twice.

packages/preview3-shim/lib/nodejs/stream.js
```diff
@@ -0,0 +1,185 @@
...
+     * Closes the writer and the stream.
+     *
+     * @returns {Promise<void>}
+     */
+    async close() {
+        if (this.#writer) {
+            await this.#writer.close();
+        }
+
+        this.#writer = null;
```
vados-cosmonic:
```suggestion
        if (!this.#writer) { return; }
        try {
            await this.#writer.close();
        } catch(err) {
            throw err;
        } finally { 
            this.#writer = null;
        }
```

The `try`/`finally` might be overkill

packages/preview3-shim/lib/nodejs/workers/filesystem-worker.js
```diff
@@ -0,0 +1,96 @@
+import fs from 'fs';
+import { promisify } from 'util';
+import { Router } from '../workers/resource-worker.js';
+
+const readAsync = promisify(fs.read);
+const writeAsync = promisify(fs.write);
+const { opendir } = fs.promises;
+
+const CHUNK = 64 * 1024;
```
vados-cosmonic: this is clearly bytes, but what do you think about naming it `CHUNK_BYTES`?

packages/preview3-shim/lib/nodejs/workers/filesystem-worker.js
```diff
@@ -0,0 +1,96 @@
...
+const buffer = Buffer.alloc(CHUNK);
```
vados-cosmonic: `BUFFER` would be great here for a global -- and a comment that the buffer is reused in certain operations (though it's fairly obvious)

packages/preview3-shim/lib/nodejs/workers/filesystem-worker.js
```diff
@@ -0,0 +1,96 @@
+import fs from 'fs';
```
vados-cosmonic: Can we not use `node:fs/promises` here? Was this the point where we had to use the sync version because of the support for `bigint`? If so, a comment (and grouping this next to the fs stuff) would go a long way!

packages/preview3-shim/lib/nodejs/workers/filesystem-worker.js
```diff
@@ -0,0 +1,96 @@
...
+const buffer = Buffer.alloc(CHUNK);
+
+/** Auto‐dispatch all ops */
+Router()
+    .op('read', handleRead)
+    .op('write', handleWrite)
+    .op('readDir', handleReadDir);
+
+/** Map fs.Dirent → WASI type */
+function lookupType(obj) {
```
vados-cosmonic: probably could have a better name here

packages/preview3-shim/lib/nodejs/workers/filesystem-worker.js
```diff
@@ -0,0 +1,96 @@
...
+        while (true) {
+            const { bytesRead } = await readAsync(fd, buffer, 0, CHUNK, pos);
+            if (bytesRead === 0) break;
+
+            await writer.write(buffer.subarray(0, bytesRead));
+            pos += BigInt(bytesRead);
+        }
+
+        await writer.close();
+        return { ok: true };
```
vados-cosmonic: Wonder if there's much use returning `ok: true` if we throw when things go wrong anyway. 

That said, if we're going to return something, maybe returning the total number of bytes read?

packages/preview3-shim/lib/nodejs/workers/filesystem-worker.js
```diff
@@ -0,0 +1,96 @@
...
+            const buf = Buffer.from(value);
+            const { bytesWritten } = await writeAsync(
+                fd,
+                buf,
+                0,
+                buf.length,
+                pos
+            );
+            pos += BigInt(bytesWritten);
+            total += BigInt(bytesWritten);
```
vados-cosmonic: coerce once, write twice? Unfortunate that `writeAsync` doesn't return a `BigInt` by default

Also, isn't `total == pos - offset` ? We may not need to store is separately

Also, we may want some common sense checks at the top on fd, offset and stream, just to exit early before we do something we didn't expect

packages/preview3-shim/lib/nodejs/workers/filesystem-worker.js
```diff
@@ -0,0 +1,96 @@
...
+        }
+        return { ok: true, bytesWritten: total };
+    } catch (err) {
+        await reader.cancel(err);
+        throw err;
+    }
+}
+
+async function handleReadDir({ fullPath, stream }) {
+    const writer = stream.getWriter();
```
vados-cosmonic, line 113: idea: Another path check here? The worker doesn't know about preopens, but maybe it should?

Really not sure if we need this extra check since the path up until now should be pretty bulletproof.

packages/preview3-shim/lib/nodejs/workers/filesystem-worker.js
```diff
@@ -0,0 +1,96 @@
...
+
+    try {
+        const walker = await opendir(fullPath);
```
vados-cosmonic, line 116: Do we need to clean up `walker`?

packages/preview3-shim/lib/nodejs/workers/resource-worker.js
```diff
@@ -0,0 +1,142 @@
...
+    #worker = null;
+    #workerUrl = null;
+    #pending = new Map();
+
+    constructor(workerUrl) {
+        this.#workerUrl = workerUrl;
+    }
+
+    #getWorker() {
+        // TODO(tandr): Use a pool of workers instead of a single one.
```
vados-cosmonic: IMO you can make this an issue and remove the line

packages/preview3-shim/lib/nodejs/workers/resource-worker.js
```diff
@@ -0,0 +1,142 @@
...
+    /** Async IPC */
+    run(msg, transferable = []) {
+        const worker = this.#getWorker();
+        const { port1, port2 } = new MessageChannel();
+
+        return new Promise((resolve, reject) => {
+            port1.once('message', ({ result, error }) => {
+                port1.close();
+                if (error) reject(error);
+                else resolve(result);
```
vados-cosmonic:
```suggestion
                if (error) {
                    reject(error);
                    return;
                }
                resolve(result);
```

packages/preview3-shim/lib/nodejs/workers/resource-worker.js
```diff
@@ -0,0 +1,142 @@
...
+                ...transferable,
+            ]);
+        });
+    }
+
+    /** Synchronous IPC */
+    runSync(msg, transferable = []) {
+        const worker = this.#getWorker();
+
+        const { port1, port2 } = new MessageChannel();
```
vados-cosmonic: think we could use `tx` and `rx` here?

packages/preview3-shim/package.json
```diff
@@ -0,0 +1,46 @@
+{
+  "name": "@bytecodealliance/preview3-shim",
+  "version": "0.1.0",
+  "description": "WASI Preview3 shim for JS environments",
+  "author": "Guy Bedford, Eduardo Rodrigues<16357187+eduardomourar@users.noreply.github.com, Tomasz Andrzejak>",
```
vados-cosmonic: I think your name got a `>` stuck in it, and could you use `authors`, and I think you can remove Guy and Eduardo

packages/preview3-shim/package.json
```diff
@@ -0,0 +1,46 @@
...
+  "type": "module",
+  "types": "./types/index.d.ts",
```
vados-cosmonic, line 7: there are no types, so we we should probably remove the `types` configurations here, and maybe make an issue to add typings sometime down the road

packages/preview3-shim/package.json
```diff
@@ -0,0 +1,46 @@
...
+  "scripts": {
+    "test": "vitest --run",
+    "bench": "vitest bench --run"
+  },
+  "files": [
+    "types",
+    "lib"
+  ],
+  "dependencies": {
+    "@bytecodealliance/preview2-shim": "file:../preview2-shim"
```
vados-cosmonic:
```suggestion
    "@bytecodealliance/preview2-shim": "file:../preview2-shim"
```
We won't be able to publish with this reference (or actually what happens is you publish a package that breaks after download) -- could you depend on `@bytecodealliance/preview2-shim` here?

packages/preview3-shim/package.json
```diff
@@ -0,0 +1,46 @@
...
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/bytecodealliance/jco.git"
+  },
+  "license": "(Apache-2.0 WITH LLVM-exception)",
+  "bugs": {
+    "url": "https://github.com/bytecodealliance/jco/issues"
+  },
+  "homepage": "https://github.com/bytecodealliance/jco#readme"
```
vados-cosmonic: This should probably point to the preview3-shim README

packages/preview3-shim/test/filesystem.test.js
```diff
@@ -0,0 +1,525 @@
...
+    beforeAll,
+    afterAll,
+    beforeEach,
+    afterEach,
+    vi,
+} from 'vitest';
+import os from 'node:os';
+import path from 'node:path';
+import { promises as fsPromises } from 'fs';
+import { fileURLToPath } from 'node:url';
```
vados-cosmonic:
```suggestion
import os from 'node:os';
import path from 'node:path';
import * as fs from 'node:fs/promises';
import { fileURLToPath } from 'node:url';

import {
    describe,
    test,
    expect,
    beforeAll,
    afterAll,
    beforeEach,
    afterEach,
    vi,
} from 'vitest';
```

packages/preview3-shim/lib/nodejs/filesystem/descriptor.js
```diff
@@ -0,0 +1,976 @@
...
+import { earlyDispose, registerDispose } from '../finalization.js';
+
+const symbolDispose = Symbol.dispose || Symbol.for('dispose');
+
+const _worker = new ResourceWorker(
+    new URL('../workers/filesystem-worker.js', import.meta.url)
+);
+
+class Descriptor {
+    // Node.js file handle for file operations (FileHandle from fs.promises.open)
```
vados-cosmonic:
```suggestion
    /** Node.js file handle for file operations (FileHandle from fs.promises.open) */
```
Super nit but IIRC JSDoc has this syntax for single lines

packages/preview3-shim/lib/nodejs/filesystem/descriptor.js
```diff
@@ -0,0 +1,976 @@
...
+
+        if (size > Number.MAX_SAFE_INTEGER) {
+            throw new FsError('invalid');
+        }
+
+        try {
+            await this.#handle.truncate(Number(size));
+        } catch (e) {
+            if (process.platform === 'win32' && e.code === 'EPERM')
+                throw new FsError('access');
```
vados-cosmonic:
```suggestion
            if (process.platform === 'win32' && e.code === 'EPERM') {
                throw new FsError('access');
            }
```

nit

packages/preview3-shim/lib/nodejs/filesystem/descriptor.js
```diff
@@ -0,0 +1,976 @@
...
+        } else {
+            this.#ensureHandle();
+            try {
+                await this.#handle.sync();
+            } catch (e) {
+                // On windows, `sync_data` uses `FileFlushBuffers` which fails with `EPERM` if
+                // the file is not upen for writing. Ignore this error, for POSIX compatibility.
+                if (process.platform === 'win32' && e.code === 'EPERM') return;
+                throw FsError.from(e);
+            }
```
vados-cosmonic: NIT: what do you think about moving these branches out to diff functions -- one for preopens and the other for non-preopens? Even moving just one branch IMO would greatly simplify this (and you can do an early return)

packages/preview3-shim/lib/nodejs/filesystem/descriptor.js
```diff
@@ -0,0 +1,976 @@
...
+     * @param {object} atimeDesc New access timestamp.
+     * @param {object} mtimeDesc New modification timestamp.
+     * @returns {Promise<void>}
+     * @throws {FsError} `payload.tag` contains mapped WASI error code.
+     */
+    async statAt(flags, path) {
+        const full = this.#getFullPath(path, flags.symlinkFollow);
+        try {
+            const fn = flags.symlinkFollow ? fs.stat : fs.lstat;
+            const s = await fn(full, { bigint: true });
```
vados-cosmonic:
```suggestion
            const statFn = flags.symlinkFollow ? fs.stat : fs.lstat;
            const s = await statFn(full, { bigint: true });
```

packages/preview3-shim/lib/nodejs/filesystem/descriptor.js
```diff
@@ -0,0 +1,976 @@
...
+
+        if (!flags.symlinkFollow && !fs.lutimes) {
+            throw new FsError('unsupported');
+        }
+
+        const full = this.#getFullPath(path, flags.symlinkFollow);
+
+        try {
+            const fn = flags.symlinkFollow ? fs.utimes : fs.lutimes;
+            await fn(full, atime, mtime);
```
vados-cosmonic:
```suggestion
            const utimesFn = flags.symlinkFollow ? fs.utimes : fs.lutimes;
            await utimesFn(full, atime, mtime);
```

packages/preview3-shim/lib/nodejs/filesystem/descriptor.js
```diff
@@ -0,0 +1,976 @@
...
+     * @throws {FsError} `payload.tag` contains mapped WASI error code.
+     */
+    async linkAt(oldFlags, oldPath, newDesc, newPath) {
+        const src = this.#getFullPath(oldPath, oldFlags.symlinkFollow);
+        const dst = newDesc.#getFullPath(newPath, false);
+
+        // On Windows, a trailing '/' means the path is treated as a
+        // directory and hard links to directory are not supported
+        if (process.platform === 'win32' && dst.endsWith('/'))
+            throw new FsError('no-entry');
```
vados-cosmonic:
```suggestion
        if (process.platform === 'win32' && dst.endsWith('/')) {
            throw new FsError('no-entry');
        }
```

packages/preview3-shim/lib/nodejs/filesystem/descriptor.js
```diff
@@ -0,0 +1,976 @@
...
+     * @async
+     * @param {object} pf Path flags.
+     * @param {string} path Relative path.
+     * @param {object} of Open flags.
+     * @param {object} df Descriptor flags.
+     * @returns {Promise<Descriptor>}
+     * @throws {FsError} `payload.tag` contains mapped WASI error code.
+     */
+    async openAt(pf, path, of, df) {
+        if (!preopenEntries.length) throw new FsError('access');
```
vados-cosmonic:
```suggestion
        if (!preopenEntries.length) {
            throw new FsError('access');
        }
```

packages/preview3-shim/lib/nodejs/filesystem/descriptor.js
```diff
@@ -0,0 +1,976 @@
...
+     * ) -> result<descriptor, error-code>
+     * ```
+     *
+     * @async
+     * @param {object} pf Path flags.
+     * @param {string} path Relative path.
+     * @param {object} of Open flags.
+     * @param {object} df Descriptor flags.
+     * @returns {Promise<Descriptor>}
+     * @throws {FsError} `payload.tag` contains mapped WASI error code.
```
vados-cosmonic:
```suggestion
     * @param {object} pf Path flags.
     * @param {string} path - Relative path.
     * @param {object} of - Open flags.
     * @param {object} df - Descriptor flags.
     * @returns {Promise<Descriptor>}
     * @throws {FsError} `payload.tag` contains mapped WASI error code.
```

packages/preview3-shim/lib/nodejs/filesystem/descriptor.js
```diff
@@ -0,0 +1,976 @@
...
+                fullPath.endsWith('/') &&
+                (await desc.getType()) !== 'directory'
+            ) {
+                desc[symbolDispose]();
+                throw new FsError('not-directory');
+            }
+            return desc;
+        } catch (err) {
+            if (err.code === 'ERR_INVALID_ARG_VALUE')
+                throw process.platform === 'win32' ? 'no-entry' : 'invalid';
```
vados-cosmonic:
```suggestion
            if (err.code === 'ERR_INVALID_ARG_VALUE') {
                throw process.platform === 'win32' ? 'no-entry' : 'invalid';
            }
```

packages/preview3-shim/lib/nodejs/filesystem/descriptor.js
```diff
@@ -0,0 +1,976 @@
...
+            }
+        }
+
+        try {
+            const handle = await fs.open(target, fsFlags);
+            const desc = descriptorCreate(handle, df, fullPath);
+
+            if (
+                fullPath.endsWith('/') &&
+                (await desc.getType()) !== 'directory'
```
vados-cosmonic: can we give this condition a name? looks like we're trying to check if it's a file -- but maybe a "non directory" is the accurate way to describe it

packages/preview3-shim/lib/nodejs/filesystem/descriptor.js
```diff
@@ -0,0 +1,976 @@
...
+     * @returns {Promise<void>}
+     * @throws {FsError} `payload.tag` contains mapped WASI error code.
+     */
+    async removeDirectoryAt(path) {
+        const full = this.#getFullPath(path, false);
+        try {
+            await fs.rmdir(full);
+        } catch (e) {
+            if (process.platform === 'win32' && e.code === 'ENOENT')
+                throw new FsError('not-directory');
```
vados-cosmonic:
```suggestion
            if (process.platform === 'win32' && e.code === 'ENOENT') {
                throw new FsError('not-directory');
            }
```

packages/preview3-shim/lib/nodejs/filesystem/descriptor.js
```diff
@@ -0,0 +1,976 @@
...
+                    throw new FsError(tag);
+                }
+            }
+
+            if (
+                process.platform === 'win32' &&
+                ['EPERM', 'EEXIST'].includes(e.code)
+            )
+                throw new FsError('no-entry');
+
```
vados-cosmonic:
```suggestion
            ) {
                throw new FsError('no-entry');
            }

```

packages/preview3-shim/lib/nodejs/filesystem/descriptor.js
```diff
@@ -0,0 +1,976 @@
...
+     * @param {object} flags Path flags.
+     * @param {string} path Relative path.
+     * @returns {Promise<{upper: bigint, lower: bigint}>}
+     * @throws {FsError} `payload.tag` contains mapped WASI error code.
+     */
+    async metadataHashAt(flags, path) {
+        const full = this.#getFullPath(path, false);
+        try {
+            const fn = flags.symlinkFollow ? fs.stat : fs.lstat;
+            const s = await fn(full);
```
vados-cosmonic:
```suggestion
            const statFn = flags.symlinkFollow ? fs.stat : fs.lstat;
            const s = await statFn(full);
```

packages/preview3-shim/lib/nodejs/filesystem/descriptor.js
```diff
@@ -0,0 +1,976 @@
...
+    }
+
+    #ensureHandle() {
+        if (!this.#handle) {
+            throw new FsError('bad-descriptor');
+        }
+    }
+}
+
+function lookupType(obj) {
```
vados-cosmonic: A more descriptive name/comment might be good here... Have I seen this logic before..?

packages/preview3-shim/lib/nodejs/filesystem/descriptor.js
```diff
@@ -0,0 +1,976 @@
...
+
+function stripTrailingSlash(path) {
+    if (path === '/') return '/';
+    return path.replace(/\/+$/, '');
+}
+
+const descriptorCreatePreopen = Descriptor._createPreopen;
+delete Descriptor._createPreopen;
+const descriptorCreate = Descriptor._create;
+delete Descriptor._create;
```
vados-cosmonic: If you could put these a bit closer to the `Descriptor` class that would be great

packages/preview3-shim/lib/nodejs/filesystem/descriptor.js
```diff
@@ -0,0 +1,976 @@
...
+
+export const types = { Descriptor };
```
vados-cosmonic:
```suggestion
export const types = { Descriptor };
```

Could this be removed? Naming exports `types` seems a bit weird, does something need access to the class directly except for via `descriptorCreate`/`descriptorCreatePreopen`?

packages/preview3-shim/lib/nodejs/filesystem/descriptor.js
```diff
@@ -0,0 +1,976 @@
...
+
+const preopenEntries = [];
+export const preopens = {
+    Descriptor,
+    getDirectories: () => preopenEntries,
+};
+
+const ROOT_PREOPEN = process.platform === 'win32' ? '//' : '/';
+_addPreopen('/', ROOT_PREOPEN);
```
vados-cosmonic: I don't think we want this by default. At the very least it should be opt in

https://github.com/bytecodealliance/jco/issues/466

vados-cosmonic (commented):
Thanks for the hard work here @andreiltd and splitting it off to make it easier to review!

packages/preview3-shim/lib/nodejs/filesystem/error.js
andreiltd: Do we care about this to be consistent with preview2?

vados-cosmonic: Definitely not -- I don't know why preview2 would have exposed types like that, and the name `mapError` is pretty unintuitive/not descriptive.

I think being able to look up filesystem error codes *is* useful though, but I'm not sure why it's exposed in this exact way. Might be a case of chesterson's fence, but I would be happy with exporting that function directly.

packages/preview3-shim/lib/nodejs/stream.js
andreiltd: Isn't `releaseLock` web stream feature? https://developer.mozilla.org/en-US/docs/Web/API/ReadableStreamDefaultReader/releaseLock

vados-cosmonic: Ah sorry you're right -- it's definitely available in both places -- I checked the web docs only for `ReadableStream` and missed that default reader was there too.

I think I made this comment elsewhere, but we're not guaranteed to get a `ReadableStreamDefaultReader` which has that, at least with the way we're validating the incoming thing right now.

I'm OK with not using the utility consumers though (some far off point in the future, we'll want to reuse this code for the browser, as much as we can), but the other points should still stand:

- We should probably check if the stream is closed/other invariants before we start doing stuff
- We probably should be a bit more certain we have something with `releaseLock`

packages/preview3-shim/lib/nodejs/filesystem/descriptor.js
andreiltd: Could we configure this in linter?

t3hmrman: Now that we have it, yep!

https://github.com/bytecodealliance/jco/pull/728 will merge soon, and you should be able to rebase to that!

vados-cosmonic: Now that we have it, yep!

https://github.com/bytecodealliance/jco/pull/728 will merge soon, and you should be able to rebase to that!

packages/preview3-shim/lib/nodejs/future.js
andreiltd: Sorry, what do you mean by "make this for the user"?

packages/preview3-shim/lib/nodejs/future.js
andreiltd: I would prefer to have this public so that we allow constructing readable/writable end without `future` util and simply wrapping them with this abstraction.

packages/preview3-shim/lib/nodejs/future.js
andreiltd: Sure!

packages/preview3-shim/lib/nodejs/workers/filesystem-worker.js
andreiltd: Yeah, AFAICT this is the only one allowing for bigint offsets.

packages/preview3-shim/lib/nodejs/workers/filesystem-worker.js
andreiltd: changed to `getFileType` ?

packages/preview3-shim/lib/nodejs/workers/filesystem-worker.js
andreiltd: I would argue that this is a private function and all the types were already checked upstream. Normally I would just assert here but I guess we will throw on mismatched types anyway.

packages/preview3-shim/lib/nodejs/workers/filesystem-worker.js
andreiltd: Same as above, since this is not exposed I think it is ok to assume the types has been checked upstream.

packages/preview3-shim/lib/nodejs/future.js
vados-cosmonic: Ah, so I was thinking here that `deferred` could be created for the user given how `intoPromise()` was implemented (people could use `FutureWriter` to create then pull out the relevant `#deferred`, but now that `intoPromise()` is a bit more strict I think this is fine!

packages/preview3-shim/lib/nodejs/workers/filesystem-worker.js
andreiltd: I get an exception: "already closed" when I try to close explicitly in the `finally` block.

packages/preview3-shim/lib/nodejs/workers/filesystem-worker.js
vados-cosmonic: `toWASIFilesystemType` ? `getWasiTypeForDirent` ? `wasiTypeFromDirent` ? `lookupWASITypeForDirent`? Naming is hard, but basically the goal is to encode the succinct comment on top of the function into the function name so anyone who sees it knows that it's just a simple type mapping between Dirent and the relevant WASI fs type.

In an ideal world this would be something we could attach to the WASI type or `fs.Dirent` via an extension trait (e.g. in Rust), but basically what I ask myself is when I see `getFileType(...)`, do I ask myself any what/who questions.

packages/preview3-shim/lib/nodejs/filesystem/descriptor.js
andreiltd: Yeah, that is in the worker as well. I guess I can import the descriptor file in the worker but not the other way around.

packages/preview3-shim/lib/nodejs/filesystem/descriptor.js
andreiltd: I moved it a bit closer, although it cannot appear  before the `Descriptor` otherwise I'm getting errors.

packages/preview3-shim/lib/nodejs/workers/filesystem-worker.js
vados-cosmonic: Yeah... this is basically our last chance for stopping weird stuff from happening to the actual underlying filesystem -- having checks at this level means no matter how we get here (or how code paths change tomorrow or next week), we're guaranteed to not try to do nonsensical things to the filesystem underneath if a new not-safe codepath is added.

packages/preview3-shim/lib/nodejs/stream.js
andreiltd: I made this optionally configureable,

packages/preview3-shim/lib/nodejs/workers/filesystem-worker.js
vados-cosmonic: Yeah I think that's right for the code paths that exist today -- but it doesn't necessarily hold for tomorrow.

If either existing or new code enabled a way to get a bad path to this point, I sure would like to be able to say that the worker cannot possibly write to a path outside a pre-open, because we always check right before the write -- a bit of defense in depth.

packages/preview3-shim/lib/nodejs/workers/filesystem-worker.js
vados-cosmonic: Ahhh OK so this is handled natively by [`for await`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for-await...of):

> If the for await...of loop exited early (e.g., a break statement is encountered or an error is thrown), the [return()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#the_iterator_protocol) method of the iterator is called to perform any cleanup. The returned promise is awaited before the loop exits.

packages/preview3-shim/lib/nodejs/future.js
andreiltd: WDYT about erroring in the constructor if the promise passed has been already fulfilled?

vados-cosmonic: IMO that's a great idea -- that's almost always a bug right? What could be the valid use case for passing a promise that is already used to try and power this future...

The only thing I could think of is if you did an operation then wanted to pass an already-resolved future (almost like a no-op) future without creating another `Promise` -- but we could support that separately IMO (or if someone actually asks for it/we find some optimization necessary)

packages/preview3-shim/lib/nodejs/stream.js
andreiltd: Isn't the place of throwing tracked by the `Error` instance? If not, sure!

vados-cosmonic: Ah good point -- yeah the constructor call will at least be there, even in a `Promise` chain -- that's probably good enough for people to find where they went wrong.

packages/preview3-shim/lib/nodejs/filesystem/descriptor.js
vados-cosmonic: That's one of the problems with this way of trying to hide those functions -- it actually requires an in order execution of the file top to bottom to work properly, which IMO is not ideal. 

I can't remember exactly why this code is structured exactly this way -- it was/is a p2-shim-ism

packages/preview3-shim/lib/nodejs/workers/filesystem-worker.js
andreiltd: OK we now pass the `preopens` to worker and check the path against it.

packages/preview3-shim/lib/nodejs/workers/filesystem-worker.js
andreiltd: I've added sanity checks.

packages/preview3-shim/lib/nodejs/future.js
andreiltd: This seems to be tricky. I cannot find anything from `Promise` that will indicate that it has been already settled.

packages/preview3-shim/lib/nodejs/filesystem/descriptor.js
andreiltd: I moved this function to separate file so that both `descriptor` and `worker` can access it.

packages/preview3-shim/lib/nodejs/future.js
andreiltd: For example: https://github.com/bytecodealliance/jco/pull/724/files#diff-9bb9cdaabe4a2b2a22d127cf66f753a1e45e8a8d25e8349759abe0200f2301ebR46

vados-cosmonic: Yeah good point -- I'm not dogmatic here -- I think if we end up changing these it'll be a version rev anyway

packages/preview3-shim/lib/nodejs/future.js
vados-cosmonic: Ah, right -- so in this case if we were going to hide the class we could be sure I think, but maybe we can just protect with a try catch here, if it would throw? As the code is written now it doesn't actually mess up the state changes so that's good at least, but might be nice if it didn't throw. 

That said, also should we be guarding against if written is already true?

andreiltd: OK I made the writer private, it is also guarding against `written` with `this.#ensureState()`.

packages/preview3-shim/lib/nodejs/stream.js
vados-cosmonic: hey do you want to do it here too?

andreiltd: Sorry, I think I meant actual stream implementation when I talked about allowing to create whatever readable/writable end and wrapping it in a `StreamReader` and `StreamWriter`. A good example of such usecase can be found here:

https://github.com/bytecodealliance/jco/pull/653/files#diff-854442158280dc8f89dea734ee92b18923ee3adaa5b1dc239e21dd17888e83a4R174

where I want to map incoming entries in the transform stream, for which I need a "custom" stream implementation.

I think we can be a bit more strict with `Futures` though. Currently `FutureReader` is public, `FutureWriter` is private. Does that make sense?

packages/preview3-shim/lib/nodejs/filesystem/error.js
```diff
@@ -0,0 +1,89 @@
...
+    EOVERFLOW: 'overflow',
+    EPERM: 'not-permitted',
+    EPIPE: 'pipe',
+    EROFS: 'read-only',
+    ESPIPE: 'invalid-seek',
+    ETXTBSY: 'text-file-busy',
+    EXDEV: 'cross-device',
+};
+
+function mapError(e) {
```
vados-cosmonic: Can we improve the name here? I wouldn't expect a function that "maps" to throw.

packages/preview3-shim/lib/nodejs/finalization.js
```diff
@@ -0,0 +1,26 @@
+const registry = new FinalizationRegistry((dispose) => dispose());
```
vados-cosmonic: Let's try and avoid module load side effects -- we can delay instantiation of the `registry` var here (and let's call it `REGISTRY` since it's essentially an internal global in this module)

packages/preview3-shim/lib/nodejs/future.js
```diff
@@ -0,0 +1,185 @@
...
+function deferred() {
+    let resolve, reject;
+    const promise = new Promise((res, rej) => {
+        resolve = res;
+        reject = rej;
+    });
+    return { promise, resolve, reject };
+}
+
+export class FutureReader {
```
vados-cosmonic, line 39:
```suggestion
class FutureReader {
```

Was this one not going to be private like the others?

packages/preview3-shim/lib/nodejs/stream.js
vados-cosmonic: That link isn't working, but I think we should be consistent here -- if you need to expose the reader and writer classes, then that's fine! I value the consistency of them being public or not.

IMO if you'd like to have a constructor that allows for a different implementation of the streams to be sent in, then could be a separate function people can call (or other options to the `future()`/`stream()` functions). That said, I'm fine either way here -- let's make them (consistently) public for now, we've bikeshed on this long enough, probably.

vados-cosmonic (approved):
LGTM :rocket:

Some minor comments but this should be ready to merge after they're resolved!

packages/preview3-shim/lib/nodejs/future.js
andreiltd: As per other thread, let's make them all public for now for consistency.

PR #731, opened 2025-06-05
feat(p3-shim): implement wasi preview3 cli API

andreiltd:

This is API is subject to change (see https://github.com/WebAssembly/wasi-cli/issues/65) but for now is consistent with implementation from https://github.com/bytecodealliance/wasip3-prototyping

### DIFF
--- a/packages/preview3-shim/lib/nodejs/cli.js
+++ b/packages/preview3-shim/lib/nodejs/cli.js
@@ -21,9 +21,12 @@ export {
     terminalStderr,
 } from '@bytecodealliance/preview2-shim/cli';
 
-const _worker = new ResourceWorker(
-    new URL('./workers/cli-worker.js', import.meta.url)
-);
+let WORKER = null;
+function worker() {
+    return (WORKER ??= new ResourceWorker(
+        new URL('./workers/cli-worker.js', import.meta.url)
+    ));
+}
 
 export const stdin = {
     /**
@@ -55,7 +58,7 @@ export const stdout = {
      */
     setStdout(streamReader) {
         const stream = streamReader.intoReadableStream();
-        _worker.run({ op: 'stdout', stream }, [stream]);
+        worker().run({ op: 'stdout', stream }, [stream]);
     },
 };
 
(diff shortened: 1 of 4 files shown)


packages/preview3-shim/lib/nodejs/cli.js
```diff
@@ -0,0 +1,77 @@
...
+    terminalInput,
+    terminalOutput,
+    terminalStdin,
+    terminalStdout,
+    terminalStderr,
+} from '@bytecodealliance/preview2-shim/cli';
+
+const _worker = new ResourceWorker(
+    new URL('./workers/cli-worker.js', import.meta.url)
+);
```
vados-cosmonic: would be great if we could avoid side effects here

vados-cosmonic (commented):
Just one quick nit about the worker init!

packages/preview3-shim/lib/nodejs/cli.js
andreiltd: The resource worker is lazily initialized so this doesn't really create a thread worker until you call `run`:

https://github.com/bytecodealliance/jco/blob/main/packages/preview3-shim/lib/nodejs/workers/resource-worker.js#L15

Do we want a double layer of laziness?

vados-cosmonic: Sorry I think I wasn't clear -- it's not that I thought the worker would be created immediately, the fact that the code relies on an effect that is triggered by loading the whole module. This is generally an anti pattern in modern JS ("impure" modules and what not) and is a problem for tree-shaking. 

I think we can remove the double layer of laziness by not having the `ResourceWorker` constructor be implicitly lazy -- now that I think of it I'm not sure what benefit that provides.

andreiltd: Thanks for explanation!

vados-cosmonic: No worries -- this is something we haven't honestly done before (ex. for p2) -- but our discussion the other day about bundle sizes made me really think about other sources of filesize. We're not doing all we could on the `jco` side to reduce the size of transpiled output/etc.

andreiltd:
@vados-cosmonic yes, I think this is expected, all PR p3 PRs are modifying `index.js` file. This PR should be resolved now.

PR #733, opened 2025-06-05
feat(p3-shim): implement wasi preview3 http API

andreiltd:


packages/preview3-shim/lib/nodejs/http/client.js
```diff
@@ -0,0 +1,101 @@
+import { ResourceWorker } from '../workers/resource-worker.js';
+import { StreamReader } from '../stream.js';
+import { FutureReader } from '../future.js';
+import { _fieldsFromEntriesChecked } from './fields.js';
+import { HttpError } from './error.js';
+import { Response } from './response.js';
+
+const _worker = new ResourceWorker(
+    new URL('../workers/http-worker.js', import.meta.url)
+);
```
vados-cosmonic: let's avoid the module side effect here

packages/preview3-shim/lib/nodejs/http/client.js
```diff
@@ -0,0 +1,101 @@
...
+
+export const HttpClient = {
```
vados-cosmonic:
```suggestion
export class HttpClient {
```

packages/preview3-shim/lib/nodejs/http/client.js
```diff
@@ -0,0 +1,101 @@
...
+    /**
+     * Send a Request, return a fully-formed Response object
+     *
+     * @param {Request} req
+     * @returns {Promise<Response>}
+     * @throws {HttpError}
+     */
+    async request(req) {
+        try {
+            const scheme = req.scheme() ?? 'http';
```
vados-cosmonic: can we reduce the scope of this `try`? how much of this can realistically fail?

packages/preview3-shim/lib/nodejs/http/client.js
vados-cosmonic: Sorry, this review was old but I never actually submitted! I think this has already been updated given the other commentary, so resolving!

packages/preview3-shim/lib/nodejs/http/client.js
vados-cosmonic: Looking at this again, it looks like the intended usage is `HttpClient.request`... that looks fine to me, but would love to know the usage we expect from this -- I thought people might be expected to `const client = new HttpClient()` or something.

Was the hope going forward/evolving the object ?

andreiltd: Hey! I don't have a good instinct here. I was hoping to express somehow that the client is stateless and there is no need to create a client object. It may as well be just a function, but I would appreciate suggestions 🙂

vados-cosmonic: IMO how you have it great! Having a client be an object (and be reusable/configurable/etc) is good IMO, it's quite a common pattern.

packages/preview3-shim/lib/nodejs/http/client.js
andreiltd: Yes, I think we can limit it to call to worker.

packages/preview3-shim/lib/nodejs/http/client.js
andreiltd: So should we change this to class?

vados-cosmonic: Yeah -- it's already *almost* a class, so IMO it's worth just making it a class, if we want the wrapper

packages/preview3-shim/lib/nodejs/http/fields.js
```diff
@@ -0,0 +1,292 @@
...
+    }
+}
+
+export const _fieldsLock = Fields._lock;
+delete Fields._lock;
+
+export const _fieldsFromEntriesChecked = Fields._fromEntriesChecked;
+delete Fields._fromEntriesChecked;
+
+export const _forbiddenHeaders = new Set(['connection', 'keep-alive', 'host']);
```
vados-cosmonic: can we get rid of these like the others?

packages/preview3-shim/lib/nodejs/http/request.js
```diff
@@ -0,0 +1,498 @@
...
+
+let REQUEST_CREATE_TOKEN = null;
+function token() {
+    return (REQUEST_CREATE_TOKEN ??= Symbol('RequestCreateToken'));
+}
+
+export class RequestOptions {
+    #connectTimeout = null;
+    #firstByteTimeout = null;
+    #betweenBytesTimeout = null;
```
vados-cosmonic: would be nice if the unites were on these members (e.x. `connectTimeoutMs` or `connectTimeoutSeconds`)

packages/preview3-shim/lib/nodejs/http/request.js
```diff
@@ -0,0 +1,498 @@
...
+    }
+
+    /**
+     * Set the timeout for the initial connect to the HTTP Server.
+     * WIT:
+     * ```
+     * set-connect-timeout: func(duration: option<duration>) -> result<_, request-options-error>
+     * ```
+     *
+     * @param {?number} duration Duration in milliseconds, or `null` to clear
```
vados-cosmonic: duration was a bigint right?

packages/preview3-shim/lib/nodejs/http/request.js
```diff
@@ -0,0 +1,498 @@
...
+            throw new HttpError(
+                'immutable',
+                'Cannot modify immutable RequestOptions'
+            );
+        }
+    }
+}
+
+export const _optionsLock = RequestOptions._lock;
+delete RequestOptions._lock;
```
vados-cosmonic: would be nice to avoid this

packages/preview3-shim/lib/nodejs/http/request.js
```diff
@@ -0,0 +1,498 @@
...
+     * set-path-with-query: func(path-with-query: option<string>) -> result;
+     * ```
+     *
+     * @param {?string} pathWithQuery
+     * @throws {HttpError} with payload.tag 'invalid-syntax' if invalid URI component
+     */
+    setPathWithQuery(pathWithQuery) {
+        if (pathWithQuery != null) {
+            try {
+                new URL(pathWithQuery, 'http://example');
```
vados-cosmonic: is this right?

packages/preview3-shim/lib/nodejs/http/response.js
```diff
@@ -0,0 +1,195 @@
...
+     * ```
+     *
+     * @returns {{ body: StreamReader, trailers: FutureReader }}
+     * @throws {HttpError} with payload.tag 'invalid-state' if body has already been opened or consumed.
+     */
+    body() {
+        if (this.#bodyOpen) {
+            throw new HttpError(
+                'invalid-state',
+                'body() already called and not yet closed'
```
vados-cosmonic:
```suggestion
                'body() already opened and not yet closed'
```

packages/preview3-shim/lib/nodejs/http/response.js
```diff
@@ -0,0 +1,195 @@
...
+     *
+     * WIT:
+     * ```
+     * headers: func() -> headers;
+     * ```
+     *
+     * @returns {Fields}
+     */
+    headers() {
+        return this.#headers;
```
vados-cosmonic, line 130: I think we may not want to give away the actual access to the underlying structure, no?

The field should be immutable (but we can't really enforce that), but at the very least we may want to avoid giving them access to the internal structure, though it would cost a copy.

packages/preview3-shim/lib/nodejs/http/server.js
```diff
@@ -0,0 +1,154 @@
...
+                    await worker().run({
+                        op: 'server-response',
+                        serverId: this.#serverId,
+                        requestId,
+                        statusCode: 500,
+                        headers: [],
+                    });
+                }
+            } catch (err) {
+                console.error('Serve loop failure:', err);
```
vados-cosmonic: what do you think about emitting these errors? and making the server an event emitter?

I wonder if we should be logging right away. This is definitely fine to merge as-is, but curious about what you think

packages/preview3-shim/lib/nodejs/workers/http-worker.js
```diff
@@ -0,0 +1,316 @@
...
+            this._resolvers.push(resolve);
+        });
+    }
+}
+
+// Map<serverId, { server, pending: Queue, inflight: Map }>
+const servers = new Map();
+
+async function handleHttpServerStart({ port, host }) {
+    const serverId = randomUUID();
```
vados-cosmonic: Do we need UUIDs here?

packages/preview3-shim/lib/nodejs/http/request.js
andreiltd: Yes, and it should be nanoseconds instead of milliseconds in the comment. Fixed.

packages/preview3-shim/lib/nodejs/http/request.js
andreiltd: I think so, yes. I just need some base to verify if the path and query are valid URL part.

packages/preview3-shim/lib/nodejs/http/server.js
andreiltd: I had to read about emitters, but yeah it sounds like a good idea. I've changed the HttpServer to extend the emitter.

packages/preview3-shim/lib/nodejs/workers/http-worker.js
andreiltd: Removed!

packages/preview3-shim/lib/nodejs/http/response.js
andreiltd: `this.#headers` are locked in the constructor so any attempt to mutate them will result in throwing `immutable` error.

packages/preview3-shim/lib/nodejs/http/server.js
vados-cosmonic: Yeah it's one of those patterns that works when it works... It's a bit node specific but given the "nature" of this error and how internal it is, it seems to be a good fit here, thanks for trying it out!

packages/preview3-shim/lib/nodejs/http/request.js
vados-cosmonic: Yeah thanks for confirming -- I figured that's what you were doing but wasn't sure... This is almost surely the fastest way to "validate" a path & query. 

Super nit: What do you think about sequestering this functionality to some kind of helper?

packages/preview3-shim/lib/nodejs/http/request.js
vados-cosmonic: *units!

packages/preview3-shim/lib/nodejs/http/response.js
vados-cosmonic: Thanks for pointing that out -- I'm a *little* skeptical, but I think that the amount of trouble someone would have to go through to try and mutate the object is probably not worth attempting to defend against.

And this way avoids a copy so, I'm happy with it. Thanks for the clarification

packages/preview3-shim/lib/nodejs/http/request.js
andreiltd: OK, I created a helper that works for path, authority and scheme.

vados-cosmonic (approved):

PR #735, opened 2025-06-05
feat(p3-shim): implement wasi preview3 sockets API

andreiltd:

### DIFF
--- a/packages/preview3-shim/lib/nodejs/sockets/address.js
+++ b/packages/preview3-shim/lib/nodejs/sockets/address.js
@@ -1,0 +1,84 @@
+// Adapted from preview2-shim/lib/io/worker-sockets.js
+
+// TODO(tandr): switch to generated types
+export const IP_ADDRESS_FAMILY = {
+    IPV4: 'ipv4',
+    IPV6: 'ipv6',
+};
+
+export const tupleToIpv6 = (segments) => {
+    if (segments.length !== 8) {
+        return null;
+    }
+    return segments.map((segment) => segment.toString(16)).join(':');
+};
+
+export const tupleToIpv4 = (segments) => {
+    if (segments.length !== 4) {
+        return null;
+    }
+    return segments.map((segment) => segment.toString(10)).join('.');
+};
+
+export const ipv6ToTuple = (ipv6) => {
+    const [lhs, rhs = ''] = ipv6.includes('::') ? ipv6.split('::') : [ipv6];
+    const lhsParts = lhs === '' ? [] : lhs.split(':');
+    const rhsParts = rhs === '' ? [] : rhs.split(':');
+    return [
+        ...lhsParts,
+        ...Array(8 - lhsParts.length - rhsParts.length).fill(0),
+        ...rhsParts,
+    ].map((segment) => parseInt(segment, 16));
+};
+
+export const ipv4ToTuple = (ipv4) => {
+    return ipv4.split('.').map((segment) => parseInt(segment, 10));
+};
+
+export const isMulticastIpAddress = ({ tag, val: { address } }) =>
+    (tag === 'ipv4' && address[0] >= 0xe0 && address[0] <= 0xef) ||
+    (tag === 'ipv6' && address[0] === 0xff);
+
+export const isIpv4MappedAddress = ({ tag, val: { address } }) =>
+    tag === 'ipv6' &&
+    address[0] === 0 &&
+    address[1] === 0 &&
+    address[2] === 0 &&
+    address[3] === 0 &&
+    address[4] === 0 &&
+    address[5] === 0xffff;
+
+export const isWildcardIpAddress = ({ val: { address } }) =>
+    address.every((b) => b === 0);
+
+export const isBroadcastIpAddress = ({ tag, val: { address } }) =>
+    tag === 'ipv4' && address.every((b) => b === 0xff);
+
+export const isUnicastIpAddress = (a) =>
+    !isMulticastIpAddress(a) &&
+    !isBroadcastIpAddress(a) &&
+    !isWildcardIpAddress(a);
+
+/**
+ * Serialize a socket‐address to text.
+ * @param {{tag:string,val:{address:number[]}}} ipAddr
+ * @returns {string|null}
+ */
+export const serializeIpAddress = ({ tag, val: { address } }) =>
+    tag === 'ipv4' ? tupleToIpv4(address) : tupleToIpv6(address);
+
+/**
+ * Create a socket‐address object.
+ * @param {'ipv4'|'ipv6'} family
+ * @param {string} host IP text
+ * @param {number|bigint} port
+ * @returns {{tag:string,val:Object}}
+ */
+export const makeIpAddress = (family, host, port) => {
+    const address = family === 'ipv4' ? ipv4ToTuple(host) : ipv6ToTuple(host);
+
+    const base = { address, port };
+    return family === 'ipv4'
+        ? { tag: 'ipv4', val: base }
+        : { tag: 'ipv6', val: { ...base, flowInfo: 0, scopeId: 0 } };
+};
--- a/packages/preview3-shim/lib/nodejs/sockets/tcp.js
+++ b/packages/preview3-shim/lib/nodejs/sockets/tcp.js
@@ -1,0 +1,761 @@
+import { StreamReader } from '../stream.js';
+import { FutureReader } from '../future.js';
+import { ResourceWorker } from '../workers/resource-worker.js';
+import { SocketError } from './error.js';
+
+import {
+    isWildcardIpAddress,
+    isUnicastIpAddress,
+    isMulticastIpAddress,
+    isIpv4MappedAddress,
+    IP_ADDRESS_FAMILY,
+} from './address.js';
+
+const _worker = new ResourceWorker(
+    new URL('../workers/tcp-worker.js', import.meta.url)
+);
+
+const STATE = {
+    UNBOUND: 'unbound',
+    BOUND: 'bound',
+    LISTENING: 'listening',
+    CONNECTING: 'connecting',
+    CONNECTED: 'connected',
+    CLOSED: 'closed',
+};
+
+export class TcpSocket {
+    #socketId = null;
+    #family = null;
+    #state = 'unbound';
+    #options = {
+        // defaults per https://nodejs.org/docs/latest/api/net.html#socketsetkeepaliveenable-initialdelay
+        keepAliveEnabled: false,
+
+        // Node.js doesn't give us the ability to detect the OS default,
+        // therefore we hardcode the default value instead of using the OS default,
+        // since we would never be able to report it as a return value otherwise.
+        // We could make this configurable as a global JCO implementation configuration
+        // instead.
+        keepAliveIdleTime: 7200_000_000_000n,
+
+        // The following options are NOT configurable in Node.js!
+        // Any configurations set will respond correctly, but underneath retain
+        // system / Node.js defaults.
+        keepAliveInterval: 1_000_000_000n,
+        keepAliveCount: 10,
+        hopLimit: 1,
+
+        // For sendBufferSize and receiveBufferSize we can at least
+        // use the system defaults, but still we can't support setting them.
+        receiveBufferSize: undefined,
+        sendBufferSize: undefined,
+    };
+
+    /**
+     * Creates a new TCP socket
+     * WIT:
+     * ```
+     * constructor(address-family: ip-address-family)
+     * ```
+     *
+     * @static
+     * @param {IP_ADDRESS_FAMILY} addressFamily - The IP address family (IPv4 or IPv6)
+     * @param {number} socketId - The internal socket identifier
+     * @returns {TcpSocket} A new TcpSocket instance
+     * @private
+     */
+    static _create(addressFamily, socketId) {
+        const socket = new TcpSocket();
+        socket.#family = addressFamily;
+        socket.#socketId = socketId;
+        return socket;
+    }
+
+    /**
+     * Binds the socket to a local address
+     * WIT:
+     * ```
+     * bind: func(local-address: ip-socket-address) -> result<_, error-code>
+     * ```
+     *
+     * @async
+     * @param {Object} localAddress - The local address to bind to
+     * @returns {Promise<void>}
+     * @throws {SocketError} with payload.tag 'invalid-state' if the socket is not in ~STATE.UNBOUND~
+     * @throws {SocketError} with payload.tag 'invalid-argument' if ~localAddress~ is invalid
+     * @throws {SocketError} for other errors, payload.tag maps the system error
+     */
+    async bind(localAddress) {
+        if (this.#state !== STATE.UNBOUND) {
+            throw new SocketError('invalid-state');
+        }
+
+        if (!this.#isValidLocalAddress(localAddress)) {
+            throw new SocketError('invalid-argument');
+        }
+        try {
+            await _worker.run({
+                op: 'tcp-bind',
+                localAddress,
+                socketId: this.#socketId,
+            });
+
+            this.#state = STATE.BOUND;
+        } catch (error) {
+            throw SocketError.from(error);
+        }
+    }
+
+    /**
+     * Connects to a remote endpoint
+     * WIT:
+     * ```
+     * connect: func(remote-address: ip-socket-address) -> result<_, error-code>
+     * ```
+     *
+     * @async
+     * @param {Object} remoteAddress - The remote address to connect to
+     * @returns {Promise<void>}
+     * @throws {SocketError} with payload.tag 'invalid-state' if socket is CONNECTING, CONNECTED, or LISTENING
+     * @throws {SocketError} with payload.tag 'invalid-argument' if ~remoteAddress~ is invalid or port == 0
+     * @throws {SocketError} for other errors, payload.tag maps the system error
+     */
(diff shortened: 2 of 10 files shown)


packages/preview3-shim/lib/nodejs/sockets/address.js
```diff
@@ -0,0 +1,84 @@
...
+
+/**
+ * Create a socket‐address object.
+ * @param {'ipv4'|'ipv6'} family
+ * @param {string} host IP text
+ * @param {number|bigint} port
+ * @returns {{tag:string,val:Object}}
+ */
+export const makeIpAddress = (family, host, port) => {
+    const address = family === 'ipv4' ? ipv4ToTuple(host) : ipv6ToTuple(host);
```
vados-cosmonic: Question: what do you think about being a bit more careful here? a `switch` with an error? Should we throw if someone tries to pass in an invalid family/etc?

packages/preview3-shim/lib/nodejs/sockets/tcp.js
```diff
@@ -0,0 +1,764 @@
...
+     * @static
+     * @param {IP_ADDRESS_FAMILY} addressFamily - The IP address family (IPv4 or IPv6)
+     * @param {number} socketId - The internal socket identifier
+     * @returns {TcpSocket} A new TcpSocket instance
+     * @private
+     */
+    static _create(addressFamily, socketId) {
+        const socket = new TcpSocket();
+        socket.#family = addressFamily;
+        socket.#socketId = socketId;
```
vados-cosmonic, line 119: do we want to do any checking here?

packages/preview3-shim/lib/nodejs/sockets/tcp.js
```diff
@@ -0,0 +1,764 @@
...
+     * @async
+     * @param {bigint} value - The backlog size
+     * @returns {Promise<void>}
+     * @throws {SocketError} with payload.tag 'invalid-argument' if ~value~ is 0
+     * @throws {SocketError} with payload.tag 'not-supported' if socket is CONNECTING or CONNECTED
+     * @throws {SocketError} with payload.tag 'invalid-state' if socket is not UNBOUND or BOUND
+     * @throws {SocketError} for other errors, payload.tag maps the system error
+     */
+    async setListenBacklogSize(value) {
+        if (value === 0n) {
```
vados-cosmonic: it's highly unlikely someone would try to put a negative number in here, but might be nice to check for it

packages/preview3-shim/lib/nodejs/sockets/tcp.js
```diff
@@ -0,0 +1,764 @@
...
+            this.#state === STATE.CONNECTED
+        ) {
+            throw new SocketError('not-supported');
+        }
+
+        if (this.#state !== STATE.UNBOUND && this.#state !== STATE.BOUND) {
+            throw new SocketError('invalid-state');
+        }
+
+        this.#options.listenBacklogSize = Number(value);
```
vados-cosmonic: A bit curious why this is `Number` when we normally deal in `BigInt` ?

packages/preview3-shim/lib/nodejs/sockets/tcp.js
```diff
@@ -0,0 +1,764 @@
...
+     * @returns {Promise<void>}
+     * @throws {SocketError} with payload.tag 'invalid-argument' if ~value~ is less than 1
+     * @throws {SocketError} for other errors, payload.tag maps the system error
+     */
+    async setKeepAliveIdleTime(value) {
+        if (value < 1n) {
+            throw new SocketError('invalid-argument');
+        }
+
+        if (value < 1_000_000_000n) {
```
vados-cosmonic, line 511: A comment on *why* we keep this minimum would be great

packages/preview3-shim/lib/nodejs/sockets/tcp.js
```diff
@@ -0,0 +1,764 @@
...
+     * ```
+     * set-receive-buffer-size: func(value: u64) -> result<_, error-code>
+     * ```
+     *
+     * @param {bigint} value - The receive buffer size in bytes
+     * @returns {void}
+     * @throws {SocketError} with payload.tag 'invalid-argument' if ~value~ is 0
+     */
+    setReceiveBufferSize(value) {
+        if (value === 0n) {
```
vados-cosmonic: bit of paranoia but a check might be nice here to ensure this isn't negative

packages/preview3-shim/lib/nodejs/sockets/tcp.js
```diff
@@ -0,0 +1,764 @@
...
+            isUnicastIpAddress(remoteAddress) &&
+            !isWildcardIpAddress(remoteAddress) &&
+            !isMulticastIpAddress(remoteAddress) &&
+            !isIpv4MappedAddress(remoteAddress)
+        );
+    }
+}
+
+const tcpSocketCreate = TcpSocket._create;
+delete TcpSocket._create;
```
vados-cosmonic: I really wonder if we can do this differently. The obvious issues with treeshaking apply here I think, and the goal of disallowing people to create `TcpSocket`s is certainly valuable but I wonder if there's another way to do this.

Could we do this lazily when  someone actually calls tcpCreateSocket?

packages/preview3-shim/lib/nodejs/sockets/udp.js
```diff
@@ -0,0 +1,470 @@
...
+     * @static
+     * @param {IP_ADDRESS_FAMILY} addressFamily - The IP address family (IPv4 or IPv6)
+     * @param {string} socketId - The internal socket identifier
+     * @returns {UdpSocket} A new UdpSocket instance
+     * @private
+     */
+    static _create(addressFamily, socketId) {
+        const sock = new UdpSocket();
+        sock.#family = addressFamily;
+        sock.#socketId = socketId;
```
vados-cosmonic, line 98: A check on address family would be great here... not sure you can do anything about socketId? Do we expect this to be a certain form?

packages/preview3-shim/lib/nodejs/sockets/udp.js
```diff
@@ -0,0 +1,470 @@
...
+     *
+     * @async
+     * @param {Object} localAddress - The local address to bind to
+     * @returns {Promise<void>}
+     * @throws {SocketError} with payload.tag 'invalid-state' if the socket is not in ~STATE.UNBOUND~
+     * @throws {SocketError} with payload.tag 'invalid-argument' if ~localAddress~ is invalid
+     * @throws {SocketError} for other errors, payload.tag maps the system error
+     *
+     */
+    async bind(localAddress) {
```
vados-cosmonic: what's interesting is that this isn't actually `async` hinted in the WIT -- what do you think about being inconsistent with upstream here?

packages/preview3-shim/lib/nodejs/sockets/tcp.js
```diff
@@ -0,0 +1,764 @@
...
+     * WIT:
+     * ```
+     * send-buffer-size: func() -> result<u64, error-code>
+     * ```
+     *
+     * @async
+     * @returns {Promise<bigint>} The send buffer size in bytes
+     * @throws {SocketError} for other errors, payload.tag maps the system error
+     */
+    async sendBufferSize() {
```
vados-cosmonic: this actually isn't `async`-hinted in the upstream WIT -- what do you think about making this sync?

I'm not 100% sure that we can async import from the preview2 stuff, but we *should* be able to do that (and add the functionality later if it's missing)

packages/preview3-shim/lib/nodejs/sockets/udp.js
```diff
@@ -0,0 +1,470 @@
...
+     * WIT:
+     * ```
+     * set-send-buffer-size: func(value: u64) -> result<_, error-code>
+     * ```
+     *
+     * @param {bigint} value - The buffer size in bytes (must be >0)
+     * @throws {SocketError} with payload.tag 'invalid-argument' if value == 0
+     */
+    setSendBufferSize(value) {
+        if (value === 0n) {
```
vados-cosmonic, line 455: sanity check here would be nice

packages/preview3-shim/lib/nodejs/workers/tcp-worker.js
```diff
@@ -0,0 +1,254 @@
...
+    .op('tcp-get-remote-address', handleGetRemoteAddress)
+    .op('tcp-set-listen-backlog-size', handleTcpSetBacklogSize)
+    .op('tcp-set-keep-alive', handleTcpSetKeepAlive)
+    .op('tcp-recv-buffer-size', handleRecvBufferSize)
+    .op('tcp-send-buffer-size', handleSendBufferSize)
+    .op('tcp-dispose', handleTcpDispose);
+
+// Create a new TCP socket
+function handleTcpCreate({ family }) {
+    const socketId = randomUUID();
```
vados-cosmonic: what do you think about using an incrementing BigInt here...? If we know people must come through here, I think we might be able to get away with a smaller data type here. I like the idea of ensuring they won't collide, but I wonder if we need that here

packages/preview3-shim/lib/nodejs/workers/tcp-worker.js
```diff
@@ -0,0 +1,254 @@
...
+        }
+
+        if (code !== 0) {
+            reject(SocketError.from(-code));
+        } else {
+            resolve();
+        }
+    });
+
+    return { success: true };
```
vados-cosmonic: we use `ok: true` in some other places, but I wonder if you even need this here? Do we `try`/`catch` this otherwise?

packages/preview3-shim/lib/nodejs/workers/tcp-worker.js
```diff
@@ -0,0 +1,254 @@
...
+    const onError = once(server, 'error').then(([err]) => {
+        throw err;
+    });
+
+    server.listen(handle, backlog);
+
+    await Promise.race([onListening, onError]);
+
+    server.on('connection', (conn) => {
+        const id = randomUUID();
```
vados-cosmonic: same note about the bigint here -- I wonder if we need the weight of a random UUID here

packages/preview3-shim/lib/nodejs/workers/tcp-worker.js
```diff
@@ -0,0 +1,254 @@
...
+    if (socket.handle) {
+        socket.handle.close();
+    }
+
+    sockets.delete(socketId);
+}
+
+let _recvBufferSize, _sendBufferSize;
+async function getDefaultBufferSizes() {
+    var s = new Socket({ type: 'udp4' });
```
vados-cosmonic: should this be `udp`? are we sharing the value for udp here?

Is that option actually used..? I can't find the `type` member for `Socket`'s constructor or NodeJS code

packages/preview3-shim/test/tcp.test.js
```diff
@@ -0,0 +1,219 @@
+import { describe, test, expect, beforeAll, afterAll } from 'vitest';
+import { StreamReader, stream } from '@bytecodealliance/preview3-shim/stream';
+
+import net from 'net';
```
vados-cosmonic:
```suggestion
import net from 'node:net';
```

packages/preview3-shim/test/tcp.test.js
```diff
@@ -0,0 +1,219 @@
...
+
+        expect(localAddr).toBeDefined();
+        expect(localAddr.tag).toBe(IP_ADDRESS_FAMILY.IPV4);
+        expect(localAddr.val.address).toStrictEqual([127, 0, 0, 1]);
+        expect(typeof localAddr.val.port).toBe('number');
+        expect(localAddr.val.port).toBeGreaterThan(0);
+    });
+});
+
+describe('TCP Socket Listen', () => {
```
vados-cosmonic, line 80: So glad to have these tests here! :+1:

packages/preview3-shim/test/tcp.test.js
```diff
@@ -0,0 +1,219 @@
...
+        expect(buf1.toString()).toBe(msg1);
+
+        // Dispose all resources
+        conn[Symbol.dispose]();
+        client[Symbol.dispose]();
+        listener[Symbol.dispose]();
+    });
+});
+
+describe('Intagration: TCP Socket Send', () => {
```
vados-cosmonic:
```suggestion
describe('Integration: TCP Socket Send', () => {
```

packages/preview3-shim/test/udp.test.js
```diff
@@ -0,0 +1,158 @@
+import { describe, test, expect, beforeAll, afterAll } from 'vitest';
+import dgram from 'node:dgram';
+import {
+    UdpSocket,
+    udpCreateSocket,
+    makeIpAddress,
+    IP_ADDRESS_FAMILY,
+} from '@bytecodealliance/preview3-shim/sockets';
```
vados-cosmonic:
```suggestion
import dgram from 'node:dgram';

import { describe, test, expect, beforeAll, afterAll } from 'vitest';
import {
    UdpSocket,
    udpCreateSocket,
    makeIpAddress,
    IP_ADDRESS_FAMILY,
} from '@bytecodealliance/preview3-shim/sockets';
```

packages/preview3-shim/test/udp.test.js
```diff
@@ -0,0 +1,158 @@
...
+
+    test('throws on invalid address family', async () => {
+        await expect(
+            udpCreateSocket.createUdpSocket('invalid')
+        ).rejects.toSatisfy((err) => err.payload.tag === 'invalid-argument');
+    });
+});
+
+describe('UDP Socket Bind', () => {
+    //
```
vados-cosmonic:
```suggestion
```

packages/preview3-shim/lib/nodejs/sockets/tcp.js
andreiltd: I was also using a bit different pattern for this, for example for sealing `Request/Response` constructor:

```js
   const REQUEST_PRIV_SYM = Symbol('RequestPrivSym');
   
    constructor(token) {
        if (token !== REQUEST_PRIV_SYM) {
            throw new Error('Use Request.new(...) to create a Request');
        }
    }
```

packages/preview3-shim/lib/nodejs/sockets/udp.js
andreiltd: Good point, we can change this easily by calling `runSync` on the worker. Should we maybe start with replicating 1:1 WIT hints and we can tweak later? It should be very easy to go either way if we decide to change this later.

andreiltd: There are also other functions here that surprisingly aren't `async` like `send`, `connect`, `listen`.

packages/preview3-shim/lib/nodejs/sockets/tcp.js
vados-cosmonic: I think that certainly works -- one more tweak to make it not a symbol creation at the module level then it's just fine IMO! :ship:

packages/preview3-shim/lib/nodejs/sockets/udp.js
vados-cosmonic: Yeah -- it may be a bit surprising that many of them aren't async hinted, but technically people can lower anything they want as async and what not, and it's up to jco to support that.

I agree with replicating the WIT 1:1 for now and tweaking later as necessary!

packages/preview3-shim/lib/nodejs/sockets/address.js
andreiltd: Sure!

packages/preview3-shim/lib/nodejs/sockets/tcp.js
andreiltd: Yup, it makes sense to keep it here as `BigInt`.

packages/preview3-shim/lib/nodejs/sockets/tcp.js
andreiltd: OK, I rework the sealing a bit. Let me know if it looks OK.

packages/preview3-shim/lib/nodejs/sockets/udp.js
andreiltd: The socketId is a `BigInt`. The family is already checked in `createUdpSocket` function.

packages/preview3-shim/lib/nodejs/workers/tcp-worker.js
andreiltd: Done!

packages/preview3-shim/lib/nodejs/workers/tcp-worker.js
andreiltd: Removed

packages/preview3-shim/lib/nodejs/sockets/udp.js
vados-cosmonic: Ah thanks -- I overlooked that!

packages/preview3-shim/lib/nodejs/sockets/tcp.js
vados-cosmonic: Yeah that new version looks good! Thanks

Hopefully in the future we can change all of them to use this method rather than the old one, but it's not a super priority just yet!

packages/preview3-shim/lib/nodejs/sockets/udp.js
andreiltd: Done!

PR #736, opened 2025-06-05
feat(bindgen): context.{get|set}, initial task.return impl

vados-cosmonic:
This commit forms the initial implementation of context.get, context.set and task.return for P3 async support.

Also introduced is a mechainsm (via ENV var JCO_DEBUG) of controlling debug output.


vados-cosmonic:
Chunking this up in the interest of having a slightly smaller commits land, tests are currently only partially executing the new functionality, because async P3 exports are not fully implemented. That said, sync <-> sync call of a the new `context.get`/`context.set` work, and `task.return` is also working but still incomplete.

PR #738, opened 2025-06-06
fix(p3-shim): avoid side effects when loading filesystem API

andreiltd:
Lazily create a worker for the descriptor.


vados-cosmonic (approved):
LGTM 🚀

PR #746, opened 2025-06-09
fix(p3-shim): remove UUID from udp worker

andreiltd:
I somehow overlooked the switch from UUID to `bigint` indices in the udp worker, but let's be consistent.


vados-cosmonic (approved):
LGTM 🚀 

Thanks for finding this -- I overlooked it to on the last review!

PR #748, opened 2025-06-10
feat(p3-shim): close socket's server explicitly on symbol dispose

andreiltd:
Small patch that ensure that we close the socket's server explicitly on symbol dispose. This also ensures that accept connection stream will be closed.


vados-cosmonic (approved):
LGTM :rocket:

PR #755, opened 2025-06-12
feat(p3-shim): generate types for preview3-shim package

andreiltd:
The patch introduces generated preview 3 types for the `preview3-shim` package.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -183,10 +183,6 @@ jobs:
           rustup target add wasm32-unknown-unknown
           rustup target add wasm32-wasip1
 
-      - uses: taiki-e/cache-cargo-install-action@4d586f211d9b0bca9e7b59e57e2a0febf36c0929 # v2.1.1
-        with:
-          tool: wasm-tools
-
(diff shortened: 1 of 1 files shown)


crates/js-component-bindgen/src/ts_bindgen.rs
```diff
@@ -136,7 +136,7 @@ pub fn ts_bindgen(
                     {
                         let import_specifier = resolve.id_of(*id).unwrap();
                         let (_, _, iface) = parse_world_key(&import_specifier).unwrap();
-                        debug!("skipping imported interface [{}] feature gate due to feature gate visibility", iface.to_string());
+                        debug!("skipping imported interface [{}] feature gate due to feature gate visibility", iface);
```
vados-cosmonic:
```suggestion
                        debug!("skipping imported interface [{iface}] feature gate due to feature gate visibility");
```

vados-cosmonic (approved):
Thanks for adding support for p3 and getting the wits in place!

Tiny nit but this should be good to go otherwise

PR #764, opened 2025-06-18
feat(ci): run clippy on codebase

andreiltd:
What is says on the tin, let's make clippy happy.


andreiltd:
> At present this PR is titled as if it's adding clippy but it's doing much more.
Do you mean fixing the clippy issues in the code or adding SARIF?

vados-cosmonic:
Adding SARIF and turning on GH Code scanning -- having clippy check and fail is certainly useful for the same reason we check the JS but the other things are a little less clear.

.github/workflows/main.yml
```diff
@@ -347,9 +347,26 @@ jobs:
...
         run: rm -r crates/jco/tests/mod.rs
+      - uses: Swatinem/rust-cache@v2
       - name: Format source code
         run: cargo fmt -- --check
 
+  clippy:
+    name: clippy
+    runs-on: ubuntu-latest
+    permissions:
+      security-events: write # to upload sarif results
```
vados-cosmonic:
```suggestion
```

.github/workflows/main.yml
```diff
@@ -347,9 +347,26 @@ jobs:
...
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v4
```
vados-cosmonic: Would you mind including the SHA hashes + versions like the other actions?

Issue #762, opened 2025-06-17
Adding a shim for wasi:logging


thomas9911:
Is it possible to add wasi:logging to the preview2-shim package? Or is this not possible because the wasi:logging is still in draft?

wasi logging proposal: https://github.com/WebAssembly/wasi-logging

thomas9911:
I currently have the issue that I use jco transpile with a component that uses the wasi:logging/logging interface. The other wasi interfaces get a nice preview2-shim so they work in javascript/typescript, but wasi:logging doesn't.

example:

```js
import { environment, exit as exit$1, stderr, stdin, stdout } from '@bytecodealliance/preview2-shim/cli';
import { preopens, types } from '@bytecodealliance/preview2-shim/filesystem';
import { error, streams } from '@bytecodealliance/preview2-shim/io';
import { log } from 'wasi:logging/logging';
```

vados-cosmonic:
Yeah no problem! I'm glad you brought this up -- it's something we've discussed at meetings but hasn't been discussed many places yet :)

thomas9911:
indeed I could get away with the --map:

```
jco transpile component.wasm --out-dir transpiled --map wasi:logging/logging=./log.js
```

with log.js:

```js
export const log = (level, ctx, message) => {
  switch (level) {
    case 'trace':
    case 'debug':
      console.debug(message);
      break;
    case 'info':
      console.info(message);
      break;
    case 'warn':
      console.warn(message);
      break;
    case 'error':
    case 'critical':
      console.error(message);
      break;
    default:
      console.log(message)
  }
}
```

vados-cosmonic:
Glad you were able to figure this out! You know what, this might make a great example, if you want to submit it -- We actually don't have an example of a host implementation of an interface, surprisingly!

thomas9911:
created an example: https://github.com/bytecodealliance/jco/pull/765

PR #765, opened 2025-06-18
feat(examples): add host function example

thomas9911:
This example shows how to implement a host function. It uses the wasi:logging interface, because that is quite simple to implement.

Suggested from issue: https://github.com/bytecodealliance/jco/issues/762


examples/components/host-logging/component.js
```diff
@@ -0,0 +1,23 @@
+/**
+ * This module is the JS implementation of the `reverser` WIT world
```
thomas9911: reverser

examples/components/host-logging/component.js
```diff
@@ -0,0 +1,23 @@
...
+export const logCharacters = {
+  /**
+   * This Javascript will be interpreted by `jco` and turned into a
+   * WebAssembly binary with a single export (this `call` function).
+   */
+  call(s) {
+    for (let char of s.split("")) {
+      log('info', "stdout", char)
+    }
+  }
```
thomas9911, line 23: run formatter, single+double quotes

vados-cosmonic (approved):
Thanks for adding this @thomas9911 this is excellent -- amazing we haven't had an example of host impls for this long.

vados-cosmonic:
One thing that would be great -- could we get a test that we can plug in to CI? You could wire it up geenerally like the other tests, and include the transpile step via `package.json` (I think we use the `all` script quite often?)

examples/components/host-logging/README.md
```diff
@@ -0,0 +1,90 @@
...
+}
+
+world component {
+    import wasi:logging/logging@0.1.0-draft;
+
+    export log-characters;
+}
+```
+
+A component that implements the `component` world exports a single interface called `log-characters`, which consists of one function called `call`, which takes a string and outputs earch character on a new line using the `log-host.js` host function.
```
thomas9911: typo earch

Issue #717, opened 2025-05-29
Moving Jco to Typescript


vados-cosmonic:
It appears to me that we might greatly benefit from moving `jco` itself to Typescript so that we can benefit from compile time catching of bugs/inconsistencies in expected and received objects, etc.

There are at least a few benefits:

- Build type catching of bugs
- Reference for various APIs
- Better editor assitance
- Ability to pull in libraries like `zod` which can use TS schemas to generate checking fns

elmerbulthuis:
I've been fantasizing a bit about this migration to TypeScript and came up with the following...

A complete migration to TypeScript would involve a few steps. Every step should bring us a bit closer to the end goal and every step should be releasable.

Step 1 would be to build all the packages. Every project get's its own `tsconfig.json` and the `.d.ts` files will be built from the `.js` files. Possibly a configuration for vscode could be added so the vscode people (I'm one of them) can run the build directly from the IDE.

Now that we have a nice setup to build npm projects we are ready for step 2! In this step we reorganize the npm packages in a way that we don't store the TypeScript build artifacts. We will end up with packages that have only `.js` files in them, the `.d.ts` files are then generated from these files. These generated files are placed in folders that are ignored by git.

Then, in step 3 we rename all `.js` files to `.ts` and automagically add type information. This is manually checked and possibly some types are added and changed. The code itself should not or hardly be changed in this step. This is kind of what I did in #822 .

The last step is to refactor code to be more "TypeScript-ish". For instance this thing:

```ts
export async function readFile(
    file: string | Buffer | URL | FileHandle,
    encoding?: BufferEncoding
): Promise<Buffer | string> {
    // bla bla...
}
```

Could be refactored to return either a `Buffer` or a `string`, depending on the value of the encoding parameter. There are quite a few places where code could be refactored to better make use of TypeScripts awesome features.

This final step is more of an optimization. We could just do this on the side or if it makes sense. Also, when new code is written then the old code will gradually disappear and then this final step is somewhat for free.

So it's 3 steps to TypeScript and then optimizing. Of course every step has some side effects that I have not described here. Like every step probably results in a slight change in the CI setup and there may be many more side effects. They should be identified when working on the phase, at this point they don't seem like a problem.

But of course first we need to want to do this...

PR #822, opened 2025-07-08
Experimental TypeScript code

elmerbulthuis:
I have been translating a small portion of the code to TypeScript to get a feel of the impact this would have.

First I renamed all `.js` files I wanted to transform to `.ts`, then I let some agent (codex) fill in all the missing types for the functions. so far everything was easy and straightforward. Then the work became a bit more difficult and interesting, I walked trought every file to manually transform the rest. The code is transformed with the least amount of effort. Some of the patterns used could (should) be optimized for TypeScript but nothing too difficult.

I feel that this little exercise may already expose some bugs or at least weak spots. There are things like using null when undefined is expected that might cause problems in rare cases,  there are are also strings being used when UInt8Arrays are expected that would likely cause more problems. I think TypeScript can provide some real benefit by being more strict about this.

I might introduced some new bugs! Sorry for that, this was just to get a feel of the thing.

Another thing that I notice is that the current setup makes for 4 files (yes, 4!) per module. So we have one .ts file, this is where we should be working in. Then we have a compiles `.js` and `.d.ts` file, and then we have a map file, `.js.map`. We want all of these files, but should they really be in the src folder? Should they really be in the repo? I'd propose a slightly more sophisticated build system to make this more ergonomic. This would then be a task that we do before moving to TypeScript.

So like I mentioned before in #717, I am in favor of moving the codebase to TypeScript, and after my little experiment I am even more into doing that! It will not be without any effort, but the sooner we move, the less code we need to move and the earlier we can reap the benefits of this move.

\o/

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -20,7 +20,7 @@ jobs:
         with:
           mdbook-version: '0.4.43'
 
-      - uses: taiki-e/install-action@f3a27926ea13d7be3ee2f4cbb925883cf9442b56 # v2.56.7
+      - uses: taiki-e/install-action@2383334cf567d78771fc7d89b6b3802ef1412cf6 # v2.56.8
         with:
           tool: >-
             mdbook-alerts@0.7.0,
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -409,7 +409,7 @@ jobs:
 
       - name: Install required rust crates
         if: ${{ matrix.project.requires-crates != '[]' }}
-        uses: taiki-e/install-action@f3a27926ea13d7be3ee2f4cbb925883cf9442b56 # v2
+        uses: taiki-e/install-action@2383334cf567d78771fc7d89b6b3802ef1412cf6 # v2
         with:
           tool: ${{ join(matrix.project.requires-crates, ',') }}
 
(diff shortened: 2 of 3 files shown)


t3hmrman:
Hey Elmer thanks for starting to explore and working on this!

I'm going to discuss it at the bext [SIG Guest Languages meeting for Javascript](https://github.com/bytecodealliance/meetings/tree/main/SIG-Guest-Languages/JavaScript) (that's happening soon) -- if you'd like to join in feel free to attend or just catch the recording later!

vados-cosmonic:
Hey Elmer thanks for starting to explore and working on this!

I'm going to discuss it at the bext [SIG Guest Languages meeting for Javascript](https://github.com/bytecodealliance/meetings/tree/main/SIG-Guest-Languages/JavaScript) (that's happening soon) -- if you'd like to join in feel free to attend or just catch the recording later!

elmerbulthuis:
O great! I'd love to join!

If I am correct the thing is not scheduled yet right? I'll keep an eye on that link then.

Cu then! :-)

vados-cosmonic:
Ah hopefully I'm not too late here but to be clear the next meeting *will* happen -- in roughly ~2 hours!

elmerbulthuis:
Oo dear! I am at the gym with my son in 2 hours! I guess I'll check out the recording and maybe next time I can join.

Cu soon!

PR #825, opened 2025-07-10
Experimental building of npm packages

elmerbulthuis:
see #717 

This demonstrates what it would look like if we would introduce a build step. This PR also removes .d.ts files from the repo and generated them in a seperate folder (dist) during build.

- [ ] Also build the test folder in a clean way
- [ ] Discuss use of Just (i like just) instead of xtask
- [ ] setup proper formatting
- [ ] ...


elmerbulthuis:
One thing I like to do is to have a script in the package.json that performs local (as in project-local) tasks. In this case compiling code. Then there is a workspace scoped task runner, currently xtask that calls those project-local tasks in the right order.

I would prefer Just to be the workspace task runner, there is already a Justfile in there! I like Just over xtask as i feel Justfiles are quite readable so it becomes more clear what exactly the thing is doing.

But I am also fine with xtask ofcourse.

vados-cosmonic:
Hey @elmerbulthuis thanks for getting this PR out there! Would you mind breaking the work for only one project (jco-transpile) off into another PR? You can leave this one open and "just" cherry pick changes over.

With regards to the `just`/`xtask`/`npm` task running situation, I'd like to lay out the rationale for the current state of things:

- `just` is a general, top/project-level task runner
- `xtask` is a useful Rust code related task runner (it can do nice things like use rust libs and even the projects in question)
- `npm` scripts are task runners native to JS code

I think all three of these have their place -- IMO the things they do arguably should *all* be accessible from `just`, but I'm not sure it makes sense to collapse all into any one of them.

To your concern specifically:

> I would prefer Just to be the workspace task runner, there is already a Justfile in there! I like Just over xtask as i feel Justfiles are quite readable so it becomes more clear what exactly the thing is doing.

Just is at a higher level of abstraction than `xtask`! I am onboard with all important `xtask`s being expressed via `just` targets!

> I build only the src folder at the moment, and I put the output in the dist folder that I don't check in. Quite happy with how clean the src folders are now. I do need to find a neat solution for the test folder, but that is for another day!

This sounds great to me, `src` and `dist` are the defaults IMO in the JS ecosystem and the `src` folder getting cleaner is a huge benefit IMO!

PR #861, opened 2025-07-18
chore(build): compile jco and remove build artifacts

elmerbulthuis:
This PR removes the build artifacts (`.d.ts`) from the repo. These are now generated by `npm run build` and put in a dist folder. This unclutters the workspace. Also this is a first and gentle step towards migrating to a TypeScript codebase.

see https://github.com/bytecodealliance/jco/issues/717

PR #863, opened 2025-07-18
chore(build): build setup for jco-transpile

elmerbulthuis:
based on https://github.com/bytecodealliance/jco/pull/862

Removes build artifacts from the repo for the jco-transpile project. The artifacts are generated by the `tsc` script and put in the dist folder, that is not in the repo.

This removes some clutter from the repo and is a preparation for a TypeScript codebase, see https://github.com/bytecodealliance/jco/issues/717.

This is similar to https://github.com/bytecodealliance/jco/pull/861, but for the jco-transpile project.


crates/xtask/src/build/workspace.rs
```diff
@@ -15,6 +15,6 @@ pub(crate) fn run(release: bool) -> anyhow::Result<()> {
     }
 
     // Build Jco TS code
-    cmd!(sh, "npx -w @bytecodealliance/jco tsc -p tsconfig.json").read()?;
+    cmd!(sh, "npm run tsc").read()?;
```
vados-cosmonic: I think the intent here was *not* to run `tsc` on every project, but to specifically run `tsc` on `jco`, so I think we can still limit the workspace here.

package.json
```diff
@@ -23,6 +23,7 @@
     "build:types:preview3-shim": "cargo xtask generate wasi-types preview3",
     "fmt": "npm run fmt --ws --if-present",
     "lint": "npm run lint --ws --if-present",
-    "lint:fix": "npm run lint:fix --ws --if-present"
+    "lint:fix": "npm run lint:fix --ws --if-present",
+    "tsc": "npm run --workspaces --if-present tsc"
```
vados-cosmonic: What do you think about naming this `build:ts` ?

packages/jco-transpile/package.json
```diff
@@ -47,11 +46,16 @@
...
+    "prepack": "npm run setup:vendor ; npm run tsc",
+    "tsc": "tsc --build"
   },
   "dependencies": {
     "@bytecodealliance/preview2-shim": "^0.17.2",
     "chalk-template": "^1",
     "terser": "^5"
+  },
+  "devDependencies": {
+    "@tsconfig/node22": "^22.0.2",
```
vados-cosmonic: What do you think about pull the settings in this file into the `tsconfig.json` directly? I think it might be worth avoiding an entirely new dependency to take in some settings, and when people are curious what very important settings like `module` or `esModuleInterop` are set to, they can see immediately.

packages/jco/src/api.d.ts
```diff
@@ -2,39 +2,39 @@
  * @param {Parameters<import('../obj/wasm-tools.js').print>[0]} binary
  * @return {Promise<ReturnType<import('../obj/wasm-tools.js').print>>}
  */
-export function print(binary: Parameters<import("../obj/wasm-tools.js").print>[0]): Promise<ReturnType<import("../obj/wasm-tools.js").print>>;
+export function print(binary: Parameters<any>[0]): Promise<ReturnType<any>>;
```
vados-cosmonic: Did this stop working? It would be great if we could keep this typed based on the import.

[EDIT] - Actually also let's back the changes for Jco out of this PR

packages/jco/package.json
```diff
@@ -63,7 +63,8 @@
     "lint:fix": "npm run lint -- --fix",
     "test": "vitest run -c test/vitest.ts",
     "test:lts": "vitest run -c test/vitest.lts.ts",
-    "prepack": "cargo xtask build release"
+    "prepack": "cargo xtask build release",
+    "tsc": "tsc --build"
```
vados-cosmonic: would you mind backing out these changes for `jco` ?

packages/jco/src/common.d.ts
```diff
@@ -12,6 +12,8 @@ export function getTmpDir(): Promise<string>;
 export function spawnIOTmp(cmd: any, input: any, args: any): Promise<Buffer<ArrayBufferLike>>;
 export function writeFiles(files: any, summaryTitle: any): Promise<void>;
 export const isWindows: boolean;
+export const ASYNC_WASI_IMPORTS: string[];
+export const ASYNC_WASI_EXPORTS: string[];
```
vados-cosmonic: Thanks for fixing this! :bow:

packages/jco-transpile/package.json
```diff
@@ -47,11 +46,16 @@
     "lint:fix": "npm run lint -- --fix",
     "test": "vitest run -c test/vitest.ts",
     "setup:vendor": "npm run -w @bytecodealliance/jco build:release && cp ../jco/obj/*.core*.wasm vendor && cp ../jco/obj/*.js vendor && cp ../jco/obj/*.ts vendor && cp -r ../jco/obj/interfaces vendor",
-    "prepack": "npm run setup:vendor"
+    "prepack": "npm run setup:vendor ; npm run tsc",
+    "tsc": "tsc --build"
```
vados-cosmonic: Same note here with the possible rename -- `build:ts` would be a really good name IMO. 

Also, this is a bit ambitious, but what do you think about adding a very "end to end" test so that we can ensure that types and source maps are included in the bundle? 

This is something we haven't had until now, but it would greatly increase confidence in releases. I added something similar (but not quite the same) for `jco` [to ensure we test release packages](https://github.com/bytecodealliance/jco/blob/main/.github/workflows/release.yml#L296) before they go out. All we would need to do would be to maybe run pack and ensure that the expected files are there in the published artifact (we don't need to go as far as checking the contents IMO. 

Definitely open to other ways of testing this/ensuring the right files make it in.

crates/xtask/src/build/workspace.rs
elmerbulthuis: The intention is to eventually do run tsc on every project. So that is why I am doing exactly that here.

Currently the only project with a tsc task is jco! So this will effectively run the script only on jco.

package.json
elmerbulthuis: I;m perfectly fine with that, however I do feel that building could be more than the transpiling that tsc does, also, tsc does currently work on ts, but on js.

I names the thing tsc, because that is exactly what it does, transpiling and ts / js is implicit because this is what the tool does.

But, tomato, tomato! I'm happy to change it :-)

packages/jco-transpile/package.json
elmerbulthuis: I'm not too worried about an extra dependency, but I get the point. I do like the simplicity of this approach as it will prevent any form of discussion of what is "better" and any deviation from this "standard" will be visible immediately.

But I don't have a very strong opinion on this.

shall I change it?

packages/jco/src/common.d.ts
elmerbulthuis: :-)

packages/jco/package.json
elmerbulthuis: Revert this?

packages/jco/src/api.d.ts
elmerbulthuis: Ye this is what tsc generated!

TBH I didn't look into this, as this will be nice TS code in the future and the code is generated and valid.

packages/jco-transpile/package.json
elmerbulthuis: I see no problem in ensuring the right files make it in. We need a little plan.

I guess it would be something in the lines of
- build tsc project
- run test project with the build project as a dependency.

So I that would mean that this e2e project is another package.

This test package could then work with the packages code, or with the package as a tgz. I would just have the test package work with the code, not the tgz, this makes the thing a bit more easy and flexible. But either is possible.

So the thing we need to do tho make this happen IMHO is to split off the test code of the 2 projects in 2 separate projects.

I feel this is not ambitious at all. This wold solve another problem as well, the compilation of the test folder. This could be done in various ways, but moving test code to another projcet would be most elegant i think.

packages/jco/src/common.d.ts
elmerbulthuis: You may thank tsc for that :-)

package.json
vados-cosmonic: I'd argue that all the *other* steps are named after what the thing *does* -- (`lint`, `fmt`, etc), there's only one that's named after a tool/command line program rather than what it does.

If we were being super literal here then "running tsc" *might* turn into something like `run:tsc`. But I think sticking close to the semantic meaning of what it's doing ("building typescript files") is reasonable here.

packages/jco-transpile/package.json
vados-cosmonic: Hmnn, I'm not sure we need to do all that! Couldn't we simply run `npm pack`, with the output going to a tempdir, unpack, and check that the files we expect are present?

What do you think about that approach? there's no need to create another package/project/folder, this test can live alongside [other tests for jco transpile](https://github.com/bytecodealliance/jco/tree/main/packages/jco-transpile/test).

packages/jco-transpile/package.json
vados-cosmonic: I'd rather avoid another dependency/reduction in clarity of the `tsconfig.json` file -- I don't think there's enough gain here.

> any deviation from this "standard" will be visible immediately.

I'm not sure it even does this, because in our `tsconfig.json` we have no idea when someone sets a value that disagrees with the upstream/extended one, because it's just overwritten, right?

packages/jco/src/api.d.ts
vados-cosmonic: So unfortunately this would be a regression for the generated code if we were to release it right? people who otherwise got good types would then be stuck with `any`?

crates/xtask/src/build/workspace.rs
vados-cosmonic: Ah thanks for clarifying the intent here! This change is fine then -- building the ts for *all* projects isn't necessary but I assume it wouldn't hurt too much either.

Let's change the comment above to represent what the line is now actually doing!

elmerbulthuis: Whahaha I thought that was what the comment said! "Build Jco TS code" anyway. I think this proves that a bit of clarification is welcome. I'll change it!

package.json
elmerbulthuis: i'll rename it to build:ts!

crates/xtask/src/build/workspace.rs
vados-cosmonic: Ahhh this is a good point, yeah, that comment is actually very ambiguous -- "jco" does sound like the whole project. Now I'm starting to think that what you had is *more* right than what was there before!

packages/jco/package.json
vados-cosmonic: Ah yes -- so if we could get away with only having jco-transpile changes in here that would be awesome. Obviously we don't want to break any existing code that was there, but this change wasn't necessary for jco-transpile right?

package.json
vados-cosmonic: Appreciate the flexibility here! 🙇

packages/jco-transpile/package.json
elmerbulthuis: True, I'll remove the dep and put the settings in the tsconfig.file!

packages/jco-transpile/package.json
elmerbulthuis: well of course that could be done, but is there a reason to believe that files would not end up in the package? I do trust npm doing this correctly.

I'm assuming a thing we test for is that the files are correctly specified in the package.json, that would be worth testing maybe, but then again we are going to need to specify the files in the tests as well, so I see no real gain here. But it could be done yes, it's a simple test.

packages/jco/package.json
elmerbulthuis: o yes you are right! this is clearly a mistake!

packages/jco-transpile/package.json
vados-cosmonic: > well of course that could be done, but is there a reason to believe that files would not end up in the package? I do trust npm doing this correctly.

So this was paranoia on my part, since during a recent release of componentize-js I found some files that I expected to be present were missing (and I've had some issues with including *too many* files in jco-transpile releases in the past).

More concretely though, I think in `packages/jco-transpile/package.json` the `dist` folder needs to be in `files`, otherwise the types won't resolve properly. That said, rather than trying to just fix that as a one-off (assuming that's right!), I wanted to add something a bit more automated.

packages/jco/src/api.d.ts
elmerbulthuis: That is very true, I'll fix this so the proper types are used.

elmerbulthuis: ok but this change is in the jco project, changes in the pr should be scoped to the jco-transpiled project. So this needs to be reverted... Doing that!

packages/jco-transpile/package.json
elmerbulthuis: Yes that is correct! Added dist to the files. Me forgetting this proves that a test of the generated package is valuable. But then if we would test for the actual files, then this is another place where we can make mistakes. And I feel that forgetting to update the files array is just one thing that could go wrong. There are many more things that could break in the final release.

That is why I would still advocate for using the released package in some kind of e2e test. This would then implicitly test for missing files and other things that could be wrong. Then we could also add some explicit tests to get a really thorough test of the thing. We could test for use of `any` for example.

Maybe this deserves a bit of extra attention, maybe in a separate issue?

vados-cosmonic: > Yes that is correct! Added dist to the files. Me forgetting this proves that a test of the generated package is valuable. But then if we would test for the actual files, then this is another place where we can make mistakes. And I feel that forgetting to update the files array is just one thing that could go wrong. There are many more things that could break in the final release.

Absolutely! I may have forgotten myself.

> That is why I would still advocate for using the released package in some kind of e2e test. This would then implicitly test for missing files and other things that could be wrong. Then we could also add some explicit tests to get a really thorough test of the thing. We could test for use of any for example.
>
> Maybe this deserves a bit of extra attention, maybe in a separate issue?

Yup, a fixture project would be great to have, and that would go great with an E2E test of the packaged artifact.

As far as `any` goes, we can actually disallow that in configuration and depend on that. 

That said, I think this has been enough discussion to warrant a new issue! Feel free to make one and close out this comment thread! Otherwise I'll do it on Monday I think.

elmerbulthuis: Will do that :rocket:

vados-cosmonic:
Closing in favor of #1097 , thanks for working on this @elmerbulthuis !

PR #876, opened 2025-07-21
chore(examples): remove unnecessary wasi:cli/environment import

vados-cosmonic:


vados-cosmonic:
It looks like although we've updated componentize-js we can't remove this just yet -- `0.18.4-rc.0` also fails here... Going to close this for now and revisit when upstream supports it

PR #899, opened 2025-08-06
chore(examples): remove WASI env

vados-cosmonic:
Remove the `wasi:cli` env imports now that we have a newer version of StarlingMonkey


andreiltd (approved):
Nice! 🚀

Issue #891, opened 2025-07-31
Add `implements Disposable` in TypeScript definition for Resources


kaivol:
JavaScript proxy objects for WIT `resource`s seem to implement the `Disposable` interface, however, this is not specified in the generated TypeScript definitions. 

I would like to suggest adding this to the definitions.

vados-cosmonic:
Hey @kaivol that's a great idea -- would you be comfortable trying to add this? If not I think we could definitely get to it soonish!

kaivol:
I'm not really familiar with the project, so I probably won't be able to do that anytime soon. 
If I find the time to look into it, I'll let you know.

iamrajiv:
hi @vados-cosmonic can i work on this if its ok?

iamrajiv:
hi @vados-cosmonic so what i understand is i just need to modify the typescript definition generator in the rust code to add implements disposable to resource classes right?

PR #901, opened 2025-08-07
feat: Generate `Disposable` interface for wasm resources in TS

iamrajiv:
Fixes: #891

### DIFF
--- a/crates/js-component-bindgen/src/ts_bindgen.rs
+++ b/crates/js-component-bindgen/src/ts_bindgen.rs
@@ -666,7 +666,7 @@ impl<'a> TsInterface<'a> {
         for (resource, source) in self.resources {
             uwriteln!(
                 self.src,
-                "\nexport class {} {{",
+                "\nexport class {} implements Partial<Disposable> {{",
                 resource.to_upper_camel_case()
             );
             if !source.has_constructor {
@@ -676,6 +676,7 @@ impl<'a> TsInterface<'a> {
                 uwriteln!(self.src, "private constructor();");
             }
             self.src.push_str(&source.src);
+            uwriteln!(self.src, "  [Symbol.dispose]?: () => void;");
             uwriteln!(self.src, "}}")
         }
         if self.src.is_empty() {
--- a/packages/jco/test/fixtures/wits/disposable-resources/disposable-resources.wit
+++ b/packages/jco/test/fixtures/wits/disposable-resources/disposable-resources.wit
@@ -1,0 +1,31 @@
+package test:disposable-resources;
+
+interface resources {
+  /// A simple resource for testing Disposable interface generation
+  resource file-handle {
+    constructor(path: string);
+    read: func() -> list<u8>;
+    write: func(data: list<u8>);
+  }
+
+  /// Another resource to test multiple resources
+  resource connection {
+    constructor(url: string);
+    send: func(data: string) -> result<string, string>;
+  }
+
+  open-file: func(path: string) -> file-handle;
+  create-connection: func(url: string) -> connection;
+}
+
+world disposable-world {
+  import resources;
+
+  /// Also test a world-level resource
+  resource database {
+    constructor(connection-string: string);
+    query: func(sql: string) -> list<string>;
+  }
+
+  export run: func() -> string;
+}
--- a/packages/jco/test/typescript.js
+++ b/packages/jco/test/typescript.js
@@ -89,4 +89,60 @@ suite(`TypeScript`, async () => {
             )
         );
     });
+
+    test(`Resource Disposable interface generation`, async () => {
+        const witSource = await readFile(
+            fileURLToPath(
+                new URL(
+                    `./fixtures/wits/disposable-resources/disposable-resources.wit`,
+                    import.meta.url
+                )
+            ),
+            'utf8'
+        );
+        const component = await componentNew(
+            await componentEmbed({
+                witSource,
+                dummy: true,
+            })
+        );
+
+        const { files } = await transpile(component, { name: 'disposable' });
+
+        const mainDtsSource = new TextDecoder().decode(
+            files['disposable.d.ts']
+        );
+
+        assert.ok(
+            mainDtsSource.includes(
+                `export class Database implements Partial<Disposable> {`
+            ),
+            'Database resource should implement partial Disposable interface'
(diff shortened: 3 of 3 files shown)


crates/js-component-bindgen/src/ts_bindgen.rs
```diff
@@ -392,12 +398,21 @@ pub fn ts_bindgen(
         uwriteln!(bindgen.src, "}}");
     }
 
+    let mut final_src = String::new();
+    if bindgen.has_world_resources && !bindgen.is_guest {
+        final_src.push_str("// Resources implement the Disposable interface for automatic cleanup\n");
+        final_src.push_str("interface Disposable {\n");
+        final_src.push_str("  [Symbol.dispose](): void;\n");
+        final_src.push_str("}\n\n");
```
vados-cosmonic: This should be built in to Typescript versions 5.2 and above, did you find that you had to add this?

I'm fine with *requiring* TS 5.2+ to have this feature (and updating docs as necessary).

iamrajiv: yeah sure @vados-cosmonic i didnt know the version just checked builtin disposable interface is available so i will remove this

crates/js-component-bindgen/src/ts_bindgen.rs
```diff
@@ -47,6 +47,8 @@ struct TsBindgen {
     /// definitions. This is used to generate `/// <reference path="..." />`
     /// directives at the top of the file.
     references: BTreeSet<String>,
+
+    has_world_resources: bool,
```
vados-cosmonic: Hey we probably don't need this anymore right? This is no longer being used (we just add the `implements Disposable` since it's built in).

packages/jco/test/fixtures/wits/disposable-resources/disposable-resources.wit
```diff
@@ -0,0 +1,31 @@
+package test:disposable-resources;
```
vados-cosmonic, line 1: This is a great test! Thanks for testing both resources in interfaces and interfaces exposed bare in the world. 

That said, do we need *both* `file-handle` and `connection`?

vados-cosmonic (commented):
Hey @iamrajiv thank for updating this, it looks *almost* perfect! I think we could get the PR even smaller :)

packages/jco/test/fixtures/wits/disposable-resources/disposable-resources.wit
iamrajiv: yeah @vados-cosmonic i mean i added this for test coverage having both file handle and connection serves a purpose it tests that multiple resources in the same interface all correctly implement disposable

crates/js-component-bindgen/src/ts_bindgen.rs
iamrajiv: ah my bad forgot to remove these thanks done @vados-cosmonic

packages/jco/test/fixtures/wits/disposable-resources/disposable-resources.wit
vados-cosmonic: Thanks for the explanation, that makes sense to me!

vados-cosmonic:
Thanks for the contribution @iamrajiv ! should be ready to go now :rocket:

vados-cosmonic:
Hey so @iamrajiv one interesting thing that came up which is quite interesting -- by making the classes implement `Disposable` we're actually also affecting exported classes as well. 

I think that for now we can ensure that objects are `Partial<Disposable>` only to remove the strict requirement that they must all implement a dispose (in the case where there's nothing special to do for a Resource). This should also avoid making this a breaking change (existing code would no longer transpile)

crates/js-component-bindgen/src/ts_bindgen.rs
```diff
@@ -676,6 +676,7 @@ impl<'a> TsInterface<'a> {
                 uwriteln!(self.src, "private constructor();");
             }
             self.src.push_str(&source.src);
+            uwriteln!(self.src, "  [Symbol.dispose]?: () => void;");
```
rajsite, line 679: @vados-cosmonic @iamrajiv this change seems to break my usage because if actually leveraging explicit resource management in TypeScript on a `using` declaration this type being optional is incorrect:
image

Before this change I had manually defined a `wit-patches.d.ts`:

```ts
declare module 'wasi:io/streams@0.2.3' {
  interface OutputStream extends Disposable {
    [Symbol.dispose]: () => void;
  }
}

```

but after upgrading these new types are incompatible with both my patch and with the `using` feature (which expects a non-optional type to be defined).

If I downgrade to before the type change (couldn't figure out a workaround to override the type) then TypeScript transpiles out  `using` when building and it works great!:
```js
const handle = (_request, responseOut) => {
    const fields = new Fields();
    const outgoingResponse = new OutgoingResponse(fields);
    const outgoingBody = outgoingResponse.body();
    {
        const env_1 = { stack: [], error: void 0, hasError: false };
        try {
            const outputStream = __addDisposableResource(env_1, outgoingBody.write(), false);
            outputStream.blockingWriteAndFlush(new Uint8Array(new TextEncoder().encode('Hello from Bundled TypeScript!\n')));
        }
        catch (e_1) {
            env_1.error = e_1;
            env_1.hasError = true;
        }
        finally {
            __disposeResources(env_1);
        }
    }
    outgoingResponse.setStatusCode(200);
    OutgoingBody.finish(outgoingBody, undefined);
    ResponseOutparam.set(responseOut, { tag: 'ok', val: outgoingResponse });
};
```

vados-cosmonic: Thanks for the report @rajsite !   The reason I introduced this as optional was to avoid breaking having this addition be a breaking change for people who exported resources and didn't implement a `dispose` -- it looks like this implementation doesn't work properly though, I will look into this and get a fix out soon!

vados-cosmonic: Please see:

https://github.com/bytecodealliance/jco/pull/954
https://github.com/bytecodealliance/jco/pull/955

Let's continue discussion there -- in the future it would be preferable to file new issues (we can link to the PR/code line) and discuss there!

PR #956, opened 2025-08-22
fix(p2-shim): nsToDateTime modulo with correct number

lrowe:
Avoids error when running in docker where timestamp is zero.

### DIFF
--- a/packages/preview2-shim/lib/nodejs/filesystem.js
+++ b/packages/preview2-shim/lib/nodejs/filesystem.js
@@ -42,7 +42,7 @@ const isMac = platform === 'darwin';
 const nsMagnitude = 1_000_000_000_000n;
 function nsToDateTime(ns) {
     const seconds = ns / nsMagnitude;
-    const nanoseconds = Number(ns % seconds);
+    const nanoseconds = Number(ns % nsMagnitude);
     return { seconds, nanoseconds };
 }
 


vados-cosmonic:
Hey @lrowe thanks for contributing this! Would you mind explaining a bit more about what happened? Did `seconds` end up being zero due to the timestamp being zero? Also was the timestamp not set to UNIX epoch?

lrowe:
I was running componentize-js from within a Dockerfile. Docker sets all timestamps to zero to produce consistent output so got a divide by zero error here.

I assume the intent was to divmod into seconds and nanoseconds in which case there should likely be a Math.floor around the seconds too to make that an integer.

vados-cosmonic:
> Docker sets all timestamps to zero to produce consistent output so got a divide by zero error here.

This part is what I'm unsure about -- I thought this was only enabled if you were attempting a reproducible build -- i.e. via `SOURCE_DATE_EPOCH=0` or something like that. 

I see now that what you're talking about is *filesystem stat time*, since those are where that gets called. 

Can you help me with a reproduction? When I try locally with this Dockerfile:

```dockerfile
FROM node:alpine
RUN date >> output.txt
RUN node -e "console.log(JSON.stringify(require('fs').statSync('./output.txt'), null, 2))" >> node-output.txt
```

I can't reproduce the phenomenon where fields like `atimeMs` and others are set to 0:

```
# cat node-output.txt
{
  "dev": 81,
  "mode": 33188,
  "nlink": 1,
  "uid": 0,
  "gid": 0,
  "rdev": 0,
  "blksize": 4096,
  "ino": 5278232,
  "size": 29,
  "blocks": 8,
  "atimeMs": 1755850496441.52,
  "mtimeMs": 1755850496441.52,
  "ctimeMs": 1755850496441.52,
  "birthtimeMs": 1755850496441.52
}
```

lrowe:
I can try and minimise tomorrow as its late here. I'm building with podman and calling a makefile from the Dockerfile. The full traceback I get without my sed version of the patch applied is:
```
node_modules/.bin/componentize-js -w wit -o target/helloworld.wasm helloworld.js
file:///examples/wasmtime-component-js/node_modules/@bytecodealliance/preview2-shim/lib/nodejs/filesystem.js:45
  const nanoseconds = Number(ns % seconds);
                                ^

RangeError: Division by zero
    at nsToDateTime (file:///examples/wasmtime-component-js/node_modules/@bytecodealliance/preview2-shim/lib/nodejs/filesystem.js:45:33)
    at Descriptor.statAt (file:///examples/wasmtime-component-js/node_modules/@bytecodealliance/preview2-shim/lib/nodejs/filesystem.js:280:28)
    at trampoline23 (file:///examples/wasmtime-component-js/node_modules/@bytecodealliance/componentize-js/lib/spidermonkey-embedding-splicer.js:2217:34)
    at wit-component:shim.indirect-wasi:filesystem/types@0.2.3-[method]descriptor.stat-at (wasm://wasm/wit-component:shim-ca8b7fb6:wasm-function[25]:0x2db)
    at wit-component:adapter:wasi_snapshot_preview1._ZN22wasi_snapshot_preview18bindings4wasi10filesystem5types10Descriptor7stat_at17h37f80d5695f6b119E (wasm://wasm/wit-component:adapter:wasi_snapshot_preview1-0001912a:wasm-function[87]:0x3b01)
    at wit-component:adapter:wasi_snapshot_preview1.path_filestat_get (wasm://wasm/wit-component:adapter:wasi_snapshot_preview1-0001912a:wasm-function[56]:0x253f)
    at wit-component:shim.adapt-wasi_snapshot_preview1-path_filestat_get (wasm://wasm/wit-component:shim-ca8b7fb6:wasm-function[5]:0x1d3)
    at splicer_component.wasm._ZN4wasi13lib_generated17path_filestat_get17h52d9dede891736b6E (wasm://wasm/splicer_component.wasm-01127772:wasm-function[6893]:0x3147e4)
    at splicer_component.wasm._ZN3std3sys2fs8metadata17h0182fd3b7e770833E (wasm://wasm/splicer_component.wasm-01127772:wasm-function[6794]:0x30f26b)
    at splicer_component.wasm._ZN3std4path4Path6is_dir17hbb3ef815f6caa9bfE (wasm://wasm/splicer_component.wasm-01127772:wasm-function[6796]:0x30f72d)

Node.js v24.6.0
```

vados-cosmonic:
Thank you!  A minimal example would be fantastic -- didn't realize you were using `podman` and not `docker` -- the behavior could be different in that case. A repro would go a long way to helping me reproduce and understand! :bow:

lrowe:
And in wash-clocks-0.2.3/package.wit we have:

```
  record datetime {
    seconds: u64,
    nanoseconds: u32,
  }
```

So I think I will update this PR to include the Math.floor too

vados-cosmonic:
Yup, that's right -- let's get this in. Thanks for putting it side by side, it was definitely modulo-ing by the wrong number!

vados-cosmonic:
Would you mind changing the commit title to:

```
fix(p2-shim): ...
```

This way the automation can pick up the changes and categorize them towards p2-shim.

vados-cosmonic (approved):

PR #963, opened 2025-08-25
fix(p2-shim): types and instantiation code for shim object

vados-cosmonic:
This commit fixes some incorrect typing and hard-to-use instantiation logic for the `WASIShim` object. While this feature is somewhat experimental and likely to change in the the future, it should at least be a bit easier to use and reflect the APIs appropriately.


vados-cosmonic:
Transient failures in GH actions (`clippy` is failing due to HTTP errors/failure to download)

PR #1071, opened 2025-10-27
chore(ci): switch to windows-latest runners

### LINKED ISSUE #1038: Return to latest windows on CI Runners
Recently the windows runners changed to Windows Server 2025, and it is causing stalls/hangs on conformance testing. 

This change happened this summer and this is reflected in sudden and repeated test failures happening around the release of 1.15.0:

Image

Obviously the release nor the CI version bumps that came after caused this breakage, the most obvious culprit is the change of Github's CI runners under the cover (note that tests also pass for all other platforms). 

While the conformance tests *shouldn't* hang on windows, they currently do, and it would be great if someone with a windows machine/that spends lots of time in windows could look at this and attempt to reproduce or narrow down the test that is causing the hang. 

The conformance test suite is generated, so testing is a bit more involved/annoying, but it's possible to debug this even from CI itself on a fork and get a reasonably fast dev loop going.

NOTE: Somewhat sneakily, the revert to `windows-2022` runners happened in [a dependabot PR](https://github.com/bytecodealliance/jco/pull/1037). To get started trying to track down the issue with the more recent windows runner, the first step is to switch back to `windows-latest` (or more accurately `windows-2025`).

vados-cosmonic:
Resolves #1038

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -60,7 +60,7 @@ jobs:
       matrix:
         os:
           - ubuntu-22.04 # https://github.com/puppeteer/puppeteer/issues/12818#issuecomment-2593659514
-          - windows-2022
+          - windows-latest
           - macos-latest
         node:
           - 18.x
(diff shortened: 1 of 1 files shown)


vados-cosmonic:
looks like we're still hanging here -- test-wasi-node seems to be now never exiting.

PR #1213, opened 2026-01-26
feat(p2-shim): add sandboxing configuration to WASIShim

andreiltd:
Add support for configuring filesystem preopens, environment variables, command-line arguments, and network access through the WASIShim constructor.

Closes: https://github.com/bytecodealliance/jco/issues/466

### DIFF
--- a/packages/preview2-shim/README.md
+++ b/packages/preview2-shim/README.md
@@ -59,26 +59,30 @@ sandboxing to restrict what guests can access.
 
 ### Using WASIShim for sandboxing
 
-The `WASIShim` class accepts configuration options to control access:
+The `WASIShim` class accepts a `sandbox` configuration option to control access:
 
 ```js
 import { WASIShim } from '@bytecodealliance/preview2-shim/instantiation';
 
 // Fully sandboxed - no filesystem, network, or env access
 const sandboxedShim = new WASIShim({
+  sandbox: {
     preopens: {},           // No filesystem access
     env: {},                // No environment variables
     args: ['arg1'],         // Custom arguments
     enableNetwork: false,   // Disable network access
+  }
 });
 
 // Limited filesystem access - map virtual paths to host paths
 const limitedShim = new WASIShim({
+  sandbox: {
     preopens: {
-        '/data': '/tmp/guest-data',  // Guest sees /data, maps to /tmp/guest-data
-        '/config': '/etc/app'        // Guest sees /config, maps to /etc/app
+      '/data': '/tmp/guest-data',  // Guest sees /data, maps to /tmp/guest-data
+      '/config': '/etc/app'        // Guest sees /config, maps to /etc/app
     },
     env: { 'ENV1': '42' },           // Only expose specific env vars
+  }
 });
 
 const component = await instantiate(loader, sandboxedShim.getImportObject());
@@ -92,8 +96,8 @@ const component = await instantiate(loader, sandboxedShim.getImportObject());
   Multiple instances with different configurations will not affect each other.
 - The direct preopen functions (`_setPreopens`, `_clearPreopens`, etc.) modify global state and
   affect all components not using `WASIShim` with explicit configuration. For isolation, prefer
-  using `WASIShim` with the `preopens` and `env` options.
-- When `enableNetwork: false`, all socket and HTTP operations will throw "access-denied" errors.
+  using `WASIShim` with the `sandbox` option containing `preopens` and `env`.
+- When `sandbox.enableNetwork: false`, all socket and HTTP operations will throw "access-denied" errors.
 
 [jco]: https://www.npmjs.com/package/@bytecodealliance/jco
 
--- a/packages/preview2-shim/lib/browser/filesystem.js
+++ b/packages/preview2-shim/lib/browser/filesystem.js
@@ -6,6 +6,25 @@ export { _setCwd } from './config.js';
 
 const { InputStream, OutputStream } = streams;
 
+/**
+ * @typedef {Object} FileDataEntry
+ * @property {Record<string, FileDataEntry>} [dir] - Directory contents (present for directories)
+ * @property {Uint8Array|string} [source] - File contents (present for files)
+ */
+
+/**
+ * @typedef {FileDataEntry} FileData
+ * Root file data structure representing a filesystem tree.
+ * Each entry is either a directory (has `dir` property) or a file (has `source` property).
+ * @example
+ * // A simple filesystem with one directory containing one file:
+ * const fileData = {
+ *   dir: {
+ *     'myfile.txt': { source: new Uint8Array([72, 101, 108, 108, 111]) }
+ *   }
+ * };
+ */
+
 export function _setFileData(fileData) {
     _fileData = fileData;
     _rootPreopen[0] = new Descriptor(fileData);
@@ -324,7 +343,7 @@ export const preopens = {
 
 /**
  * Replace all preopens with the given set.
- * @param {Record<string, object>} preopensConfig - Map of virtual paths to file data entries
+ * @param {Record<string, FileData>} preopensConfig - Map of virtual paths to file data entries
  */
 export function _setPreopens(preopensConfig) {
     _preopens = [];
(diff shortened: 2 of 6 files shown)


packages/preview2-shim/README.md
```diff
@@ -51,6 +51,50 @@ const component = await instantiate(loader, new WASIShim().getImportObject());
 // TODO: Code that uses your component's exports goes here.
 ```
 
+## Sandboxing
```
vados-cosmonic, line 54: Thanks for adding this documentation! Would you mind adding this also to the markdown for the Jco book? (separate issue is also fine, that way anyone with free time could pick it up as a good first issue)!

packages/preview2-shim/README.md
```diff
@@ -51,6 +51,50 @@ const component = await instantiate(loader, new WASIShim().getImportObject());
...
+
+```js
+import { WASIShim } from '@bytecodealliance/preview2-shim/instantiation';
+
+// Fully sandboxed - no filesystem, network, or env access
+const sandboxedShim = new WASIShim({
+    preopens: {},           // No filesystem access
+    env: {},                // No environment variables
+    args: ['arg1'],         // Custom arguments
+    enableNetwork: false,   // Disable network access
```
vados-cosmonic, line 73: What do you think about putting these under a `sandbox` member? My thinking here is that since all of these options are somewhat related and there may be other configuration points for the shims in the future I like the idea of separating these sandbox specific options from anything that might come after.

packages/preview2-shim/README.md
```diff
@@ -51,6 +51,50 @@ const component = await instantiate(loader, new WASIShim().getImportObject());
...
+    },
+    env: { 'ENV1': '42' },           // Only expose specific env vars
+});
+
+const component = await instantiate(loader, sandboxedShim.getImportObject());
+```
+
+### Notes on sandboxing
+
+- By default (when no options are passed), the shim is providing full access to match typical
```
vados-cosmonic, line 93: Thanks for including this section -- this is definitely going to turn into a reference for people.

packages/preview2-shim/lib/browser/filesystem.js
```diff
@@ -322,6 +322,56 @@ export const preopens = {
     },
 };
 
+/**
+ * Replace all preopens with the given set.
+ * @param {Record<string, object>} preopensConfig - Map of virtual paths to file data entries
```
vados-cosmonic: It would be great to lay out the type of that `object` via a `@typedef` block like in other bits of the code. 

Someday, preview2-shim will be ported to typescript (along with all the other sub projects) but until that time IDEs can pick up on the extended JSDoc

packages/preview2-shim/lib/browser/filesystem.js
```diff
@@ -322,6 +322,56 @@ export const preopens = {
...
+    const descriptor = new Descriptor(fileData);
+    _preopens.push([descriptor, virtualPath]);
+    if (virtualPath === '/') {
+        _rootPreopen = [descriptor, virtualPath];
+    }
+}
+
+/**
+ * Clear all preopens, giving the guest no filesystem access.
+ */
```
vados-cosmonic, line 373:
```suggestion
 * Clear all preopens, giving the guest no filesystem access.
 * 
 * This functionality exists mostly to maintain backwards compatibility. Prefer setting preopens
 * via `WASIShim` rather than making top level changes to preopens using these functions.
 */
```

Not sure on the wording here, but what do you think about adding a note to this effect? The idea is to get people away from just importing and using the shims out of the box, but instead using an explicit `WASIShim` object -- more leverage, isolation (somewhat), etc.

packages/preview2-shim/lib/browser/filesystem.js
```diff
@@ -322,6 +322,56 @@ export const preopens = {
...
+    _preopens = [];
+    _rootPreopen = null;
+}
+
+/**
+ * Get current preopens configuration.
+ * @returns {Array<[Descriptor, string]>} Array of [descriptor, virtualPath] pairs
+ */
+export function _getPreopens() {
+    return [..._preopens];
```
vados-cosmonic: Does it matter if people can change the actual descriptors themselves? We've certainly stopped them from adding/removing stuff from the list here, but they can still edit individual descriptors and an edit in the resulting array will affect the original `_preopens` array

packages/preview2-shim/lib/browser/filesystem.js
```diff
@@ -322,6 +322,56 @@ export const preopens = {
...
+}
+
+/**
+ * Create a preopen descriptor for file data.
+ * This is used internally to create isolated preopen instances.
+ * @param {object} fileData - The file data object representing the directory
```
vados-cosmonic: A `@typedef` would be great to go with this `object` here/reused from elsewhere

packages/preview2-shim/lib/common/instantiation.js
```diff
@@ -66,15 +89,66 @@ export class WASIShim {
...
+     * @param {object} [config.filesystem] - Custom filesystem shim
+     * @param {object} [config.io] - Custom I/O shim
+     * @param {object} [config.random] - Custom random shim
+     * @param {object} [config.clocks] - Custom clocks shim
+     * @param {object} [config.sockets] - Custom sockets shim
+     * @param {object} [config.http] - Custom HTTP shim
+     * @param {Record<string, string>} [config.preopens] - Filesystem preopens mapping (virtual path -> host path)
+     * @param {Record<string, string>} [config.env] - Environment variables
+     * @param {string[]} [config.args] - Command-line arguments
+     * @param {boolean} [config.enableNetwork=true] - Whether to enable network access
```
vados-cosmonic: Thanks for spelling this out, I should have used a @typedef here but this also works/will be picked up by IDEs properly. 

Also, I think you *might* be able to use the `WasiShimConfig` type that you added in the `.d.ts` here and IDEs will pick it up?

packages/preview2-shim/README.md
andreiltd: I've added (copied mostly) a section to the book.

packages/preview2-shim/README.md
andreiltd: That makes sense, yes!

packages/preview2-shim/lib/browser/filesystem.js
andreiltd: IIUC, the fix here would be to make copy of an array rather than giving back a reference to original array? That is what I fixed but let me know if you meant something else. :-)

vados-cosmonic: Ah yes sorry if it wasn't clear -- you're already making a copy of the array/making a new array, but I was wondering if it mattered if the *elements* were still accessible. 

In the copied array that gets returned, you can still access (and modify) the individual `Descriptor` elements -- should that be preventable after they've been added?

andreiltd: Good question, I'm not sure how we achieve that. We could maybe map preopen tuples `(Descriptor, Path)` and only return path part of it? But also isn't `_getPreopens` part of public API even though its name suggests otherwise? 🙂

vados-cosmonic: Yeah, actually a bunch of these are part of the public API... But ideally we'd remove them at some point and people should only use the `WASIShim` object that they build themselves.

That said... maybe we could return the original file data object? Could we store it along with the descriptor when it gets made? Maybe it's just not that important if the Descriptors are passed back/available to users, as hosts are free to do lots of stuff.

vados-cosmonic: Let's resolve this and deal with it in the future before any breaking changes -- at this point trying to protect ourselves from the user that is acting as the host (and has full access to the code) is probably silly.

andreiltd: Maybe we can create an issue and address this in a separate venue? I'm happy to take on that as well. We will need symmetrical changes in the `nodejs` implementation and those helpers were added here only so that common instantiation API can interact with both node and browser? WDYT?

Issue #1206, opened 2026-01-20
Node.js Compat


MendyBerger:
Many npm packages fail to run with `jco transpile` because they depend on NodeJS built-ins and globals like `Buffer` and `process` that are missing in wasi environments. I'd like to propose adding a Node compatibility mode to automatically polyfill these APIs and inject required globals during the componentization process.

This integration would allow jco to handle `node:*` imports and common globals out of the box, making the massive npm ecosystem significantly more accessible to the Component Model with a simple flag like `--node-compat`.

MendyBerger:
See https://github.com/unjs/unenv for a polyfill with pretty good api coverage

vados-cosmonic:
Experimental NodeJS Compat is now out as of Jco 1.33.0 (via `jco-std`). The Jco book has been updated, and while coverage is not perfect (there are lots of paradigms that do not map cleanly to components/can't cross the component->host boundary properly), I think we don't need a general issue any more for this and can move on to specific bugs/features.

Note there are also some examples of the new NodeJS API support in action:

* [node-compat](https://github.com/bytecodealliance/jco/tree/main/examples/components/node-compat)
* [node-grpc-server](https://github.com/bytecodealliance/jco/tree/main/examples/components/node-grpc-server)
* [node-mcp-server](https://github.com/bytecodealliance/jco/tree/main/examples/components/node-mcp-server)

PR #1223, opened 2026-01-29
Added node compat example

MendyBerger:
See also #1206 

I'm not sure that this fully resolves the issue, as setting this up took quite a bit of research on how to make unenv work nicely with rollup, and someone with a slightly different build setup would have to figure it out on their own.

Having jco bundle in all this polyfill code with the glue code would make the experience so much better.

Also, turns out unenv has a lot more holes than I originally thought. A significant amount of node apis just trap, and there's no clear documentation on what parts are actually implemented.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18,13 +18,14 @@
         "examples/components/http-server-fetch-handler",
         "examples/components/http-server-hono",
         "examples/components/http-server-hono-with-bindings",
+        "examples/components/node-compat",
         "examples/components/node-fetch",
         "examples/components/string-reverse",
         "examples/components/string-reverse-upper",
-        "examples/components/webidl-book-library",
-        "examples/components/ts-resource-import",
         "examples/components/ts-resource-export",
-        "examples/components/typegen-async-export"
+        "examples/components/ts-resource-import",
+        "examples/components/typegen-async-export",
+        "examples/components/webidl-book-library"
       ],
       "devDependencies": {
         "@actions/github": "^6.0.1",
@@ -88,6 +89,14 @@
         "rolldown": "^1.0.0-beta.47",
         "terminate": "^2.8.0",
         "typescript": "^5.9.2"
+      }
+    },
+    "examples/components/node-compat": {
+      "version": "0.0.1",
+      "dependencies": {
+        "@bytecodealliance/jco": "^1.17.0",
+        "rolldown": "^1.0.0-beta.47",
+        "unenv": "^2.0.0-rc.24"
       }
     },
     "examples/components/node-fetch": {
@@ -3198,7 +3207,6 @@
       "cpu": [
         "arm64"
       ],
-      "dev": true,
       "license": "MIT",
       "optional": true,
       "os": [
@@ -3215,7 +3223,6 @@
       "cpu": [
         "arm64"
       ],
-      "dev": true,
       "license": "MIT",
       "optional": true,
       "os": [
@@ -3232,7 +3239,6 @@
       "cpu": [
         "x64"
       ],
-      "dev": true,
       "license": "MIT",
       "optional": true,
       "os": [
@@ -3249,7 +3255,6 @@
       "cpu": [
         "x64"
       ],
-      "dev": true,
       "license": "MIT",
       "optional": true,
       "os": [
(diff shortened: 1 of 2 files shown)


examples/components/node-compat/index.js
```diff
@@ -0,0 +1,57 @@
+import { URL } from "node:url";
```
vados-cosmonic: Would you mind adding a comment about why this line and teh one below it are special/the focus of the example and where the `node:*` import will actually end up coming from?

examples/components/node-compat/index.js
```diff
@@ -0,0 +1,57 @@
...
+import * as querystring from "node:querystring";
+
+const testUrls = [
```
vados-cosmonic:
```suggestion
const TEST_URLS = [
```

examples/components/node-compat/index.js
```diff
@@ -0,0 +1,57 @@
...
+    console.log(`  Hostname: ${urlParts.hostname}`);
+    console.log(`  Port: ${urlParts.port || "(default)"}`);
+    console.log(`  Pathname: ${urlParts.pathname}`);
+    console.log(`  Hash: ${urlParts.hash || "(none)"}`);
+    console.log(`  Queries:`);
+    for (const [key, value] of Object.entries(urlParts.queries)) {
+        console.log(`    ${key}: ${value}`);
+    }
+    console.log();
+}
```
vados-cosmonic: What do you think about building a string here rather than `console.log`ing?

examples/components/node-compat/index.js
```diff
@@ -0,0 +1,57 @@
...
+
+export function printUrlBufferInfo(urlString) {
+    const buffer = Buffer.from(urlString, "utf8");
+    console.log("  Buffer info:");
+    console.log(`    Length: ${buffer.length} bytes`);
+    console.log(`    Hex: ${buffer.toString("hex")}`);
+    console.log(`    Base64: ${buffer.toString("base64")}`);
+    console.log();
+}
```
vados-cosmonic: Same note about building a string here

examples/components/node-compat/index.js
```diff
@@ -0,0 +1,57 @@
...
+    console.log(`  Hash: ${urlParts.hash || "(none)"}`);
+    console.log(`  Queries:`);
+    for (const [key, value] of Object.entries(urlParts.queries)) {
+        console.log(`    ${key}: ${value}`);
+    }
+    console.log();
+}
+
+export function printUrlBufferInfo(urlString) {
+    const buffer = Buffer.from(urlString, "utf8");
```
vados-cosmonic: A note here on the fact that `Buffer` is actually a NodeJS-ism rather than a web platform or standards backed global object that is present. 

An explicit import of Buffer from `node:buffer` might be nice as well.

examples/components/node-compat/index.js
vados-cosmonic, line 1: What do you think about renaming this file to `component.js` ? I'm not sure `index.js` is very descriptive/fits the server-side paradigm.

Also, would you mind adding a descriptive README that matches some of the other examples? Updating `components/README.md` would also be nice to list this example next to the others.

examples/components/node-compat/rollup.config.mjs
```diff
@@ -0,0 +1,25 @@
+import nodeResolve from "@rollup/plugin-node-resolve";
+import commonjs from '@rollup/plugin-commonjs';
+import alias from '@rollup/plugin-alias';
+import inject from '@rollup/plugin-inject';
+import { defineEnv } from "unenv";
+
+const { env } = defineEnv({});
```
vados-cosmonic: Would you mind mind adding a comment (likely before the `defineEnv` line?) about Unenv and what the effect of this line is/how it changes the project and what it makes possible?

examples/components/node-compat/rollup.config.mjs
vados-cosmonic, line 1: Would you consider using [Rolldown](https://rolldown.rs/) here rather than Rollup?

Adopting more of the oxc stack is definitely a long term goal, and you may be able to remove some dependencies (and have things run faster). The config generally looks pretty similar and rolldown is explicitly supported by unenv in the docs)

vados-cosmonic (commented):
Thanks for working on this @MendyBerger -- this example is a great start to showing how to do NodeJS compat ahead of a more built-in integration with Jco/other ecosystem projects.

PR #1249, opened 2026-02-07
feat(preview2-shim): implement browser HTTP fetch support via JSPI

danbopes:
Implement the browser WASI HTTP shim using the fetch() API with JSPI for async suspension, replacing the previous non-functional stubs.

  - Rewrite http.js with Fields, OutgoingRequest, IncomingResponse, IncomingBody, FutureIncomingResponse, and outgoing-handler.handle()
  - Fix Pollable in io.js: add ready(), block(), subscribe() support
  - Fix clocks.js: subscribeDuration(0) yields via setTimeout, fix subscribeInstant this-binding for standalone function calls
  - Fix blockingRead double-reader bug on ReadableStream
  - Add browser tests ported from wasmtime p2_* suite (clocks, poll, HTTP)
  - Replace hand-written WIT deps with canonical versions via wkg

### DIFF
--- a/packages/preview2-shim/lib/browser/http.js
+++ b/packages/preview2-shim/lib/browser/http.js
@@ -1,155 +1,692 @@
-/**
- * @param {import("../../types/interfaces/wasi-http-types").Request} req
- * @returns {string}
- */
-export function send(req) {
-    console.log(`[http] Send (browser) ${req.uri}`);
-    try {
-        const xhr = new XMLHttpRequest();
-        xhr.open(req.method.toString(), req.uri, false);
-        const requestHeaders = new Headers(req.headers);
-        for (let [name, value] of requestHeaders.entries()) {
-            if (name !== 'user-agent' && name !== 'host') {
-                xhr.setRequestHeader(name, value);
-            }
-        }
-        xhr.send(req.body && req.body.length > 0 ? req.body : null);
-        const body = xhr.response
-            ? new TextEncoder().encode(xhr.response)
-            : undefined;
-        const headers = [];
-        xhr.getAllResponseHeaders()
-            .trim()
-            .split(/[\r\n]+/)
-            .forEach((line) => {
-                var parts = line.split(': ');
-                var key = parts.shift();
-                var value = parts.join(': ');
-                headers.push([key, value]);
-            });
-        return {
-            status: xhr.status,
-            headers,
-            body,
-        };
-    } catch (err) {
-        throw new Error(err.message);
+import { streams, poll } from './io.js';
+
+const { InputStream, OutputStream } = streams;
+const { Pollable } = poll;
+
+const symbolDispose = Symbol.dispose || Symbol.for('dispose');
+const utf8Decoder = new TextDecoder();
+const forbiddenHeaders = new Set(['connection', 'keep-alive', 'host']);
+const DEFAULT_HTTP_TIMEOUT_NS = 600_000_000_000n;
+
+// RFC 9110 compliant header validation
+const TOKEN_RE = /^[!#$%&'*+\-.^_`|~0-9A-Za-z]+$/;
+const FIELD_VALUE_RE = /^[\t\x20-\x7E\x80-\xFF]*$/;
+
+function validateHeaderName(name) {
+    if (!TOKEN_RE.test(name)) {
+        throw { tag: 'invalid-syntax' };
     }
 }
+
+function validateHeaderValue(value) {
+    const str = typeof value === 'string' ? value : utf8Decoder.decode(value);
+    if (!FIELD_VALUE_RE.test(str)) {
+        throw { tag: 'invalid-syntax' };
+    }
+}
+
+class Fields {
+    #immutable = false;
+    /** @type {[string, Uint8Array][]} */ #entries = [];
+    /** @type {Map<string, [string, Uint8Array][]>} */ #table = new Map();
+
+    static fromList(entries) {
+        const fields = new Fields();
+        for (const [key, value] of entries) {
+            fields.append(key, value);
+        }
+        return fields;
+    }
+
+    get(name) {
+        const tableEntries = this.#table.get(name.toLowerCase());
+        if (!tableEntries) {
+            return [];
+        }
+        return tableEntries.map(([, v]) => v);
+    }
+
+    set(name, values) {
+        if (this.#immutable) {
+            throw { tag: 'immutable' };
+        }
+        validateHeaderName(name);
+        for (const value of values) {
+            validateHeaderValue(value);
+        }
+        const lowercased = name.toLowerCase();
+        if (forbiddenHeaders.has(lowercased)) {
+            throw { tag: 'forbidden' };
+        }
+        const tableEntries = this.#table.get(lowercased);
+        if (tableEntries) {
+            this.#entries = this.#entries.filter(
+                (entry) => !tableEntries.includes(entry)
+            );
+            tableEntries.splice(0, tableEntries.length);
+        } else {
+            this.#table.set(lowercased, []);
+        }
+        const newTableEntries = this.#table.get(lowercased);
+        for (const value of values) {
+            const entry = [name, value];
+            this.#entries.push(entry);
+            newTableEntries.push(entry);
+        }
+    }
+
+    has(name) {
+        return this.#table.has(name.toLowerCase());
+    }
+
+    delete(name) {
+        if (this.#immutable) {
+            throw { tag: 'immutable' };
+        }
+        const lowercased = name.toLowerCase();
+        const tableEntries = this.#table.get(lowercased);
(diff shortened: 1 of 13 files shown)


packages/preview2-shim/lib/browser/io.js
```diff
@@ -178,19 +182,70 @@ export const error = { Error: IoError };
...
+    ready() {
+        return this.#ready;
+    }
+
+    block() {
+        if (this.#ready) {
+            return Promise.resolve();
+        }
+        return this.#promise;
+    }
```
vados-cosmonic, line 209: This isn't quite right -- block should likely be async because it must return immediately or wait until the pollable is ready.

https://github.com/WebAssembly/WASI/blob/184b0c0e9fd437e5e5601d6e327a28feddbbd7f7/proposals/io/wit/poll.wit#L17

packages/preview2-shim/lib/browser/io.js
```diff
@@ -178,19 +182,70 @@ export const error = { Error: IoError };
...
+
+    [symbolDispose]() {
+        this.#promise = null;
+    }
+}
+
+function pollList(list) {
+    const ready = [];
```
vados-cosmonic, line 223: Let's throw an error if the list is empty here or if the list length is greater than the max size of an u32.

https://github.com/WebAssembly/WASI/blob/184b0c0e9fd437e5e5601d6e327a28feddbbd7f7/proposals/io/wit/poll.wit#L26

packages/preview2-shim/lib/browser/io.js
```diff
@@ -178,19 +182,70 @@ export const error = { Error: IoError };
...
+        list.map((p, i) => p.block().then(() => {
+            const result = [i];
+            for (let j = 0; j < list.length; j++) {
+                if (j !== i && list[j].ready()) {
+                    result.push(j);
+                }
+            }
+            return new Uint32Array(result);
+        }))
+    );
```
vados-cosmonic, line 244: I do wonder if we need this complexity here... I think it's OK to simply go through and count ready pollables? Would you mind explaining the thinking here?

packages/preview2-shim/lib/browser/io.js
```diff
@@ -178,19 +182,70 @@ export const error = { Error: IoError };
...
 }
 
-function pollOne(_poll) {
-    // TODO
+function pollOne(poll) {
+    if (poll.ready()) {
+        return;
+    }
+    return poll.block();
```
vados-cosmonic, line 248: arguably should this function be async and we wait for the `poll.block()` inside this function? 

I think this might just reduce to `poll.block()` (I think `poll-one` might have been eventually removed actually)

packages/preview2-shim/lib/browser/http.js
```diff
@@ -1,155 +1,684 @@
...
-            });
+        return fields;
+    }
+}
+const fieldsLock = Fields._lock;
+delete Fields._lock;
+const fieldsFromEntriesChecked = Fields._fromEntriesChecked;
+delete Fields._fromEntriesChecked;
+
+const defaultHttpTimeout = 600_000_000_000n;
```
vados-cosmonic:
```suggestion
const DEFAULT_HTTP_TIMEOUT_NS = 600_000_000_000n;
```

Would you mind moving this to the top?

packages/preview2-shim/lib/browser/http.js
```diff
@@ -1,155 +1,684 @@
...
+
+class RequestOptions {
+    #connectTimeout = defaultHttpTimeout;
+    #firstByteTimeout = defaultHttpTimeout;
+    #betweenBytesTimeout = defaultHttpTimeout;
+    connectTimeout() {
+        return this.#connectTimeout;
+    }
+    setConnectTimeout(duration) {
+        this.#connectTimeout = duration;
```
vados-cosmonic, line 160: `duration`s are u64s, would you mind checking for negative values here? Small think but would be nice to be sure

packages/preview2-shim/lib/browser/http.js
```diff
@@ -1,155 +1,684 @@
...
+const FIELD_VALUE_RE = /^[\t\x20-\x7E\x80-\xFF]*$/;
+
+function validateHeaderName(name) {
+    if (!TOKEN_RE.test(name)) {
+        throw { tag: 'invalid-syntax' };
+    }
+}
+
+function validateHeaderValue(value) {
+    const str = typeof value === 'string' ? value : new TextDecoder().decode(value);
```
vados-cosmonic: What do you think about reusing the `TextDecoder()` here and later in the file? Could make it a function + initially undefined module global

vados-cosmonic (commented):
This is fantastic work @danbopes , thanks for contributing so much implementation to the browser preview2-shim, it has not gotten any attention for a long time. 

I have a few nitpicks/small changes proposed, please take a look and let me know if you have any questions/disagree!

packages/preview2-shim/lib/browser/io.js
danbopes: This already returns a promise. Making it async would just add another useless microtask.

packages/preview2-shim/lib/browser/io.js
danbopes: The complexity handles the case where none of the pollables are ready yet: we need to actually wait for one to resolve. The first loop (lines 218-224) is the fast path that does exactly what you're describing. It just checks ready and returns immediately. When that finds nothing, we can't spin or return empty (spec says result is always non-empty). `Promise.race` lets us yield to the event loop until the first pollable fires, then we sweep the rest for anything that became ready concurrently. Without `SharedArrayBuffer` (which is the Node.js path), this is the only way to block without freezing the tab.

packages/preview2-shim/lib/browser/io.js
vados-cosmonic: My point was to match the semantics -- the blocking is supposed to happen *inside* `block()`. 

That said, on second thought you're right, as we can't force the user to `await` and we'd be awaiting the promise in any case.

packages/preview2-shim/lib/browser/io.js
vados-cosmonic: Thanks for the explanation, a comment would be great to make this obvious for anyone else who can't immediately tell what this is for.

packages/preview2-shim/lib/browser/http.js
```diff
@@ -1,155 +1,692 @@
...
-            if (name !== 'user-agent' && name !== 'host') {
-                xhr.setRequestHeader(name, value);
+import { streams, poll } from './io.js';
+
+const { InputStream, OutputStream } = streams;
+const { Pollable } = poll;
+
+const symbolDispose = Symbol.dispose || Symbol.for('dispose');
+const utf8Decoder = new TextDecoder();
+const forbiddenHeaders = new Set(['connection', 'keep-alive', 'host']);
```
vados-cosmonic, line 8: Can we turn these into functions that look more like this:

```js
let _SYMBOL_DISPOSE;
function getSymbolDispose() {
    if (_SYMBOL_DISPOSE) { return SYMBOL_DISPOSE; }
    _SYMBOL_DISPOSE = Symbol.dispose || Symbol.for('dispose');
    return SYMBOL_DISPOSE;
}
```

(I'm not picky on the variable names, but I'd argue that these are globals)

This would keep the code trivially treeshakable.

danbopes: `symbolDispose` is used as a computed property key in class bodies (`[symbolDispose]() {}`), which is evaluated at class definition time: a lazy getter can't be used there without rewriting every class. For the other constants (utf8Decoder, forbiddenHeaders, DEFAULT_HTTP_TIMEOUT_NS), they're all direct dependencies of the module's core exports (Fields, OutgoingRequest, RequestOptions), so a bundler can't tree-shake them out regardless. Since this module is consumed as a whole by transpiled components, lazy init would add complexity without a real tree-shaking benefit.

vados-cosmonic: Thanks for the explanation on your rationale here -- these are good points, though I disagree on some of them.

Thinking about it some more given the information hiding code that is strewn around we can't tree shake this module anyway.

vados-cosmonic:
Hey @danbopes thanks for your hard work here and iterating so fast and explaining stuff to me where necessary. 

If I can ask for one more thing -- can you either format your commits to match the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) standard or squash to one commit with a title like the following:

```
feat(p2-shim): implement browser HTTP fetch support via JSPI
```

The tagging is picked up by the automation in-repo, and we don't squash actually -- we fast-forward so that we can (easily) catch non-conforming commit titles in CI

danbopes:
@vados-cosmonic Done

PR #1367, opened 2026-04-07
fix(bindgen): fix async future JS codegen producing invalid output

wondenge:
The async future intrinsic templates produced invalid JavaScript that failed to parse when transpiling any component with future<T> types. These bugs made P3 async futures completely unusable.

- Removed a stray closing parenthesis in the elementTypeRep range check
- Changed self to this in four places inside drop() - in JavaScript class methods, self refers to globalThis, not the instance
- Used the current variant name instead of always emitting futureWrite, which caused a duplicate declaration when rendering futureRead
- Same for futureCancelRead which was emitted for both cancel variants
- Renamed the second futureIdx parameter to futureEndIdx — the duplicate left futureEndIdx undefined at runtime
- Also fixed the matching duplicate in the debug log object
- Added the missing throw keyword before new Error in the instanceof guard, which was constructing an error and silently discarding it
- Removed a broken line that called .remaining(bufferID) without the BUFFER_MGR receiver and had no closing brace
- Added componentIdx to the cancel function parameter list and debug log, since the function body referenced it but never received it
- Added the missing lookup to fetch the future from the global map before comparing element type reps
- Replaced streamEndIdx with futureEndIdx, a copy-paste leftover from the stream equivalent
- Added the missing closing parenthesis in the instanceof check

### DIFF
--- a/crates/js-component-bindgen/src/intrinsics/p3/async_future.rs
+++ b/crates/js-component-bindgen/src/intrinsics/p3/async_future.rs
@@ -217,7 +217,7 @@ impl AsyncFutureIntrinsic {
                             if (!args?.elementTypeRep || typeof args.elementTypeRep !== 'number') {{
                                 throw new TypeError('missing elementTypeRep [' + args.elementTypeRep + ']');
                             }}
-                            if (args.elementTypeRep <= 0 || args.elementTypeRep > 2_147_483_647 ))  {{
+                            if (args.elementTypeRep <= 0 || args.elementTypeRep > 2_147_483_647)  {{
                                 throw new TypeError('invalid  elementTypeRep [' + args.elementTypeRep + ']');
                             }}
                             this.#elementTypeRep = args.elementTypeRep;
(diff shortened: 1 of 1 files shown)


vados-cosmonic (approved):

vados-cosmonic:
Hey @wondenge thanks for working on this and submitting the fixes -- as you can tell the `future` implementation is not yet done! Glad to get these fixes in fast and release a new version (hopefully this week) -- hopefully with an actually working futures implementation!

wondenge:
> Hey @wondenge thanks for working on this and submitting the fixes -- as you can tell the `future` implementation is not yet done! Glad to get these fixes in fast and release a new version (hopefully this week) -- hopefully with an actually working futures implementation!

Happy to help move futures forward. Async is critical for the browser use case I'm working on.

I found these while building tooling that scans jco's transpiled output to understand the boundary structure. That tooling made it straightforward to spot the codegen issues since I was already reading the generated JS closely.

Looking forward to the release and a working futures implementation. Let me know if there's anything else I can help with on the async side.

PR #1373, opened 2026-04-10
fix(bindgen): use IndexMap for deterministic export iteration order

wondenge:
## Description

Running `jco transpile` on the same unchanged component twice can produce different JS output. The `postReturn` and `callback` variable assignments pick different representative export names across runs. This shows up when multiple `cabi_post_*` or `[callback]` exports alias the same core function, which wit-bindgen does pretty commonly for shared deallocation logic. The wasm is correct and runtime behaviour is fine. It is just the JS picking a different name each time, which breaks reproducible builds and makes diffing transpiled output unreliable.

[`Translation::exports()`](https://github.com/bytecodealliance/jco/blob/main/crates/js-component-bindgen/src/core.rs#L204) returns a `HashMap<String, EntityIndex>`, and when [`core_export_var_name`](https://github.com/bytecodealliance/jco/blob/main/crates/js-component-bindgen/src/transpile_bindgen.rs#L3923) iterates it with `find_map` to resolve a function index back to an export name, `HashMap`'s iteration order varies between runs (Rust randomizes the hash seed per process). This changed from `&IndexMap` to `HashMap` in [`39e6972d`](https://github.com/bytecodealliance/jco/commit/39e6972d886f7e0997db169c0ed7991eb1943d85) during the wasmtime deps update when `Module.exports` moved to `Atom` keyed maps, the conversion to `String` keys needed a new map, and `HashMap` was picked instead of `IndexMap`.

This PR switches the return type back to `IndexMap` so export iteration preserves wasm binary order. `indexmap = "2"` is added to `Cargo.toml` as it's already in `Cargo.lock` (resolves to 2.13.0) as a transitive dependency, so no new crates to pull in. Verified across 5 transpile runs on the same component and the output is now identical every time. Also tested with an instrumented component (additional imports, trampolines, multi-memory) same result, fully deterministic.

Before: Two runs of the same component produced different JS:

```diff
-    postReturn0 = exports1['cabi_post_example:repro/handler@0.1.0#check-string'];
+    postReturn0 = exports1['cabi_post_example:repro/handler@0.1.0#check-alias'];
```

After: Five runs, identical output every time `diff` is now empty:

```bash
$ diff /tmp/run1/component.js /tmp/run2/component.js
$
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -789,6 +789,7 @@ dependencies = [
  "anyhow",
  "base64",
  "heck 0.5.0",
+ "indexmap",
  "log",
  "semver",
  "wasm-encoder 0.245.1",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -789,7 +789,6 @@ dependencies = [
  "anyhow",
  "base64",
  "heck 0.5.0",
- "indexmap",
  "log",
  "semver",
  "wasm-encoder 0.245.1",
--- a/crates/js-component-bindgen/Cargo.toml
+++ b/crates/js-component-bindgen/Cargo.toml
@@ -28,7 +28,6 @@ all-features = true
 anyhow = { workspace = true }
 base64 = { workspace = true, features = [ "alloc" ] }
 heck = { workspace = true }
-indexmap = "2"
 log = { workspace = true }
 semver = { workspace = true }
 wasm-encoder = { workspace = true }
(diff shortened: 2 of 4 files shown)


vados-cosmonic:
Great catch @wondenge, thanks for the contribution!

I'm going to add a little bit (probably a test) and then get this merged.

vados-cosmonic (approved):
LGTM 🚀

wondenge:
> Great catch @wondenge, thanks for the contribution!
> 
> I'm going to add a little bit (probably a test) and then get this merged.

Thanks @vados-cosmonic , happy to help.

PR #1378, opened 2026-04-12
Fix getGlobalCurrentTaskMeta null handling

wondenge:
## Description

`_getGlobalCurrentTaskMeta` treats a cleared task slot (`null`) as a valid meta object. `_clearCurrentTask` sets `CURRENT_TASK_META[componentIdx] = null`, but the getter only checks for `undefined`, so `null` passes through and `{...null}` returns `{}` truthy, but with no `taskID`. This causes `_lowerImportBackwardsCompat` to skip its auto-task-creation fallback and throw `invalid/missing async task meta`.

This shows up when an import fires between two export calls. Export trampolines for types that need allocation (strings, lists, etc.) call `realloc` via `_utf8AllocateAndEncode` before creating their task. If the component has imports wired into internal functions (allocator internals like `dlmalloc::memalign`), those imports go through `_lowerImportBackwardsCompat` in a window where `CURRENT_TASK_META` is `null` from the previous export's cleanup but no new task exists yet.

```
Error: invalid/missing async task meta
    at _lowerImportBackwardsCompat (component.js)
    at component.wasm._ZN8dlmalloc8dlmalloc17Dlmalloc$LT$A$GT$8memalign17h4d574c3a3414c418E
```

The fix adds a `null` check alongside the existing `undefined` check in the intrinsic codegen. A codegen test is added to `initialization.js` that transpiles the `non-wizered-init` component and verifies the generated `_getGlobalCurrentTaskMeta` includes the null check. This is a string match on generated code rather than a behavioral test, a proper integration test would need a component with imports wired into allocator internals to actually trigger the crash, which is hard to set up generically. The codegen assertion is a pragmatic alternative that catches accidental removal of the null check. I'll file a [follow-up issue](https://github.com/bytecodealliance/jco/issues/1379) for a behavioral test that exercises the actual failure path.

### DIFF
--- a/crates/js-component-bindgen/src/intrinsics/mod.rs
+++ b/crates/js-component-bindgen/src/intrinsics/mod.rs
@@ -884,7 +884,7 @@ impl Intrinsic {
                     r#"
                       function {get_current_global_task_meta_fn}(componentIdx) {{
                           const v = {global_current_task_meta_obj}[componentIdx];
-                          if (v === undefined) {{ return v; }}
+                          if (v === undefined || v === null) {{ return undefined; }}
                           return {{ ...v }};
                       }}
                     "#,


packages/jco/test/initialization.js
```diff
@@ -40,6 +42,29 @@ suite("non-wizered raw initialize export", () => {
...
+    // export's task is created).
+    test("_getGlobalCurrentTaskMeta handles null after task clear", async () => {
+        const name = "non-wizered-init";
+        const componentPath = join(LOCAL_TEST_COMPONENTS_DIR, `${name}.wasm`);
+        const { files } = await transpile(await readFile(componentPath));
+
+        const jsFile = Object.keys(files).find((f) => f.endsWith(".js") && !f.includes(".d."));
+        assert.ok(jsFile, "transpile should produce a JS file");
+
+        const source = Buffer.from(files[jsFile]).toString("utf8");
```
vados-cosmonic: We can use the actual name of the expected file in here, as it's consistently easy to predict.

packages/jco/test/initialization.js
```diff
@@ -40,6 +42,29 @@ suite("non-wizered raw initialize export", () => {
...
+        assert.include(
+            source,
+            "v === null",
+            "_getGlobalCurrentTaskMeta must check for null (set by _clearCurrentTask), not just undefined"
```
vados-cosmonic: This is not a great check -- if we're going to look for the check this could match anywhere.

You're gonna have to prompt a bit harder here! :)

packages/jco/test/initialization.js
```diff
@@ -40,6 +42,29 @@ suite("non-wizered raw initialize export", () => {
         await cleanup();
     });
 
+    // _clearCurrentTask sets CURRENT_TASK_META[componentIdx] = null, so
+    // _getGlobalCurrentTaskMeta must treat null the same as undefined.
+    // Without this, {...null} produces {} (truthy, no taskID), which
+    // skips the auto-task-creation fallback in _lowerImportBackwardsCompat
+    // and throws "invalid/missing async task meta" when an import fires
+    // between two export calls (e.g. during realloc before the next
+    // export's task is created).
```
vados-cosmonic:
```suggestion
    // Ensure `_getGlobalCurrentTaskMeta` consistently checks for missing tasks
    // see: https://github.com/bytecodealliance/jco/pull/1378
```

We can be more concise here I think.

vados-cosmonic (commented):
Thanks for the contribution @wondenge -- I'd like to get the behavioral test in here before this merges -- the generated-code-only check doesn't really do much to help IMO, and regardless it would be better to check in the regression test with the fix.

packages/jco/test/initialization.js
vados-cosmonic: I removed the changes here manually

packages/jco/test/initialization.js
vados-cosmonic: I removed the changes here manually

packages/jco/test/initialization.js
vados-cosmonic: I removed the changes here manually

vados-cosmonic (approved):

wondenge:
> Thanks for the contribution @wondenge -- I'd like to get the behavioral test in here before this merges -- the generated-code-only check doesn't really do much to help IMO, and regardless it would be better to check in the regression test with the fix.

Hey @vados-cosmonic sorry for the delay, been swamped this week. Thanks for cleaning things up in the meantime. I'd still like to add a behavioral test to this PR before it merges if that works for you -  a test component in `crates/test-components` that hits the actual failure path rather than the string match. That way the fix and its regression test go in together.

vados-cosmonic:
Hey @wondenge I really appreciate it, please take your time!

BTW, Jco 1.17.7 is on it's way out the door ([this link](https://www.npmjs.com/package/@bytecodealliance/jco/v/1.17.7) should work soon) -- so feel free to give that a try as well. The first implementations of `future<t>` coming *out* of a function are now in place (being able to send *in* a `future<t>` will be in soon as well).

vados-cosmonic:
Hey @wondenge do you still have time to work on this? I wonder if this bug is still relevant actually as we've had quite a few changes to the code in the meantime -- the branch still looks good so it may be time to merge this and add an issue to fill in the test case later?

vados-cosmonic (approved):

PR #1382, opened 2026-04-13
feat(p3-shim): back StreamReader with async iterator

andreiltd:
Replace ReadableStream internals with async iterator.

### DIFF
--- a/packages/preview3-shim/lib/nodejs/http/client.js
+++ b/packages/preview3-shim/lib/nodejs/http/client.js
@@ -1,5 +1,5 @@
 import { ResourceWorker } from "../workers/resource-worker.js";
-import { StreamReader } from "../stream.js";
+import { StreamReader, readableStreamFromIterator } from "../stream.js";
 import { FutureReader, future } from "../future.js";
 import { _fieldsFromEntriesChecked } from "./fields.js";
 import { HttpError } from "./error.js";
@@ -51,7 +51,7 @@ export const client = {
     const { port1: tx, port2: rx } = new MessageChannel();
 
     const transfer = [rx];
-    const stream = body?.intoReadableStream();
+    const stream = body ? readableStreamFromIterator(body.intoAsyncIterator()) : undefined;
     if (stream) {
       transfer.unshift(stream);
     }
--- a/packages/preview3-shim/lib/nodejs/stream.js
+++ b/packages/preview3-shim/lib/nodejs/stream.js
@@ -1,4 +1,29 @@
 /**
+ * Creates a transferable ReadableStream from an async iterator.
+ *
+ * Use this when a StreamReader's contents need to be sent to a worker thread,
+ * since async iterators are not transferable.
+ *
+ * @param {AsyncIterator} iterator - The async iterator to wrap.
+ * @returns {ReadableStream} A transferable ReadableStream that pulls from the iterator.
+ */
+export function readableStreamFromIterator(iterator) {
+  return new ReadableStream({
+    async pull(controller) {
+      const { done, value } = await iterator.next();
+      if (done) {
+        controller.close();
+      } else {
+        controller.enqueue(value);
+      }
+    },
+    cancel() {
+      return iterator.return?.();
+    },
+  });
+}
+
+/**
  * Creates a bidirectional stream with separate reader and writer interfaces.
  *
  * Uses a TransformStream internally.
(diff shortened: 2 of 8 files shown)


packages/preview3-shim/lib/nodejs/http/client.js
```diff
@@ -51,7 +51,9 @@ export const client = {
     const { port1: tx, port2: rx } = new MessageChannel();
 
     const transfer = [rx];
-    const stream = body?.intoReadableStream();
+    const stream = body
+      ? readableStreamFromIterator(body.intoAsyncIterator())
+      : undefined;
```
vados-cosmonic: In what kinds of cases does a fallback of `undefined` make sense? Is the worker OK with getting an `undefined` body versus an empty one?

packages/preview3-shim/lib/nodejs/stream.js
```diff
@@ -42,25 +67,33 @@ export function stream(opts = {}) {
...
    *
-   * @param {ReadableStream} readable - The readable stream to consume.
-   * @throws {Error} If the provided stream does not implement `getReader`.
+   * @param {AsyncIterable} source - An async iterable to consume
+   * @throws {Error} If the provided source does not implement `[Symbol.asyncIterator]`.
    */
-  constructor(readable) {
-    if (!readable || typeof readable.getReader !== "function") {
+  constructor(source) {
+    if (!source || typeof source[Symbol.asyncIterator] !== "function") {
```
vados-cosmonic: Would you mind adding support for regular iterators here? I'm not sure we talked about this *but* I'd like to be able to support those simply as well (the code is in place in the host already)

packages/preview3-shim/lib/nodejs/stream.js
```diff
@@ -42,25 +67,33 @@ export function stream(opts = {}) {
...
+    this.#source = source;
+    // For ReadableStream, use preventCancel so the underlying stream is
+    // not cancelled when the iterator is released.
+    if (source instanceof ReadableStream) {
+      this.#iterator = source[Symbol.asyncIterator]({
+        preventCancel: true,
+      });
+    } else {
+      this.#iterator = source[Symbol.asyncIterator]();
+    }
```
vados-cosmonic, line 113: we could probably push this if clause into the actual `preventCancel` setting.

Also the async iterator function takes no arguments IIRC -- for `ReadableStream` we need `stream.values`, no?

packages/preview3-shim/lib/nodejs/stream.js
```diff
@@ -69,9 +102,9 @@ export class StreamReader {
    * @returns {Promise<any>} Resolves with the next chunk or null if the stream is done.
    */
   async read() {
-    this.#ensureReader();
+    this.#ensureIterator();
 
-    const { done, value } = await this.#reader.read();
+    const { done, value } = await this.#iterator.next();
```
vados-cosmonic, line 127: Note that it is possible to get `done: true` *and* a terminal value -- I think this doesn't account for that case.

packages/preview3-shim/lib/nodejs/stream.js
```diff
@@ -107,37 +140,46 @@ export class StreamReader {
    * @returns {Promise<undefined>}
    */
   async cancel(reason) {
-    this.#ensureReader();
-    return this.#reader.cancel(reason);
+    this.#ensureIterator();
+    if (typeof this.#iterator.return === "function") {
+      return this.#iterator.return(reason);
```
vados-cosmonic: In this case I think the type doc above is now wrong -- since we can now *sometimes* return the `IteratorResult`...

Also, not sure why return takes an arg here... do we end up using that?

packages/preview3-shim/lib/nodejs/http/client.js
andreiltd: It's just the next line prevents undefined from being added to transfer list: `if (stream)`

packages/preview3-shim/lib/nodejs/stream.js
andreiltd: The reason that we pass to the `cancel` is unused otherwise, alternatively we can just get rid of reason arg form the method?

packages/preview3-shim/lib/nodejs/http/client.js
vados-cosmonic: Oh yeah I saw that, but it actually does get passed through if it's undefined at the bottom of there right -- just wanted to make sure the worker is doing similar checks/prepared for `undefined` to come thorugh. 

I just checked though in the https://github.com/bytecodealliance/jco/blob/main/packages/preview3-shim/lib/nodejs/workers/http-worker.js and it looks just fine in there!

packages/preview3-shim/lib/nodejs/stream.js
vados-cosmonic: If we want to expose the reason for cancellation maybe we can store that somewhere else and add a custom API for getting at it? This is another one of those reasons why I think we are right in specifying the interfaces the stream (and later future) objects will adhere to rather than specifically using builtins. 

That said, I'm also OK with adding this when we need it (and letting it go unused for now)!

packages/preview3-shim/lib/nodejs/stream.js
andreiltd: I've added sync iterator as well - internally I just transform it into async iterator.

vados-cosmonic (approved):
LGTM 🚀 

Let's get this in quickly and fixup any issues when we (hopefully soon) are able to test E2E w/ the actual impl.

PR #1388, opened 2026-04-14
chore(deps): update deps

vados-cosmonic:


vados-cosmonic:
OK so the problem here is actually quite important -- `wasi:http@0.2.3` is no longer usable from `componentize-js` 0.20.0 -- this means that we need to figure out a way to not break people... *or* release a breaking change version of Jco. 

Right now `componentize-js` includes all binaries that are necessary (debug/release/weval) and that means componentize-js is ~40MB installed. keeping two versions around is going add another 40MB to Jco.

image

Another option here is to start downloading and setting up required componentize-js setups on the fly, or requiring a separate install of componentize-js (which will be dynamically imported). This is less than ideal, as it would require network access or more friction for users...  And be breaking change on top of all that.

PR #1395, opened 2026-04-15
fix(p2-shim): ns mangnitude used in fs while building datetimes

vados-cosmonic:


vados-cosmonic:
This will be rolled into #1397

PR #1410, opened 2026-04-17
Update `use jco` directive to `use components`

alexcrichton:
This is a small change to the `use jco` directive intended to be used for telemetry to make this less tool-specific and instead leave open the opportunity for other projects to use this as well which may not be using jco specifically.

### DIFF
--- a/crates/js-component-bindgen/src/transpile_bindgen.rs
+++ b/crates/js-component-bindgen/src/transpile_bindgen.rs
@@ -416,7 +416,7 @@ impl JsBindgen<'_> {
         }
 
         // Render the telemery directive
-        uwriteln!(output, r#""use jco";"#);
+        uwriteln!(output, r#""use components";"#);
 
         let js_intrinsics = render_intrinsics(RenderIntrinsicsArgs {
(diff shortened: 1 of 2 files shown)


alexcrichton:
I'll note as well that this is under the assumption that this telemetry isn't actually hooked up anywhere yet. If it is, this'll be a bit more difficult to change!

cc @lukewagner on this as well

vados-cosmonic (approved):
LGTM 🚀 

As far as this possibly being elsewhere (and in use), I'm not against just having both :)

vados-cosmonic:
We're quite strict on commit titles, so to save you some time, the below message should work (mostly `refactor(bindgen)` and ensuring not to use a capital case letter after `: `):

```
refactor(bindgen): update 'use jco' directive to 'use components'
```

As silly as it sounds, it would be ideal if we split up the change to `test/jco` into *another* commit so it's picked up cleanly by the automation:

```
test(jco): update directive test to check for 'use components'
```

lukewagner:
Seems reasonable to me too

alexcrichton:
Thanks @vados-cosmonic!

PR #1421, opened 2026-04-21
fix(jco): `u64/s64` parameters and returns in `--js` mode

### LINKED ISSUE #1393: `jco transpile --js` output unusable for `u64`/`s64` params/returns
The `--js` flag of `transpile` runs the core Wasm modules of the component through `wasm2js`, translating Wasm to the asm.js subset of JS.

asm.js does not support 64-bit integers, so such values and operations are split into 32-bit ones by `wasm2js`. For instance:
```wat
(func (;0;) (type 0) (param i64 i64) (result i64)
  local.get 0
  local.get 1
  i64.add
)
```
becomes
```js
var setTempRet0 = env.setTempRet0;
var i64toi32_i32$HIGH_BITS = 0;
function legalstub$0($0, $1, $2, $3) {
  $1 = $1 + $3 | 0;
  $0 = $0 + $2 | 0;
  i64toi32_i32$HIGH_BITS = $0 >>> 0 < $2 >>> 0 ? $1 + 1 | 0 : $1;
  setTempRet0(i64toi32_i32$HIGH_BITS | 0);
  return $0;
}
```

-----------

The problem is that this interface (i64 split into i32) no longer matches WebAssembly. Wasm has used BigInt for i64 in JS integration for many years now.

This is evident in the compoenent-level wrapper generated around it to call the export from other JS code (some parts omitted for brevity):
```js
function add(arg0, arg1) {
      ...
      const hostProvided = false;
      
      const [task, _wasm_call_currentTaskID] = ...
      
      const started = task.enterSync();
      let ret =   _withGlobalCurrentTaskMeta({
        ...
        fn: () => exports0Add(toUint64(arg0), toUint64(arg1)),
      });
      
      ...
      task.resolve([BigInt.asUintN(64, BigInt(ret))]);
      task.exit();
      return BigInt.asUintN(64, BigInt(ret));
    }
    ({ exports: exports0 } = yield instantiateCore(yield module0));
    exports0Add = exports0.add;
    
    return { add,  };
  })();
```

Currently this means the using `jco transpile --js` on componets using u64 or s64 in the interface (e.g. `export add: func(a: u64, b: u64) -> u64;`) will not produce valid/usable code for such exports.

The wrapper will try to call a export expecting `4 x Number` with `2 x BigInt` (and returns Number instead of BigInt).

QuantumSegfault:
Fixes #1393

Currently, `jco` is oblivious to the differences between the output of `wasm2js` and proper WebAssembly. Modern WebAssembly has `BigInt` integration to allow passing and receiving them where core Wasm expects an `i64`. `jco` is built around this assumption, but the `asm.js` output from `wasm2js` created with the `--js` option does not follow it. It uses the old scheme of splitting i64 => (i32, i32)

This PR papers over this difference by merging and splitting `BigInt` to/from pairs of i32 at the component <-> core boundary. Component imports or exports expecting `s64` or `u64` still take `BigInt` regardless of `--js` vs not, but it is translated before hitting the core "Wasm" (asmjs) function.


QuantumSegfault:
@vados-cosmonic 

Requesting review

QuantumSegfault:
Oh! It did it automatically (request review). Cool.

-----

Also, while not strictly related, I also added a fix to the TypeScript bindings I found while doing this (`--js` instantiate has different signature).

vados-cosmonic:
Hey @QuantumSegfault thanks for working on this! The CI is going to ding you on commit titles I think, but commits with titles like:

```
fix(bindgen): the thing you fixed in packages/js-component-bindgen{,component}
fix(transpile): the thing you fixed in packages/jco-transpile
```

Should pass CI

QuantumSegfault:
Okay. I squashed the commits (and fixed the format issues) and renamed it. Choosing just one category was difficult... most of the changes are in "bindgen", even though its really a change in the results of "transpile".

I just went with `jco`.

vados-cosmonic:
Ah so *ideally* we'd separate the commits to the different parts of the codebase, but all good -- since bindgen gets rolled up into Jco this is certainly fine :)

Should be able to get his reviewed and merged in soon!

crates/js-component-bindgen/src/intrinsics/conversion.rs
```diff
@@ -379,6 +386,13 @@ impl ConversionIntrinsic {
                 ));
             }
 
+            Self::SplitBigInt64 => output.push_str("
+                const splitInt64 = i => [Number(i & 0xFFFFFFFFn)|0, Number((i >> 32n) & 0xFFFFFFFFn)|0];
+            "),
```
vados-cosmonic: Small thing but let's use `self.name()` here rather than hard-coding `splitInt64`, *just* in case we ever do. 

This file doesn't do it consistently but it's something we *should* be doing everywhere

crates/js-component-bindgen-component/wit/js-component-bindgen.wit
```diff
@@ -72,6 +72,8 @@ world js-component-bindgen {
 
     /// Configure whether to generate code that includes strict type checks
     strict: option<bool>,
+    /// Whether the core module(s) were actually transpiled from Wasm to JS (asm.js) and thus needs shimming for i64
+    asmjs: bool,
```
vados-cosmonic: The reason most of those other ones are `option<t>`s is to avoid forcing it to be specified IIRC and to make clear the case where the user didn't choose one way or the other -- any reason we broke from this pattern here?

Maybe I'm wrong here and we wont' see other tests fail where `asmjs` was not specified, but that's what I would expect (just because of how `option<bool>` vs `bool` is transpiled by Jco)

crates/js-component-bindgen/src/transpile_bindgen.rs
```diff
@@ -2780,6 +2787,39 @@ impl<'a> Instantiator<'a, '_> {
             assert!(prev.is_none());
         }
 
+        if self.bindgen.opts.asmjs {
+            let component_instance_idx = instance.unwrap().as_u32();
```
vados-cosmonic:
```suggestion
            let component_instance_idx = instance.expect("missing runtime component index during static module instantiation").as_u32();
```

crates/js-component-bindgen/src/ts_bindgen.rs
```diff
@@ -360,48 +360,74 @@ pub fn ts_bindgen(
...
-                        instantiateCore?: (module: WebAssembly.Module, imports: Record<string, any>) => WebAssembly.Instance
-                    ): {camel};
-                    export function instantiate(
-                        getCoreModule: (path: string) => WebAssembly.Module | Promise<WebAssembly.Module>,
-                        imports: ImportObject,
-                        instantiateCore?: (module: WebAssembly.Module, imports: Record<string, any>) => WebAssembly.Instance | Promise<WebAssembly.Instance>
-                    ): {camel} | Promise<{camel}>;
+                /**
+                 * Instantiates this component with the provided imports and
+                 * returns a map of all the exports of the component.
```
vados-cosmonic, line 370: What do you think about adding a note here about this being the version that is used only for asmjs?

crates/js-component-bindgen/src/ts_bindgen.rs
```diff
@@ -360,48 +360,74 @@ pub fn ts_bindgen(
...
+                 *
+                 * This function is intended to be similar to the
+                 * `WebAssembly.instantiate` function. The `imports`
+                 * argument is the \"import object\" for wasm, except here it
+                 * uses component-model-layer types instead of core wasm
+                 * integers/numbers/etc.
+                 */
+                export function instantiate(
+                    imports: ImportObject,
```
vados-cosmonic, line 380: I think this bordering on a breaking change (though the previous state is not ideal)?

What do you think of either of the following:

- Leaving the previous structure in place (and ignoring the unneeded args as necessary)
- Introducing a new `instantiateAsmJS` or something similar and threading that through

Note that in general we're also going away from the [`WebAssembly.instantiate()` signature](https://developer.mozilla.org/en-US/docs/WebAssembly/Reference/JavaScript_interface/instantiate_static). IMO we actually don't need to follow that signature (I prefer passing arg objects even with the overhead), but that is a thing we should change at a breaking change point

vados-cosmonic (commented):
This is *almost* perfect -- I had a few nits and some questions.

Also, you should be able to fmt the whole codebase by running `npm run fmt` from the top level (or `cargo fmt`, `cargo clippy`, etc).

Thanks again for fixing this (and adding a test!)

crates/js-component-bindgen/src/ts_bindgen.rs
QuantumSegfault: Well, that's an odd situation. Technically its a breaking change in the generated TypeScript types, but its done to reflect the state the JS is actually in.

I think there's an argument to be made that from the outside, one should not be able to distinguish `--js` and normal Wasm output of `jco`. `--js` can be used, and seamlessly upgraded to Wasm, or code that would build upon `jco` can provide 100% JS as an easy change, without refactoring much, if any code. Just plug and play.

So I think that between the two, it makes more sense to make the `--js` instantiate have the same (JS) signature as the normal one.

crates/js-component-bindgen-component/wit/js-component-bindgen.wit
QuantumSegfault: No particular reason. Just an oversight on my part. The `asmjs` option here has the requirements as any other option/argument to `generate`

crates/js-component-bindgen/src/ts_bindgen.rs
QuantumSegfault: Let me know which approach you want to go with and I'll make the changes. But so you get an idea:

```js
export function instantiate(imports) {
  const wasm_file_to_asm_index = {
    'adder.component.core.wasm': 0
  };

  return _instantiate(
      module_name => wasm_file_to_asm_index[module_name],
      imports,
      (module_index, imports) => ({ exports: asmInit[module_index](imports) })
  );
}
```
becomes
```js
export function instantiate(getCoreModule, imports, instantiateCore = undefined) {
    // ... exact same body
}
```

vados-cosmonic: Yeah considering this is a fix I'm a bit conflicted... 

How about the second one? I know it's a bit annoying, but maybe you could do something like add warnings or even throw an error if a user attempts to supply the other arguments?

Separately, normally I'm generally not a fan of function calls that take arguments and try to differentiate at runtime which case you're in, but maybe that pattern is worth doing here as a way to have our cake and eat it too.

So basically I'm thinking something like this:

```js
// (assuming this is the instantiate built for asmjs use only)
export function instantiate(getCoreModule, imports, instantiateCore = undefined) {
    if (isImportsObj(getModuleCore)) {
        imports = getModuleCore;
    } else if (getCoreModule !== undefined) {
        throw new Error('getCoreModule must not be specified when performing instantiation with wasm2js, only imports should be supplied');
    }

    if (!isImportsObj(imports)) {
        throw new Error('invalid import object during instantiation');
    }

   // Not sure if we really need to check this too but might be helpful in failing fast/early w/ bad config
   if (instantiatetCore !== undefined) { ... }
}
```

Happy to go with a warning (debug log) or something less harsh than an error, or whatever you think would make sense.

With the changes above people could use it that more natural way (with no placeholder args) *and* the other way that matches the normal instantiate a bit more. 

At this point we're like one step away from bikeshedding so happy to go with what you decide here and get this in quickly.

QuantumSegfault: Hmmm...

In my mind that's a no-go as it does not conform to the TypeScript definition. Then again...those using TypeScript will use the full signature, while those using unchecked raw JS won't have their existing code broken.

But also, thinking about it, doing this does not help with seamlessness. Those using trying to just enable `--js` on their existing code will get errors cause of `getCoreModule`. Then again, perhaps that's a non-goal since the kinds of environments you'd be inclined to use `--js` for are the kinds that don't have WebAssembly support anyway, so they'd have to give a bogus/modified `getCoreModule` anyway, no?

QuantumSegfault: In that respect maybe `#2` is better, because you could in theory do runtime checks for which exists (instantiateAsmJS vs normal instantiate) and call the correct one, in programs that support both interchangeably.

------

It all sucks in its own way...

If we are trying to avoid causing a "breaking change"...lets think of it this way. Those who might be successfully using `--js` mode will be using the proper JS signature with one arg, regardless of what TypeScript says. Those relying on the TypeScript bindings can't use `--js` cause it will mismatch and cause problems (TypeScript says give 2 args, but JS will error out cause you gave 2 to a function wanting only 1).

All that considered, it feels that the LEAST breaking change is to modify the TS bindings, like I already have. Then we can reconsider unifying (or further dividing) the API later?

vados-cosmonic: > In my mind that's a no-go as it does not conform to the TypeScript definition. Then again...those using TypeScript will use the full signature, while those using unchecked raw JS won't have their existing code broken.

So this is a good point but I think we *could* change the TS definition to capture this and actually maintain backwards comaptibility. That said the amount we've had to discuss it probably means it's just not worth the effort.

> But also, thinking about it, doing this does not help with seamlessness. Those using trying to just enable --js on their existing code will get errors cause of getCoreModule. Then again, perhaps that's a non-goal since the kinds of environments you'd be inclined to use --js for are the kinds that don't have WebAssembly support anyway, so they'd have to give a bogus/modified getCoreModule anyway, no?

Yeah, I think whether it's a good idea to fail loudly here is debatable, but the user does legitimately put themselves in a different situation when they chose to enable `--js`. Of course, it would take just a few seconds of debugging to figure out that the loader you passed isn't getting used/etc, but maybe the explicit error is worth it. 

*Ah* you know what, this is a perfect use for the [recently introduced `--strict` flag](https://github.com/bytecodealliance/jco/commit/68c793ac7627ed7c206f9fc36e5a6072dac14591). At the cost of over-engineering/unneeded complexity, we could throw the error *only* if the `--strict` flag was passed, and otherwise print a debug log message.

vados-cosmonic: >  Those who might be successfully using --js mode will be using the proper JS signature with one arg, regardless of what TypeScript says. Those relying on the TypeScript bindings can't use --js cause it will mismatch and cause problems (TypeScript says give 2 args, but JS will error out cause you gave 2 to a function wanting only 1).
>
> All that considered, it feels that the LEAST breaking change is to modify the TS bindings, like I already have. Then we can reconsider unifying (or further dividing) the API later?

Yeah with the reasoning laid out, I agree -- let's go with what you have to modify the bindings and call it a day! I think we can revisit this in the future when it's time for breaking changes.

vados-cosmonic: (resolving so it doesn't block merging -- I think there's only one comment left)

crates/js-component-bindgen/src/ts_bindgen.rs
QuantumSegfault: In that case, you want this done?

QuantumSegfault: Okay, I tweaked it a bit to explicitly mention `wasm2js`. Is that better?

vados-cosmonic: Hey yeah that looks great, thanks!

vados-cosmonic (approved):
LGTM 🚀

PR #1422, opened 2026-04-22
feat(jco): add future lower tests

vados-cosmonic:


vados-cosmonic:
NOTE: this group of changes is actually *incomplete* -- there are a few issues remaining (leaving aside the reported bugs from earlier that are also in the issue tracker):

https://github.com/bytecodealliance/jco/issues/1428
https://github.com/bytecodealliance/jco/issues/1430

This group of changes *does* introduce working future lowering for the rest of the tested types.

PR #1433, opened 2026-04-29
fix(bindgen): use waitable index in p3 future event payload

andreiltd:
The producer side of `future.read/future.write` is emitting `componentIdx` as the event's `payload0`, but that value identifies the component instance and not the specific future so a wasm task waiting on multiple futures couldn't tell which one
resolved.


vados-cosmonic (approved):
LGTM 🚀 

Thanks for the fix!

PR #1434, opened 2026-04-29
fix(bindgen): return state-only handle when async-lowered import resolves eagerly

andreiltd:
Per CanonicalABI.md an async-lowered import whose callee resolves synchronously returns `Subtask.State.RETURNED` state only and no subtask handle is packed into the result.

Separateley The CanonicalABI.md packs state into 4 bits so the check bounds are adjusted.

This the relevant spec section:

> In the async case, if the callee already called on_resolve, then the RETURNED code is eagerly returned to the core wasm caller without needing to add a Subtask to the current component instance's handles table.

```python
    if subtask.resolved():
      assert(flat_results == [])
      subtask.deliver_resolve()
      return [Subtask.State.RETURNED]
    else:
      subtaski = thread.task.inst.handles.add(subtask)
      def on_progress():
        def subtask_event():
          if subtask.resolved():
            subtask.deliver_resolve()
          return (EventCode.SUBTASK, subtaski, subtask.state)
        subtask.set_pending_event(subtask_event)
      assert(0 < subtaski <= Table.MAX_LENGTH < 2**28)
      assert(0 <= subtask.state < 2**4)
      return [subtask.state | (subtaski << 4)]
```

### DIFF
--- a/crates/js-component-bindgen/src/intrinsics/p3/async_task.rs
+++ b/crates/js-component-bindgen/src/intrinsics/p3/async_task.rs
@@ -2163,11 +2163,6 @@ impl AsyncTaskIntrinsic {
 
                         // NOTE: at this point we know that we are working with an async lowered import
 
-                        const subtaskState = subtask.getStateNumber();
-                        if (subtaskState < 0 || subtaskState > 2**5) {{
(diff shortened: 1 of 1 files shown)


vados-cosmonic (approved):
LGTM 🚀

PR #1435, opened 2026-04-29
fix(p3-shim): adapt shim to latest bindings

andreiltd:
This is a batch of small adjustments for p3 shim to work with the latest bindgen.

### DIFF
--- a/packages/preview3-shim/lib/nodejs/filesystem/descriptor.js
+++ b/packages/preview3-shim/lib/nodejs/filesystem/descriptor.js
@@ -125,14 +125,14 @@ class Descriptor {
    * ```
    *
    * @async
-   * @param {object} data A data source implementing `intoAsyncIterator()`.
+   * @param {object} data A data source implementing `[Symbol.asyncIterator]()`.
    * @param {bigint} offset The offset within the file.
    * @returns {Promise<void>}
    * @throws {FSError} `payload.tag` contains mapped WASI error code.
    */
   async writeViaStream(data, offset) {
     this.#ensureHandle();
-    const stream = readableStreamFromIterator(data.intoAsyncIterator());
+    const stream = readableStreamFromIterator(data[Symbol.asyncIterator]());
 
     try {
       await worker().run({ op: "write", fd: this.#handle.fd, offset, stream }, [stream]);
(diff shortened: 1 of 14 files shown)


packages/preview3-shim/lib/nodejs/filesystem/descriptor.js
```diff
@@ -125,14 +125,14 @@ class Descriptor {
...
-   * @param {object} data A data source implementing `intoAsyncIterator()`.
+   * @param {object} data A data source implementing `[Symbol.asyncIterator]()`.
    * @param {bigint} offset The offset within the file.
    * @returns {Promise<void>}
    * @throws {FSError} `payload.tag` contains mapped WASI error code.
    */
   async writeViaStream(data, offset) {
     this.#ensureHandle();
-    const stream = readableStreamFromIterator(data.intoAsyncIterator());
+    const stream = readableStreamFromIterator(data[Symbol.asyncIterator]());
```
vados-cosmonic, line 135: IMO would be nice to check for this/ensure that what we got had the async iterator and throw a reasonable `TypeError` or something?

(same for the other spots, if we do it, wonder if it's worth factoring out into it's own logic -- something like `ensureAsyncIterable(arg)`? Maybe not)

vados-cosmonic (approved):
LGTM 🚀 

Left a single nit, and fmt isn't passing yet, but this should be good to go, feel free to merge!

packages/preview3-shim/lib/nodejs/sockets/tcp.js
```diff
@@ -239,21 +239,22 @@ export class TcpSocket {
...
-   * @throws {SocketError} With payload.tag 'invalid-argument' if ~data~ is not a {StreamReader}
+   * @throws {SocketError} With payload.tag 'invalid-argument' if `data` does not implement [Symbol.asyncIterator]
    * @throws {SocketError} for other errors, payload.tag maps the system error
    */
   async send(data) {
     if (this.#state !== STATE.CONNECTED) {
       throw new SocketError("invalid-state");
     }
-    if (!(data instanceof StreamReader)) {
+    if (data == null || typeof data[Symbol.asyncIterator] !== "function") {
```
vados-cosmonic: also `undefined` could slip through here I think

PR #1437, opened 2026-04-29
fix(bindgen): avoid async future and stream host injection deadlocks

andreiltd:
The spec is subtle here but IIUC we should start host side future and stream injection without awaiting it before the canonical read operation can see `BLOCKED`.

So if the copy does not immediately produce a pending event, the implementation must set `ASYNC_COPYING` and return `BLOCKED`. Our old code awaited the host promise during host injection before allowing this path to return `BLOCKED`, which deadlocks the guest scheduler.

This is consistent with python code from canon ABI spec:

```python
  if not e.has_pending_event():
    if not opts.async_:
      e.state = CopyState.SYNC_COPYING
      thread.wait_until(e.has_pending_event)
    else:
      e.state = CopyState.ASYNC_COPYING
      return [BLOCKED]
  code,index,payload = e.get_pending_event()
  assert(code == event_code and index == i)
  return [payload]
```

This is the same for futures and streams. An example of deadlock on the guest side with an old implementation looks like this:

```rust
let (mut tx, rx) = wit_stream::new();
futures::join!(
    async {
        wasi::cli::stdout::write_via_stream(rx).await.unwrap();
    },
    async {
        tx.write(b"hello, world\n".to_vec()).await;
        drop(tx);
    },
);
```

I've added concurrency tests for both streams and futures that this patch fixes.

### DIFF
--- a/crates/js-component-bindgen/src/intrinsics/p3/async_future.rs
+++ b/crates/js-component-bindgen/src/intrinsics/p3/async_future.rs
@@ -673,8 +673,9 @@ impl AsyncFutureIntrinsic {
 
 
                                   // Before performing this read, if we're dealing with a host-controlled
-                                  // future, then we should inject a write, but we can't wait for it to complete
-                                  // as we must do the rendesvous read below for the write to complete.
+                                  // future, start injecting the write. The injection may depend on sibling
+                                  // guest work running, so cleanup is attached without awaiting here; the
+                                  // canonical read must be able to return BLOCKED first.
                                   let injectedWritePromise;
                                   if (this.#hostInjectFn) {{
                                       injectedWritePromise = this.#hostInjectFn({{ count: 1 }});
@@ -687,15 +688,10 @@ impl AsyncFutureIntrinsic {
                                   }});
 
                                   if (injectedWritePromise) {{
-                                      if (this.hasPendingEvent()) {{
-                                          const cleanupFn = await injectedWritePromise;
-                                          cleanupFn();
-                                      }} else {{
-                                          injectedWritePromise.then(
-                                              cleanupFn => cleanupFn(),
-                                              err => this.setPendingEvent(() => {{ throw err; }}),
-                                          );
-                                      }}
+                                      injectedWritePromise.then(
+                                          cleanupFn => cleanupFn(),
+                                          err => this.setPendingEvent(() => {{ throw err; }}),
+                                      );
                                   }}
 
                                   return {{ buffer }};
--- a/crates/js-component-bindgen/src/intrinsics/p3/async_future.rs
+++ b/crates/js-component-bindgen/src/intrinsics/p3/async_future.rs
@@ -687,8 +687,15 @@ impl AsyncFutureIntrinsic {
                                   }});
 
                                   if (injectedWritePromise) {{
-                                      const cleanupFn = await injectedWritePromise;
-                                      cleanupFn();
+                                      if (this.hasPendingEvent()) {{
+                                          const cleanupFn = await injectedWritePromise;
+                                          cleanupFn();
+                                      }} else {{
+                                          injectedWritePromise.then(
+                                              cleanupFn => cleanupFn(),
+                                              err => this.setPendingEvent(() => {{ throw err; }}),
+                                          );
+                                      }}
                                   }}
 
                                   return {{ buffer }};
@@ -807,13 +814,6 @@ impl AsyncFutureIntrinsic {
                                            }});
                                        }});
 
-                                       // Perform another write, reusing the buffer
-                                       const {{ buffer }} = await this.guestRead({{
-                                           buffer,
-                                           stringEncoding,
-                                           isAsync: true,
-                                       }});
-
                                        if (!this.hasPendingEvent()) {{
                                            throw new Error("missing pending event after blocked future read");
                                        }}
@@ -860,13 +860,6 @@ impl AsyncFutureIntrinsic {
                                                clearInterval(waitInterval);
                                                resolve();
                                            }});
-                                       }});
-
-                                       // Perform another write, reusing the buffer
-                                       const {{ buffer }} = await this.guestWrite({{
-                                           buffer,
-                                           stringEncoding,
-                                           isAsync: true,
                                        }});
 
                                        if (!this.hasPendingEvent()) {{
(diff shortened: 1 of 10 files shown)


andreiltd:
This requires some more thinking - I will keep it as a draft until I figure out what caused a regression.

andreiltd:
I think the problem is in this code:

https://github.com/bytecodealliance/jco/blob/main/crates/js-component-bindgen/src/intrinsics/p3/async_stream.rs#L912

I'm not sure why we call `copy()` again here instead of getting result  from pending event? I believe the python code counterpart does that:

```python
  e.copy(thread.task.inst, buffer, on_copy, on_copy_done)

  if not e.has_pending_event():
    if not opts.async_:
      e.state = CopyState.SYNC_COPYING
      thread.wait_until(e.has_pending_event)
    else:
      e.state = CopyState.ASYNC_COPYING
      return [BLOCKED]
  code,index,payload = e.get_pending_event()  # <= here, do not copy again but just get result from event
  assert(code == event_code and index == i and payload != BLOCKED)
  return [payload]
```

The way I understand this is that after a blocked copy is awaken, the pending event contains the completed copy and calling `copy()` again will re-enter the stream state machine? The problem is symmetric so it's the same for read and write.

I can try to fix that, but I won't be confident to land it without thorough review 😄

andreiltd:
Last commit seems to fix the regression and with previous changes the deadlock is also fixed.

crates/js-component-bindgen/src/intrinsics/p3/async_future.rs
```diff
@@ -687,8 +687,15 @@ impl AsyncFutureIntrinsic {
                                   }});
 
                                   if (injectedWritePromise) {{
-                                      const cleanupFn = await injectedWritePromise;
-                                      cleanupFn();
+                                      if (this.hasPendingEvent()) {{
```
vados-cosmonic: NIT: what do you think about doing this inside the cleanup function itself?

crates/js-component-bindgen/src/intrinsics/p3/async_stream.rs
```diff
@@ -745,15 +741,22 @@ impl AsyncStreamIntrinsic {
                                  componentIdx,
                              }});
 
+                             let injectedWritePromise;
+                             if (injectHostWrite && !this.hasPendingEvent()) {{
+                                 injectedWritePromise = this.#hostInjectFn({{ count }});
```
vados-cosmonic: NIT: What do you think about moving this inside the host injection function? 

i.e. having it check if there is an event and do nothing?

Definitely don't feel strongly about this, but this feels like it's saying "if there's a pending event, don't write" and the host injection function can be smart enough to essentially turn into a no-op itself, I think.

crates/js-component-bindgen/src/intrinsics/p3/async_stream.rs
```diff
@@ -745,15 +741,22 @@ impl AsyncStreamIntrinsic {
...
                              if (!this.hasPendingEvent()) {{
-                                 if (injectHostWrite) {{
-                                     throw new Error('reader unexpectedly blocked after injected write');
-                                 }}
-
                                  if (isAsync) {{
                                      this.setCopyState({stream_end_class}.CopyState.ASYNC_COPYING);
                                      {debug_log_fn}('[{stream_end_class}#copy()] blocked', {{ componentIdx, eventCode, self: this }});
+                                     if (injectedWritePromise) {{
+                                         injectedWritePromise.then(
```
vados-cosmonic, line 752: Would you mind adding a quick explainer on why we don't wait for the injected write to actually complete here?

And with the previous nits about trying to contain the logic to the injected write promise function itself or the cleanup fn it returns -- would be nice if we could do that here and avoid the chaining here (I *think* this logic is consistent across places we're using the injected write promise, no?)

packages/jco/test/p3/deadlock-regressions.js
```diff
@@ -0,0 +1,103 @@
...
+                            assert.deepEqual(values, [42]);
+                            return 42;
+                        },
+                    },
+                },
+            },
+            jco: {
+                transpile: {
+                    extraArgs: {
+                        asyncExports: ["jco:test-components/local-run-async#run"],
```
vados-cosmonic: You shouldn't need to add this if the func is `async` in the WIT -- did you find that you had to? is there a bug there?

packages/jco/test/p3/deadlock-regressions.js
```diff
@@ -0,0 +1,103 @@
+import { join } from "node:path";
+
+import { suite, test, assert } from "vitest";
+
+import { WASIShim } from "@bytecodealliance/preview2-shim/instantiation";
+
+import { setupAsyncTest } from "../helpers.js";
+import { LOCAL_TEST_COMPONENTS_DIR } from "../common.js";
+
+function deferred() {
```
vados-cosmonic: This seems equivalent to `Promise.withResolvers` ? We use a pony fill in the bindgen code but here you can just use it IMO

vados-cosmonic (approved):
LGTM 🚀 

I left some nits, but in general this looks good to me -- thanks for adding the new tests and fixing so many impl bugs!

crates/js-component-bindgen/src/intrinsics/p3/async_future.rs
andreiltd: That's a nice simplification, yes. The caveat is that we still don't await injected write here because the host promise may depend on concurrent guest work running, so the canonical read must be able to return `BLOCKED` first.

packages/jco/test/p3/deadlock-regressions.js
andreiltd: Oh yeah, definitely a leftover, it works without it.

crates/js-component-bindgen/src/intrinsics/p3/async_stream.rs
andreiltd: This is very similar to future fix, although I wasn't able to prove that with the test. After fixing the future deadlock I just wondered if the faulty pattern exists in the stream implementation and yes it does very similar thing.  If the copy has no pending event, we set `ASYNC_COPYING` and return `BLOCKED`. If we await the injected write we would delay that scheduling point and can block guest executor. I've added a short comment in the code as well.

crates/js-component-bindgen/src/intrinsics/p3/async_stream.rs
andreiltd: I refactored this to do checks in injection function.

PR #1441, opened 2026-05-04
feat(bindgen): add support for stream reads with count > 1

andreiltd:
This adds `readMany(count)` support to generated p3 streams so host-side byte consumers can perform Canonical ABI stream reads instead of pulling one byte at a time. It uses the transferred count from the packed stream result when reading host buffers.

I believe that the current implementation is correct, it's just that iteratrd on single bytes is very inefficient. I originally came across this when tried to run `p3_cli_hello_stdout` test from wasmtime:

```rust
    async fn run() -> Result<(), ()> {
        let (mut tx, rx) = wit_stream::new();
        futures::join!(
            async {
                wasi::cli::stdout::write_via_stream(rx).await.unwrap();
            },
            async {
                tx.write(b"hello, world\n".to_vec()).await;
                drop(tx);
            },
        );
        Ok(())
    }
```

The `write` function is a bit surprising in the sense that it will not loop until the data is sent over, but instead it will try to send the data at most once. In `jco` we will receive a single byte and report `transferred = 1`, which is correct behavior. The looping function is `write_all` which with our current implementation will indeed print the whole string.

I need to have a look if `tx.write` reports warning that we are dropping result that must be used, which would make the debugging much easier.

### DIFF
--- a/packages/preview3-shim/lib/nodejs/stream.js
+++ b/packages/preview3-shim/lib/nodejs/stream.js
@@ -1,3 +1,10 @@
+export const DEFAULT_BYTE_STREAM_CHUNK_SIZE = 64 * 1024;
+
+let BYTE_STREAM_ENCODER = null;
+function encoder() {
+  return (BYTE_STREAM_ENCODER ??= new TextEncoder());
+}
+
 /**
  * Creates a transferable ReadableStream from an async iterator.
  *
@@ -31,8 +38,6 @@ export function readableStreamFromIterator(iterator, name = "iterator") {
     },
   });
 }
-
-export const DEFAULT_BYTE_STREAM_CHUNK_SIZE = 64 * 1024;
 
 /**
  * Creates a transferable ReadableStream from a byte-stream reader.
@@ -119,7 +124,7 @@ function byteChunk(value) {
     return Uint8Array.of(assertByte(value));
   }
   if (typeof value === "string") {
-    return new TextEncoder().encode(value);
+    return encoder().encode(value);
   }
 
   throw new TypeError(
--- a/packages/preview3-shim/lib/nodejs/stream.js
+++ b/packages/preview3-shim/lib/nodejs/stream.js
@@ -32,6 +32,108 @@ export function readableStreamFromIterator(iterator, name = "iterator") {
   });
 }
 
+export const DEFAULT_BYTE_STREAM_CHUNK_SIZE = 64 * 1024;
+
+/**
+ * Creates a transferable ReadableStream from a byte-stream reader.
+ *
+ * Generated p3 stream readers expose `read({ count })`, which performs one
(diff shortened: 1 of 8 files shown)


vados-cosmonic:
Hey @andreiltd should I be reviewing this just yet? You just rebased I think but should I wait for the other PRs to land in `main`?

andreiltd:
Hey @vados-cosmonic  yeah, not yet, this depends (includes): https://github.com/bytecodealliance/jco/pull/1445 

but it looks like 1445 is failing so I will have to figure out that first and then I will rebase this one.

andreiltd:
It's ready now.

packages/preview3-shim/lib/nodejs/stream.js
```diff
@@ -32,6 +32,108 @@ export function readableStreamFromIterator(iterator, name = "iterator") {
...
+        const result = await reader.read({ count: chunkSize });
+        if (isIteratorResult(result)) {
+          return result;
+        }
+        return { value: result, done: result === null };
+      },
+    };
+  }
+
+  if (typeof reader?.[Symbol.asyncIterator] === "function") {
```
vados-cosmonic, line 91: NIT: Can we prefer async iterator here?

packages/preview3-shim/lib/nodejs/stream.js
```diff
@@ -32,6 +32,108 @@ export function readableStreamFromIterator(iterator, name = "iterator") {
...
+    for (const byte of value) {
+      assertByte(byte);
+    }
+    return Uint8Array.from(value);
+  }
+  if (typeof value === "number") {
+    return Uint8Array.of(assertByte(value));
+  }
+  if (typeof value === "string") {
+    return new TextEncoder().encode(value);
```
vados-cosmonic: I assume `new TextEncoder()` is quite optimized at this point, but would be nice to avoid it/reuse an instance

vados-cosmonic (approved):
LGTM 🚀 

Nothing but nits!

packages/preview3-shim/lib/nodejs/stream.js
andreiltd: Do you mean prefer over read() function? This helper prefers `read({ count })` specifically so that it can pass default chunk size to read function while iterator will call the same function with default `count = 1` and yield single byte.

vados-cosmonic: Ah yeah, thanks for the explanation -- that sounds fine. I think for this kind of interop we probably need a more concrete way (probably making our own symbols available or something like that), but this is definitely fine as a direction (and maybe even the final direction).

vados-cosmonic (approved):
LGTM 🚀 

Trying to fix the puppeteer failures in CI, and will get this merged!

PR #1442, opened 2026-05-04
fix(p3-shim): export types namespace

andreiltd:
Export some missing p3 namespaces

### DIFF
--- a/packages/preview3-shim/lib/nodejs/sockets/ip-name-lookup.js
+++ b/packages/preview3-shim/lib/nodejs/sockets/ip-name-lookup.js
@@ -1,0 +1,92 @@
+import { lookup } from "node:dns/promises";
+import { isIP } from "node:net";
+import { domainToASCII } from "node:url";
+
+import { ipv4ToTuple, ipv6ToTuple } from "./address.js";
+import { SocketError } from "./error.js";
+
+const DNS_ERROR_MAP = {
+  EACCES: "access-denied",
+  EPERM: "access-denied",
+  EINVAL: "invalid-argument",
+  ENOTFOUND: "name-unresolvable",
+  EAI_ADDRFAMILY: "name-unresolvable",
+  EAI_AGAIN: "temporary-resolver-failure",
+  EAI_FAIL: "permanent-resolver-failure",
+  EAI_NODATA: "name-unresolvable",
+  EAI_NONAME: "name-unresolvable",
+};
+
+function ipAddress(address, family) {
+  switch (family) {
+    case 4:
+      return { tag: "ipv4", val: ipv4ToTuple(address) };
+    case 6:
+      return { tag: "ipv6", val: ipv6ToTuple(address) };
+    default:
+      throw new SocketError("unknown", `Unknown IP address family: ${family}`);
+  }
+}
+
+function normalizeLookupName(name) {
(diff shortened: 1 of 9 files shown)


packages/preview3-shim/lib/nodejs/sockets/ip-name-lookup.js
```diff
@@ -0,0 +1,92 @@
...
+
+      if (result.length === 0) {
+        throw new SocketError("name-unresolvable");
+      }
+      return result;
+    } catch (error) {
+      if (error instanceof SocketError) {
+        throw error;
+      }
+      throw new SocketError(DNS_ERROR_MAP[error?.code] ?? "unknown", error?.message, undefined, {
```
vados-cosmonic, line 87: Lazy ask -- should this be `other` ?

https://github.com/WebAssembly/WASI/blob/main/proposals/sockets/wit-0.3.0-draft/types.wit#L92C1-L92C31

Does that get converted elsewhere?

packages/preview3-shim/lib/nodejs/sockets/tcp.js
```diff
@@ -84,6 +84,10 @@ export class TcpSocket {
     sendBufferSize: undefined,
   };
 
+  static create(addressFamily) {
```
vados-cosmonic, line 101: This is a utility function right? 👍 

Any reason to not go with a `constructor` here? I'm definitely happy either way -- `TcpSocket.create` vs `new TCPSocket(...)`

Oh and if we go with `TcpSocket.create`, how about making sure no one can call the constructor?

vados-cosmonic (approved):
LGTM 🚀 

Few notes,  but you can take them or leave them (they're certainly not blocking)

packages/preview3-shim/lib/nodejs/sockets/tcp.js
andreiltd: This is actually from `wit` ( I will add docs here):

https://github.com/WebAssembly/WASI/blob/main/proposals/sockets/wit-0.3.0-draft/types.wit#L195

```wit
        /// Create a new TCP socket.
        create: static func(address-family: ip-address-family) -> result<tcp-socket, error-code>;
```

packages/preview3-shim/lib/nodejs/sockets/ip-name-lookup.js
andreiltd: That looks like a good idea, yes. Let me fix this.

packages/preview3-shim/lib/nodejs/sockets/tcp.js
vados-cosmonic: 100% forgot about this, thanks!

packages/preview3-shim/lib/nodejs/sockets/ip-name-lookup.js
andreiltd: OK, it looks like version wasi version that we currently use does not have `other` variant just yet:

https://github.com/bytecodealliance/jco/blob/main/packages/jco/test/fixtures/p3/wit/deps/wasi-sockets-0.3.0-rc-2026-02-09/package.wit#L698

I will leave this as is and change when we update `wasi`

PR #1445, opened 2026-05-05
fix(bindgen): lift numeric p3 lists and streams as typed arrays

andreiltd:
Change p3 flat list lifts and numeric stream reads to use the JS typed-array representation.

### DIFF
--- a/crates/js-component-bindgen/src/intrinsics/lift.rs
+++ b/crates/js-component-bindgen/src/intrinsics/lift.rs
@@ -820,7 +820,10 @@ impl LiftIntrinsic {
                     function {lift_flat_list_fn}(meta) {{
                         const {{ elemLiftFn, elemSize32, elemAlign32, knownLen, typedArray }} = meta;
 
-                        const listValue = values => typedArray === undefined ? values : new typedArray(values);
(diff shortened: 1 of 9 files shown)


vados-cosmonic (approved):
LGTM :rocket:

Just one minor nit

crates/js-component-bindgen/src/intrinsics/lift.rs
```diff
@@ -818,7 +818,9 @@ impl LiftIntrinsic {
 
                 output.push_str(&format!(r#"
                     function {lift_flat_list_fn}(meta) {{
-                        const {{ elemLiftFn, elemSize32, elemAlign32, knownLen }} = meta;
+                        const {{ elemLiftFn, elemSize32, elemAlign32, knownLen, typedArray }} = meta;
+
+                        const listValue = values => typedArray === undefined ? values : new typedArray(values);
```
vados-cosmonic: NIT: we can do this check once and keep the function the same every time it's ued right?

vados-cosmonic (approved):
LGTM 🚀 

(this time the nit is properly in the review?)

PR #1446, opened 2026-05-05
feat(p3-shim): bump latest wasi rc to 2026-03-15

andreiltd:
It's hard to believe this is a few weeks worth of diff (2026-02-09 -> 2026-03-15) 😅


packages/preview3-shim/lib/nodejs/http/request.js
```diff
@@ -303,7 +292,7 @@ export class Request {
    */
   setPathWithQuery(pathWithQuery) {
     validateUrlPart(pathWithQuery, UrlPart.PATH_WITH_QUERY);
-    this.#pathWithQuery = pathWithQuery;
+    this.#pathWithQuery = pathWithQuery ?? undefined;
```
vados-cosmonic, line 295: so if `pathWithQuery` is missing, I think `validateUrlPart` will fail -- it only "allows" `null`, AFAICT.

vados-cosmonic (approved):
LGTM 🚀

packages/preview3-shim/lib/nodejs/http/request.js
andreiltd: I think this should work because `validateUrlPart` uses `value == null`, which will catch both null and undefined, right?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/null#examples:

```js
null === undefined; // false
null == undefined; // true
```

PR #1448, opened 2026-05-06
fix(ci): detect jspi support in nodejs

andreiltd:
It looks like `jspi` option is gone in node 26. This makes CI fail with:

```
/Users/runner/hostedtoolcache/node/26.0.0/arm64/bin/node: bad option: --experimental-wasm-jspi
```

This patch tries to detect if `jspi` support is already there and skips experimental flag it that's the case.

https://github.com/WebAssembly/js-promise-integration/blob/main/proposals/js-promise-integration/Overview.md#specification


vados-cosmonic (approved):
LGTM 🚀

PR #1449, opened 2026-05-06
fix(bindgen): use async ABI for async imports

andreiltd:
What it says on the tin plus similar fix for resources.


vados-cosmonic (approved):
LGTM 🚀 

Thanks for fixing this!

PR #1450, opened 2026-05-06
fix(bindgen): populate imported resource lower metadata

andreiltd:
The failure shows up with async imports that return resources inside result, for example filesystem ops like this:

```
result<own<descriptor>, error-code>
```

throws:

```
missing/invalid resource metadata
```

Basically `imports_resource_types` maps WIT type id to the runtime resource type. But the lowering for `own<T>` needs the reverse mapping: given the concrete resource type, find the WIT resource type id so it can generate correct func.


vados-cosmonic (approved):
LGTM 🚀 

Thanks for fixing this bit of the metadata (it's as bit messy to have both those lookups, I know!)

PR #1452, opened 2026-05-08
fix(bindgen): ensure moving storage pointers by abi size

andreiltd:
The patch includes lower and lift side fixes. The lower side fix is basically: treat `ctx.storagePtr` as a canon abi cursor and move it by abi size rather than stopping where nested lift ended.

The general pattern is this:
- align cursor
- remember slot start
- lower nested value
- advance cursor to at least slot start + abi slot size

The last step is really a gist of it, because nested values can write less than their abi slot size.

For example for records that looks like this:

```
const originalPtr = ctx.storagePtr;

 for each field:
   align ctx.storagePtr to field align
   fieldPtr = ctx.storagePtr
   lower field
   ctx.storagePtr = Math.max(ctx.storagePtr, fieldPtr + fieldSize)

 ctx.storagePtr = Math.max(ctx.storagePtr, originalPtr + recordSize)
 align ctx.storagePtr to record align
```

The lifting is basically the same but it accounts for direct params reading so the algorithm annoyingly checks if pointers are not undefined. I think there is a refactor opportunity here: split direct params and memory indirection modes but I wanted to focus on correctness fixes here.

The transpiler fix just ensures that each nested element gets its own abi metadata rather than using parent metadata: so for instance record gets fields meta and tuples get each element meta.

I first across this issue for descriptor types:

```wit
 record directory-entry {
   type: variant { directory, file, symlink, unknown, ... },
   name: string,
 }
``` 

If the variant case has no payload, old lowering/lifting left the cursor too early, and the following `name: string` was read/written at the wrong offset.

### DIFF
--- a/crates/js-component-bindgen/src/intrinsics/lift.rs
+++ b/crates/js-component-bindgen/src/intrinsics/lift.rs
@@ -730,15 +730,37 @@ impl LiftIntrinsic {
                 let lift_flat_record_fn = self.name();
                 output.push_str(&format!(
                     r#"
-                    function {lift_flat_record_fn}(keysAndLiftFns) {{
+                    function {lift_flat_record_fn}(meta) {{
+                        const {{ fieldMetas, size32: recordSize32, align32: recordAlign32 }} = meta;
                         return function {lift_flat_record_fn}Inner(ctx) {{
                             {debug_log_fn}('[{lift_flat_record_fn}()] args', {{ ctx }});
 
+                            const originalPtr = ctx.storagePtr;
                             const res = {{}};
-                            for (const [key, liftFn, _size32, _align32] of keysAndLiftFns) {{
+                            for (const [key, liftFn, size32, align32] of fieldMetas) {{
+                                let fieldPtr;
+                                if (ctx.storagePtr !== undefined) {{
+                                    const rem = ctx.storagePtr % align32;
+                                    if (rem !== 0) {{ ctx.storagePtr += align32 - rem; }}
+                                    fieldPtr = ctx.storagePtr;
+                                }}
+
                                 let [val, newCtx] = liftFn(ctx);
                                 res[key] = val;
                                 ctx = newCtx;
+
+                                if (fieldPtr !== undefined) {{
+                                    ctx.storagePtr = Math.max(ctx.storagePtr, fieldPtr + size32);
+                                }}
+                            }}
+
+                            if (originalPtr !== undefined) {{
+                                ctx.storagePtr = Math.max(ctx.storagePtr, originalPtr + recordSize32);
+                            }}
+
+                            if (ctx.storagePtr !== undefined) {{
+                                const rem = ctx.storagePtr % recordAlign32;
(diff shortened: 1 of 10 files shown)


vados-cosmonic (approved):
LGTM 🚀 


Thanks for fixing this and updating the tests -- glad those bits were easy to extend. Tests look good to me (and feel free to refactor/aggressively cut if any of it awkward)

crates/js-component-bindgen/src/intrinsics/lift.rs
```diff
@@ -789,21 +811,42 @@ impl LiftIntrinsic {
...
+                            if (origUseParams) {{
+                                if (caseFlatCount === undefined || variantFlatCount === undefined) {{
+                                    throw new Error('variant flat count metadata is missing');
+                                }}
+                                if (caseFlatCount === null || variantFlatCount === null) {{
+                                    throw new Error('cannot lift variant with unknown flat count');
+                                }}
+                                const remainingPayloadParams = variantFlatCount - 1 - caseFlatCount;
+                                if (remainingPayloadParams < 0) {{
+                                    throw new Error(`invalid variant flat count metadata);
```
vados-cosmonic:
```suggestion
                                    throw new Error('invalid variant flat count metadata');
```

vados-cosmonic:
NOTE: I'm force merging this as the failures in CI (on only macOS) seem to be unrelated:

```
dlopen /Users/runner/.cache/puppeteer/chrome/mac_arm-139.0.7258.68/chrome-mac-arm64/Google Chrome for Testing.app/Contents/MacOS/../Frameworks/Google Chrome for Testing Framework.framework/Versions/139.0.7258.68/Google Chrome for Testing Framework: dlopen(/Users/runner/.cache/puppeteer/chrome/mac_arm-139.0.7258.68/chrome-mac-arm64/Google Chrome for Testing.app/Contents/MacOS/../Frameworks/Google Chrome for Testing Framework.framework/Versions/139.0.7258.68/Google Chrome for Testing Framework, 0x0105): tried: '/Users/runner/.cache/puppeteer/chrome/mac_arm-139.0.7258.68/chrome-mac-arm64/Google Chrome for Testing.app/Contents/MacOS/../Frameworks/Google Chrome for Testing Framework.framework/Versions/139.0.7258.68/Google Chrome for Testing Framework' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/Users/runner/.cache/puppeteer/chrome/mac_arm-139.0.7258.68/chrome-mac-arm64/Google Chrome for Testing.app/Contents/MacOS/../Frameworks/Google Chrome for Testing Framework.framework/Versions/139.0.7258.68/Google Chrome for Testing Framework' (no such file), '/Users/runner/.cache/puppeteer/chrome/mac_arm-139.0.7258.68/chrome-mac-arm64/Google Chrome for Testing.app/Contents/MacOS/../Frameworks/Google Chrome for Testing Framework.framework/Versions/139.0.7258.68/Google Chrome for Testing Framework' (no such file).
```

PR #1455, opened 2026-05-08
feat: add preview3 transpile mappings

andreiltd:
@vados-cosmonic I'm not sure if this is what we want here -- adding p3 mappings unconditionally?


vados-cosmonic (approved):
LGTM 🚀 

Hey @andreiltd yeah, this is a bit unfortunate because with the way Jco is currently configured we basically include p2-shim unconditionally, so being consistent there is fine. 

That said, this is something we need to do explicitly in the next breaking change -- i.e. Jco 2.x should likely *not* do this by default and should ask the user more specifically what they want and give them the control instead of trying to guess.

Will add that to the list of things to change before the next breaking change.

PR #1459, opened 2026-05-11
fix(bindgen): allow external crates to construct RenderIntrinsicsArgs

chaynabors:
Since #811, the fields on RenderIntrinsicsArgs and the AsyncDeterminismProfile enum are pub(crate). There is no public constructor. componentize-js uses the transpile-bindgen feature and cannot construct RenderIntrinsicsArgs to call render_intrinsics.

This exposes both as pub. No behavior change.

### DIFF
--- a/crates/js-component-bindgen/src/intrinsics/mod.rs
+++ b/crates/js-component-bindgen/src/intrinsics/mod.rs
@@ -1043,7 +1043,7 @@ impl Intrinsic {
 
 /// Profile for determinism to be used by async implementation
 #[derive(Debug, Default, PartialEq, Eq)]
-pub(crate) enum AsyncDeterminismProfile {
+pub enum AsyncDeterminismProfile {
     /// Allow random ordering non-determinism
     #[default]
     Random,
@@ -1069,15 +1069,15 @@ impl std::fmt::Display for AsyncDeterminismProfile {
 /// Arguments to `render_intrinsics`
 pub struct RenderIntrinsicsArgs<'a> {
     /// List of intrinsics being built for use
-    pub(crate) intrinsics: &'a mut BTreeSet<Intrinsic>,
+    pub intrinsics: &'a mut BTreeSet<Intrinsic>,
     /// Whether to use NodeJS compat
-    pub(crate) no_nodejs_compat: bool,
+    pub no_nodejs_compat: bool,
     /// Whether instantiation has occurred
-    pub(crate) instantiation: bool,
+    pub instantiation: bool,
     /// The kind of determinism to use
-    pub(crate) determinism: AsyncDeterminismProfile,
+    pub determinism: AsyncDeterminismProfile,
     /// Options provided when performing transpilation
-    pub(crate) transpile_opts: &'a TranspileOpts,
+    pub transpile_opts: &'a TranspileOpts,
 }
 
 /// Intrinsics that should be rendered as early as possible


chaynabors:
Closing and filing issue downstream instead.

vados-cosmonic:
Hey @chaynabors is `js-component-bindgen` being used in a downstream dep or vendored version? This *is* probably a change we want to take (as you pointed out, I think it's been impossible to use for quite a while!)

chaynabors:
Thanks for checking in @vados-cosmonic.

I'm working on [this draft](https://github.com/bytecodealliance/ComponentizeJS/pull/336) in componentize-js since our team needs access to component model streams. A few things came up so it wasn't as simple as making a handful of fields public.

Want me to create an issue anywhere? I think some of this work warrants a discussion, and I'm investigating a couple of solutions inside JCO to be able to contribute to that discussion myself. Any chance you could afford an opinion on the draft as to how some of this might look?

Either way, I'll pitch a few things in the coming days if we're relatively aligned.

For more context, we're also driving some async work in the executor as well:
https://github.com/bytecodealliance/wasmtime/issues/12991

vados-cosmonic:
Hey @chaynabors no problem, thanks for the note!

Actually the current upgrade in componentize-js is actually [quite hard right now](https://bytecodealliance.zulipchat.com/#narrow/channel/409526-jco/topic/Updating.20.60ComponentizeJS.60.20deps/with/591868151) -- it's been on the priority list for a bit (I was out for the last ~2 weeks, so it hasn't been done yet). 

I'm going to dig into making that change and this one as well ASAP (currently someone else is pushing forward P3 work in Jco). 

Thanks for the upstream link to `wasmtime` -- thanks for working on the C API!

vados-cosmonic:
Hey heads up https://github.com/bytecodealliance/jco/pull/1462 is merging soon which takes the builder approach

chaynabors:
Thanks so much @vados-cosmonic. We have a major release scheduled for next month, so let me know if you can afford to delegate any of this. If it's faster for you to pick up then don't let me slow you down. I know at least two other things blocking version bumps.

1. Jco needs to expose CanonicalOptions and some post-compile fields.
2. Wirm needs to restore CustomSections::delete or some other solution needs to be implemented.

Both might require API changes depending on how we want to do them, then I'm likely to start funneling WIT streams through if there are no complaints.

Any chance you have a sense for how long the version upgrade will take? Apologies for pushing.

vados-cosmonic:
No problem!

So the [RC for the next release of `js-component-bindgen` is out](https://crates.io/crates/js-component-bindgen/1.19.0-rc.0), I'm going to try to update `componentize-js` very very soon.

It looks like we may need another RC but I'll make one if that is the case!

> Any chance you have a sense for how long the version upgrade will take? Apologies for pushing.

 Unfortunately I can't give a super strict timeline on that (mostly for fear of being wrong!), but I should be done by the end of this week.

vados-cosmonic:
Hey @chaynabors would you mind giving https://crates.io/crates/js-component-bindgen/1.19.0-rc.5 a try? It's not working in `componentize-js` just yet (there's [a PR](https://github.com/bytecodealliance/ComponentizeJS/pull/337)), but if you are using `js-component-bindgen` directly then it should be explorable before it gets released.

chaynabors:
@vados-cosmonic I did get a chance to play around. I hacked an asyncify implementation into the tree and got most of the tests passing from 0, but it had problems like tripling the wasm component size and taking several minutes to build. I ultimately got boxed in needing wasmtime 45 before I could continue getting streams working in this way.

In practice, I don't see a great path forward since we'd either need JIT or stack switching for an implementation I'm personally content with.

I've switched to spinning a custom poll loop in the host just to get something by next month and will create another issue if anything comes up. Thanks for the help!

vados-cosmonic:
> I hacked an asyncify implementation into the tree and got most of the tests passing from 0, but it had problems like tripling the wasm component size and taking several minutes to build. I ultimately got boxed in needing wasmtime 45 before I could continue getting streams working in this way.

Yeah I'm definitely not surprised that the asyncify implementation went that way -- we were very very hesitant to take it on in the past (when pulling in JSPI stuff) for fear of what you ended up running into. 

Thanks for sharing your progress, and thanks in advance for the issues/contributions!

PR #1481, opened 2026-05-14
feat(bindgen): allow expanding numeric chunks in stream lowering

andreiltd:
Adapt stream lowering to expand numeric chunks like `Uint8Array` while preserving leftovers across guest reads. Also split host stream writes into public `write(value)` and internal `writeMany(values)` batch handling.


vados-cosmonic (approved):
LGTM 🚀

PR #1482, opened 2026-05-14
test: add regression test for p2-shim resolve-addresses

### LINKED ISSUE #1454: Add regression test for p2-shim resolve-addresses
What's left here is to add a regression test to make sure this stays fixed/doesn't break in the future.

[EDIT] - We don't have many tests that explore socket functionality (outside of the excellent tests in p3-shim) -- so it might be worth adding a new Rust component to test the shim and call [`resolve-addresses`](https://github.com/WebAssembly/WASI/blob/5b0aa5140c4021071a4acdc35ec2b56a2fb45854/proposals/sockets/wit/ip-name-lookup.wit#L29) to exercise this shim path:

https://github.com/bytecodealliance/jco/blob/9012818bb6d7dc80a7803bdc5150d52285b471cf/packages/preview2-shim/lib/nodejs/sockets.js#L138

_Originally posted by @vados-cosmonic in https://github.com/bytecodealliance/jco/issues/1453#issuecomment-4405127830_

cijiugechu:
Closes #1454

### DIFF
--- a/packages/preview2-shim/test/test.js
+++ b/packages/preview2-shim/test/test.js
@@ -500,6 +500,40 @@ suite("Node.js Preview2", () => {
       }),
     );
   });
+
+  suite("WASI Sockets (IP Name Lookup)", async () => {
+    test(
+      "ipNameLookup.resolveAddresses(): should return valid IP addresses",
+      testWithGCWrap(async () => {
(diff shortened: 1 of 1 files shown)


vados-cosmonic (approved):
LGTM 🚀 

Thanks for the contribution @cijiugechu !

PR #1483, opened 2026-05-14
fix(p3-shim): handle filesystem stream edge cases

andreiltd:
- Accept non zero file offsets by converting wasi BigInt positions to safe numeric node fs positions before read/write calls from worker.
- Reject write/append streams on non writable descriptors.


vados-cosmonic (approved):
LGTM 🚀

PR #1484, opened 2026-05-14
fix(p3-shim): align socket behavior with WASI tests

andreiltd:
The fixes are driven by wasi upstream tests: allow wildcard binds, reject non-bindable addresses, and detect local bind conflicts for tcp and udp sockets.

### DIFF
--- a/packages/preview3-shim/lib/nodejs/sockets/address.js
+++ b/packages/preview3-shim/lib/nodejs/sockets/address.js
@@ -1,4 +1,5 @@
 // Adapted from preview2-shim/lib/io/worker-sockets.js
+import { networkInterfaces } from "node:os";
 
 // TODO(tandr): switch to generated types
 export const IP_ADDRESS_FAMILY = {
@@ -47,7 +48,7 @@ export const ipv4ToTuple = (ipv4) => {
 
 export const isMulticastIpAddress = ({ tag, val: { address } }) =>
   (tag === "ipv4" && address[0] >= 0xe0 && address[0] <= 0xef) ||
-  (tag === "ipv6" && address[0] === 0xff);
+  (tag === "ipv6" && (address[0] & 0xff00) === 0xff00);
 
 export const isIpv4MappedAddress = ({ tag, val: { address } }) =>
   tag === "ipv6" &&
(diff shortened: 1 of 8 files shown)


packages/preview3-shim/lib/nodejs/sockets/address.js
```diff
@@ -66,6 +67,19 @@ export const isBroadcastIpAddress = ({ tag, val: { address } }) =>
...
   !isMulticastIpAddress(a) && !isBroadcastIpAddress(a) && !isWildcardIpAddress(a);
 
+export const isLoopbackIpAddress = ({ tag, val: { address } }) =>
+  (tag === "ipv4" && address[0] === 127) ||
+  (tag === "ipv6" && address.slice(0, 7).every((segment) => segment === 0) && address[7] === 1);
+
+export const ipAddressConflict = (a, b) =>
+  a.tag === b.tag &&
+  Number(a.val.port) !== 0 &&
+  Number(b.val.port) !== 0 &&
```
vados-cosmonic: SUPER MEGA NIT: we can check equals first then only check if one of them is not equal to 0

packages/preview3-shim/lib/nodejs/workers/udp-worker.js
```diff
@@ -61,14 +89,37 @@ async function handleBind({ socketId, localAddress }) {
...
   const socket = sockets.get(socketId);
+  const preserveLocalAddress =
+    socket.connected &&
+    socket.localAddress &&
+    isLoopbackIpAddress(socket.connected) &&
+    isLoopbackIpAddress(remoteAddress);
+
+  if (preserveLocalAddress) {
+    const localAddress = socket.localAddress;
+    await new Promise((resolve) => socket.udp.close(resolve));
```
vados-cosmonic, line 106: could we move the socket closing outside of the if? Not sure if `socket.localAddress` changes as a result of the close, but if it doesn't it feels like we first close *then* either do the bind or set the local address.

packages/preview3-shim/lib/nodejs/workers/udp-worker.js
```diff
@@ -83,34 +134,40 @@ async function handleSend({ socketId, data, remoteAddress }) {
...
 
   await new Promise((resolve, reject) => {
-    if (isConnected) {
-      socket.udp.send(data, (err) => (err ? reject(err) : resolve()));
-    } else {
-      const addr = serializeIpAddress(remoteAddress);
-      const port = remoteAddress.val.port;
-
-      socket.udp.send(data, port, addr, (err) => (err ? reject(err) : resolve()));
+    try {
```
vados-cosmonic: NIT: What do you think about making this lambda an async one?

vados-cosmonic (dismissed):
LGTM 🚀 

Nothing but nits

PR #1486, opened 2026-05-14
fix(p3): use async lower result pointer from params trail

andreiltd:
Async canonical lower appends a result pointer to flat params when flat_results is non-empty. The async task metadata was treating the first raw param as the result pointer, which corrupts params for async imports that have both params and results.

The relevant spec snippet can be found here: 
https://github.com/WebAssembly/component-model/blob/main/design/mvp/CanonicalABI.md#flattening


vados-cosmonic (approved):
LGTM 🚀

PR #1504, opened 2026-05-18
fix(p3-shim): complete HTTP transmit futures from worker

andreiltd:
This includes few fixes driven by upstream tests but most importantly it fixes sending http request bodies and trailers concurrently with response handling, and settle the request transmit future only after upload completion.


vados-cosmonic (approved):
LGTM 🚀

PR #1505, opened 2026-05-18
test(jco): increase default test timeout

vados-cosmonic:


vados-cosmonic:
So there is a legitimate bug in lowering on particularly Node 26 (at least, node 26.1.0) -- the fix here isn't tot increase the test timeout -- I'll need to look into this.

PR #1507, opened 2026-05-19
fix(p2-shim): Fields.set fails on valid input

### LINKED ISSUE #1503: preview2-shim/nodejs: Fields.set fails on valid input
## TL;DR

`Fields.set` in [`lib/nodejs/http.js`](https://github.com/bytecodealliance/jco/blob/main/packages/preview2-shim/lib/nodejs/http.js#L502-L534) fails on input that both the WIT spec and Node's own header validators consider valid. Two specific defects need fixing:

- **Post-validation throw** ([L517](https://github.com/bytecodealliance/jco/blob/main/packages/preview2-shim/lib/nodejs/http.js#L517)) — an unconditional `throw { tag: "invalid-syntax" }` placed right after the `validateHeaderValue` try/catch. Every call hits this regardless of input.
- **Unguarded splice** ([L527](https://github.com/bytecodealliance/jco/blob/main/packages/preview2-shim/lib/nodejs/http.js#L527)) — `tableEntries.splice(...)` is called outside the `if (tableEntries)` guard. For any new key this would `TypeError`, currently masked because the post-validation throw fires first.

The browser variant at [`lib/browser/http.js#L49-L74`](https://github.com/bytecodealliance/jco/blob/main/packages/preview2-shim/lib/browser/http.js#L49-L74) has neither defect and can serve as the reference. There is no existing test in `packages/preview2-shim/test/test.js` that calls `Fields.set(...)`.

---

## Observed facts

- A Rust wasi-http component calling the following panics with `HeaderError::InvalidSyntax` (verified on `0.17.9`):

  ```rust
  Fields::new().set(
      &"content-type".to_string(),
      &[b"text/plain; charset=utf-8".to_vec()],
  )
  ```

- Switching `set` → `append` on the same `Fields` value succeeds and produces a Node `http.ServerResponse` whose `content-type` header is set as expected:

  ```rust
  Fields::new().append(
      &"content-type".to_string(),
      &b"text/plain; charset=utf-8".to_vec(),
  )
  ```

- The same input passes `node:http`'s own validation:

  ```js
  import http from "node:http";
  http.validateHeaderName("content-type");                                // OK
  http.validateHeaderValue("content-type", "text/plain; charset=utf-8");  // OK
  ```

- `lib/nodejs/http.js` contains the following structures:

  ```js
  // L511-L518: post-validation throw inside the for-loop, after the try/catch
  for (const value of values) {
    try {
      validateHeaderValue(name, new TextDecoder().decode(value));
    } catch {
      throw { tag: "invalid-syntax" };
    }
    throw { tag: "invalid-syntax" };  // post-validation throw — always reached after validation succeeds
  }
  ```

  ```js
  // L524-L530: unguarded splice — called outside the if (tableEntries) guard
  const tableEntries = this.#table.get(lowercased);
  if (tableEntries) {
    this.#entries = this.#entries.filter((entry) => !tableEntries.includes(entry));
  }
  tableEntries.splice(0, tableEntries.length);  // unguarded splice — TypeError if tableEntries is undefined
  ```

  With the post-validation throw locally removed, the unguarded splice raises `TypeError: Cannot read properties of undefined (reading 'splice')` for any key not already present in `this.#table`.

- The browser variant has different structure:

  ```js
  // L61-L67 of lib/browser/http.js
  const tableEntries = this.#table.get(lowercased);
  if (tableEntries) {
    this.#entries = this.#entries.filter((entry) => !tableEntries.includes(entry));
    tableEntries.splice(0, tableEntries.length);
  } else {
    this.#table.set(lowercased, []);  // ← initialize new key
  }
  ```

---

## Spec reference

`wasi:http/types@0.2.0`'s `fields.set` is declared as:

```wit
set: func(name: field-key, value: list<field-value>) -> result<_, header-error>;
```

The accompanying doc comment says:

> *Fails with `header-error.immutable` if the `fields` are immutable.*

`header-error` variants are `invalid-syntax`, `forbidden`, and `immutable`. The spec does not state that valid input on a mutable, non-forbidden field name should fail.

---

## What I think this means (opinion, derived from facts above)

- The Rust input is valid by Node's own header validation, and is permitted by the WIT spec. Yet `Fields.set` returns `invalid-syntax`. The post-validation throw is the only path that can produce this for valid input → **the post-validation throw is firing on input that the spec considers valid**.
- The unguarded splice on a possibly-undefined value is structurally unsafe regardless of the post-validation throw: the guard above only protects `filter`, not `splice`. The reason it has not surfaced in practice is that the post-validation throw fires first.
- Reading `packages/preview2-shim/test/test.js` (969 lines), the only references to `Fields` are:
  - L136: `const { OutgoingRequest, OutgoingBody, Fields } = http.types;`
  - L139: `new Fields([...])` constructor call

  There is no test that calls `Fields.set(...)` on this target. If such a test existed, the post-validation throw would have surfaced.

---

## Proposed fix (opinion)

Port the structure from `lib/browser/http.js`. Edits against the current
`lib/nodejs/http.js` are inlined as comments:

```js
set(name, values) {
  if (this.#immutable) throw { tag: "immutable" };
  try { validateHeaderName(name); } catch { throw { tag: "invalid-syntax" }; }
  for (const value of values) {
    try { validateHeaderValue(name, new TextDecoder().decode(value)); }
    catch { throw { tag: "invalid-syntax" }; }
    // [REMOVE] the unconditional `throw { tag: "invalid-syntax" }` at L517 — delete it.
  }
  const lowercased = name.toLowerCase();
  if (_forbiddenHeaders.has(lowercased)) throw { tag: "forbidden" };
  // [CHANGE] `const` → `let` so the new else-branch below can reassign.
  let tableEntries = this.#table.get(lowercased);
  if (tableEntries) {
    this.#entries = this.#entries.filter(e => !tableEntries.includes(e));
    // [MOVE] `tableEntries.splice(...)` moved INSIDE the `if (tableEntries)` guard
    // (was on L527 outside the guard).
    tableEntries.splice(0, tableEntries.length);
  } else {
    // [ADD] initialize the entry list for a new key before pushing below.
    this.#table.set(lowercased, []);
    tableEntries = this.#table.get(lowercased);
  }
  for (const value of values) {
    const entry = [name, value];
    this.#entries.push(entry);
    tableEntries.push(entry);
  }
}
```

Plus a regression test that constructs a `Fields`, calls `set("content-type", [encode("text/plain")])`, then verifies via `entries()`.

Happy to send a PR.

glassmonkey:
Fixes #1503.

## Changes

`packages/preview2-shim/lib/nodejs/http.js` — port the `Fields.set` structure from `lib/browser/http.js`:

- Drop the unconditional `throw { tag: "invalid-syntax" }` placed inside the values loop after the `validateHeaderValue` try/catch.
- Change `const tableEntries` to `let` so the new branch can reassign.
- Move `tableEntries.splice(...)` inside the `if (tableEntries)` guard (was previously outside, crashing with `TypeError` for any new key once the unconditional throw was bypassed).
- Add an `else` branch that initializes `this.#table.set(lowercased, [])` before the subsequent push loop.

`packages/preview2-shim/test/test.js` — seven regression tests, each covering a distinct branch of `Fields.set`. Both green-path tests were verified TDD-style (RED with the bug present, GREEN with the fix):

- `Fields.set on a fresh Fields` — new-key path, single value (regression for #1503).
- `Fields.set replaces existing values` — existing-key path through the splice branch.
- `Fields.set with an empty values list registers the key with no entries` — locks in the behavior of the new else-branch when `values` is empty.
- `Fields.set throws immutable when fields are locked` — verifies the immutable guard against a Fields locked via `OutgoingRequest`.
- `Fields.set throws invalid-syntax for an invalid name` — `validateHeaderName` rejection path.
- `Fields.set throws invalid-syntax for an invalid value` — `validateHeaderValue` rejection path (one element of `values` contains an invalid byte).
- `Fields.set throws forbidden for restricted header names` — `_forbiddenHeaders` (case-insensitive) for Connection / Keep-Alive / Host.

Tests use `Fields.fromList([])` to mirror the construction style from `lib/browser/http.js` tests.

## Open question: browser-side tests

Should this PR also add the equivalent `Fields.set` tests to `packages/preview2-shim/test/browser.js`?

State of the browser-side tests today:

- `lib/browser/http.js`'s `Fields.set` implementation is structurally correct (it's the reference we ported from), so the bug being fixed here does not exist on the browser target.
- However, `test/browser.js` only exercises `Fields.set` once (L919, for the `immutable` error case). The success paths (new-key / replace / empty values) and the other error variants (`invalid-syntax`, `forbidden`) on `Fields.set` are not covered there either.

Trade-offs:

- **Adding parity now** broadens this PR but gives symmetric coverage across both targets.
- **Leaving it out** keeps this PR focused on the nodejs bug; a follow-up issue can track adding equivalent browser-side coverage. Since the browser implementation is not buggy today, this is purely coverage extension rather than regression repair.

I lean toward leaving the browser side as-is in this PR and filing a follow-up issue, but happy to add the browser tests here if you'd prefer. Please advise.

## Test plan

- [x] `npm test --workspace=@bytecodealliance/preview2-shim` locally: 60/61 pass (the remaining failure is `WASI Sockets (TCP) > tcp.connect()`, which is unrelated and was already failing on `upstream/main` before this change in this local environment).
- [x] All seven new tests verified individually against the fix.
- [x] `npm run lint` / `npm run fmt:check` pass locally (oxlint / oxfmt darwin-arm64 native bindings required a manual `npm install --no-save @oxlint/binding-darwin-arm64 @oxfmt/binding-darwin-arm64` to work around an npm optional-deps resolution issue; the formatter then made one small whitespace change which is included in this commit).

### DIFF
--- a/packages/preview2-shim/lib/nodejs/http.js
+++ b/packages/preview2-shim/lib/nodejs/http.js
@@ -499,6 +499,19 @@ class Fields {
     }
     return tableEntries.map(([, v]) => v);
   }
+  /**
+   * WIT spec (https://github.com/WebAssembly/WASI/blob/91bb44c3c3a9b1e09187db23c85fa844d1fd6b15/proposals/http/wit/types.wit#L215-L223):
+   *
+   * > Set all of the values for a name. Clears any existing values for that
+   * > name, if they have been set.
+   * >
+   * > Fails with `header-error.immutable` if the `fields` are immutable.
+   * >
+   * > Fails with `header-error.invalid-syntax` if the `field-name` or any of
+   * > the `field-value`s are syntactically invalid.
+   *
+   * The existing-branch splice/reuse is an allocation optimization — values are cleared, not retained.
+   */
   set(name, values) {
     if (this.#immutable) {
       throw { tag: "immutable" };
--- a/packages/preview2-shim/lib/nodejs/http.js
+++ b/packages/preview2-shim/lib/nodejs/http.js
@@ -514,17 +514,19 @@ class Fields {
       } catch {
         throw { tag: "invalid-syntax" };
       }
-      throw { tag: "invalid-syntax" };
     }
     const lowercased = name.toLowerCase();
     if (_forbiddenHeaders.has(lowercased)) {
       throw { tag: "forbidden" };
     }
-    const tableEntries = this.#table.get(lowercased);
+    let tableEntries = this.#table.get(lowercased);
     if (tableEntries) {
       this.#entries = this.#entries.filter((entry) => !tableEntries.includes(entry));
+      tableEntries.splice(0, tableEntries.length);
+    } else {
+      this.#table.set(lowercased, []);
+      tableEntries = this.#table.get(lowercased);
     }
-    tableEntries.splice(0, tableEntries.length);
     for (const value of values) {
       const entry = [name, value];
       this.#entries.push(entry);
--- a/packages/preview2-shim/lib/browser/http.js
+++ b/packages/preview2-shim/lib/browser/http.js
@@ -46,6 +46,19 @@ class Fields {
     return tableEntries.map(([, v]) => v);
   }
 
+  /**
+   * WIT spec (https://github.com/WebAssembly/WASI/blob/91bb44c3c3a9b1e09187db23c85fa844d1fd6b15/proposals/http/wit/types.wit#L215-L223):
+   *
+   * > Set all of the values for a name. Clears any existing values for that
+   * > name, if they have been set.
+   * >
+   * > Fails with `header-error.immutable` if the `fields` are immutable.
+   * >
+   * > Fails with `header-error.invalid-syntax` if the `field-name` or any of
+   * > the `field-value`s are syntactically invalid.
+   *
+   * The existing-branch splice/reuse is an allocation optimization — values are cleared, not retained.
+   */
   set(name, values) {
     if (this.#immutable) {
       throw { tag: "immutable" };
--- a/packages/preview2-shim/test/test.js
+++ b/packages/preview2-shim/test/test.js
@@ -128,6 +128,108 @@ suite("Node.js Preview2", () => {
     }
   });
 
+  test("Fields.set on a fresh Fields", async () => {
+    const { http } = await import("@bytecodealliance/preview2-shim");
+    const { Fields } = http.types;
+    const encoder = new TextEncoder();
+
+    // valid set on a new key must succeed (regression for #1503)
+    const fields = Fields.fromList([]);
+    fields.set("content-type", [encoder.encode("text/plain; charset=utf-8")]);
+
+    const entries = fields.entries();
+    assert.strictEqual(entries.length, 1);
+    assert.strictEqual(entries[0][0], "content-type");
+    assert.deepEqual(entries[0][1], encoder.encode("text/plain; charset=utf-8"));
+  });
+
+  test("Fields.set replaces existing values", async () => {
+    const { http } = await import("@bytecodealliance/preview2-shim");
+    const { Fields } = http.types;
+    const encoder = new TextEncoder();
+
+    const fields = Fields.fromList([["content-type", encoder.encode("text/plain")]]);
+    fields.set("content-type", [encoder.encode("application/json")]);
+
+    const entries = fields.entries();
+    assert.strictEqual(entries.length, 1);
+    assert.deepEqual(entries[0][1], encoder.encode("application/json"));
+  });
+
+  test("Fields.set with an empty values list registers the key with no entries", async () => {
+    // The fix introduces a new else-branch that initializes
+    // `this.#table.set(lowercased, [])` for new keys. Calling set with an
+    // empty values list exercises that branch without pushing any entry,
+    // so this locks in the resulting state.
+    const { http } = await import("@bytecodealliance/preview2-shim");
+    const { Fields } = http.types;
+
+    const fields = Fields.fromList([]);
+    fields.set("x-foo", []);
+
+    assert.strictEqual(fields.has("x-foo"), true);
+    assert.deepEqual(fields.get("x-foo"), []);
+    assert.strictEqual(fields.entries().length, 0);
+  });
+
+  test("Fields.set throws immutable when fields are locked", async () => {
+    const { http } = await import("@bytecodealliance/preview2-shim");
+    const { Fields, OutgoingRequest } = http.types;
+    const encoder = new TextEncoder();
+
+    // OutgoingRequest's constructor locks the Fields passed in.
+    const hdrs = Fields.fromList([]);
+    new OutgoingRequest(hdrs);
+
+    throws(
+      () => hdrs.set("content-type", [encoder.encode("text/plain")]),
+      (err) => err?.tag === "immutable",
+    );
+  });
+
+  test("Fields.set throws invalid-syntax for an invalid name", async () => {
+    const { http } = await import("@bytecodealliance/preview2-shim");
+    const { Fields } = http.types;
+    const encoder = new TextEncoder();
+
+    const fields = Fields.fromList([]);
+    throws(
+      // names with spaces are rejected by node:http's validateHeaderName
+      () => fields.set("bad header", [encoder.encode("ok value")]),
+      (err) => err?.tag === "invalid-syntax",
+    );
+  });
+
+  test("Fields.set throws invalid-syntax for an invalid value", async () => {
+    const { http } = await import("@bytecodealliance/preview2-shim");
+    const { Fields } = http.types;
+    const encoder = new TextEncoder();
+
+    const fields = Fields.fromList([]);
+    throws(
+      // values containing CR/LF are rejected by node:http's validateHeaderValue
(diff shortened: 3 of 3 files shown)


packages/preview2-shim/lib/nodejs/http.js
```diff
@@ -514,17 +514,19 @@ class Fields {
...
     }
     const lowercased = name.toLowerCase();
     if (_forbiddenHeaders.has(lowercased)) {
       throw { tag: "forbidden" };
     }
-    const tableEntries = this.#table.get(lowercased);
+    let tableEntries = this.#table.get(lowercased);
     if (tableEntries) {
       this.#entries = this.#entries.filter((entry) => !tableEntries.includes(entry));
+      tableEntries.splice(0, tableEntries.length);
```
vados-cosmonic, line 538: `set()` is actually supposed to [clear any existing values for the given name](https://github.com/WebAssembly/WASI/blob/91bb44c3c3a9b1e09187db23c85fa844d1fd6b15/proposals/http/wit/types.wit#L168), so it shouldn't actually maintain any existing entries.

It seems like this code copies the existing values and eventually adds on to them.

vados-cosmonic (commented):
Thanks for making this contribution -- one concern about the logic for `set()`!

packages/preview2-shim/lib/nodejs/http.js
glassmonkey: You're right, sorry — I missed that set() should clear existing values per the spec. I'll fix Fields.set to clear-and-replace properly

glassmonkey: after tracing through, I now think this PR's implementation does match the spec.

**Evidence:**
- On the current branch, `set("X-Foo", [v1, v2])` then `set("X-Foo", [v3])` returns `["v3"]` from `get()` and `[["X-Foo","v3"]]` from `entries()`. The [`filter` on `#entries`](https://github.com/glassmonkey/jco/blob/9ef7c534/packages/preview2-shim/lib/nodejs/http.js#L524) removes the old refs and the [`splice`](https://github.com/glassmonkey/jco/blob/9ef7c534/packages/preview2-shim/lib/nodejs/http.js#L525) empties the table array in place before the push loop.
- The nodejs implementation is structurally equivalent to the existing [`lib/browser/http.js`](https://github.com/glassmonkey/jco/blob/9ef7c534/packages/preview2-shim/lib/browser/http.js#L49-L74) one.

That said, the splice/reuse pattern hides the intent enough to misread — it tripped me when I ported it. Do you think it's worth refactoring
  Fields.set on both targets to build a fresh array and Map.set(lowercased, freshArray) it in, so clear-and-replace reads from one line?

vados-cosmonic: No worries! And yeah, I definitely think that it shouldn't have been written in a way that makes it so easy to misuse -- I can only guess the intent was to avoid allocations by using the existing list. What do you think? Can you think of a different reason? Bit of a chesteron's fence we have here. 

With hindsight, I think one alternative solution would have been be to add a comment to `set()` that makes it clear what the function is supposed to do/caveats that are important with a link (possibly dead, over time) to the relevant WIT.

glassmonkey: Thank you for your kind words.

I share your perspective on this. I suspect the intention behind that implementation was to prioritize memory efficiency and minimize allocations as an internal quality measure.


The trade-off, of course, is that such an implementation can be prone to bugs. However, since we have coverage through tests, I feel that the risk is sufficiently mitigated and does not pose a significant disadvantage.


Given this, I'll keep the current implementation and handle the situation by adding a descriptive comment.

glassmonkey: Added a JSDoc comment to `Fields.set` on both [nodejs](https://github.com/glassmonkey/jco/blob/8ae05549/packages/preview2-shim/lib/nodejs/http.js#L502-L515) and [browser](https://github.com/glassmonkey/jco/blob/8ae05549/packages/preview2-shim/lib/browser/http.js#L49-L62) targets, quoting the WIT spec text and noting that the existing-branch splice/reuse is an allocation optimization (values are cleared, not retained). Pushed in 8ae05549.

Thanks again for the Chesterton's fence framing — comment over refactor felt like the right call given the test coverage we now have.

vados-cosmonic: Glad I could help! These changes look great I'm going to get this in ASAP.

A bunch of the repo has changed under you (we're on `pnpm` now!) but I don't see any conflicts so I think we're good.

vados-cosmonic (approved):
LGTM 🚀 

Thanks for working on this @glassmonkey !

vados-cosmonic:
Ah it looks like there is an issue with the commit title

```
⧗   input: fix(p2-shim): Fields.set fails on valid input
```

Our `commitlint` is a bit picky -- capital letters cannot be used after the scope... The following would work:

```
fix(p2-shim): ensure Fields.set does not fail on valid input
```

glassmonkey:
@vados-cosmonic 

My apologies for the oversight. I will rebase the commit with the corrected message and push the changes.

glassmonkey:
thx

PR #1510, opened 2026-05-19
test(jco): add missing component for testing clocks sleep

vados-cosmonic:


vados-cosmonic:
We already have this actually -- it's just not under cli but under clocks instead!

PR #1515, opened 2026-05-20
fix(bindgen): lift imported own handles via captureTable

andreiltd:
This fixes issue with p3 http handlers throwing:

```
ReferenceError: finalizationRegistry{N} is not defined
```

For `InterfaceType::Own` lift  `ensure_resource_table` only declares `finalizationRegistry{tid}` for guest defined resources and imported resources are tracked by `captureTable{rid}` instead. The pathc just mirrors the logic for `Instruction::HandleLift` from function_bindgen.rs file.

### DIFF
--- a/crates/js-component-bindgen/src/transpile_bindgen.rs
+++ b/crates/js-component-bindgen/src/transpile_bindgen.rs
@@ -5227,6 +5227,7 @@ pub fn gen_flat_lift_fn_js_expr(
             instantiator.add_intrinsic(Intrinsic::Lift(LiftIntrinsic::LiftFlatOwn));
             instantiator.add_intrinsic(Intrinsic::JsHelper(JsHelperIntrinsic::EmptyFunc));
             instantiator.add_intrinsic(Intrinsic::SymbolResourceHandle);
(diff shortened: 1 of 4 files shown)


crates/js-component-bindgen/src/transpile_bindgen.rs
```diff
@@ -5285,43 +5286,66 @@ pub fn gen_flat_lift_fn_js_expr(
...
-                                          value: handle,
-                                      }});
-                                      finalizationRegistry{tid}.register(resourceObj, handle, resourceObj);
-                                      {dtor_setup_js}
-                                      return resourceObj;
-                                  }}
-                                 "#
-                                );
+                                // Mirrors `Instruction::HandleLift` in `function_bindgen.rs`:
+                                let create_resource_fn_js = if !*imported {
```
vados-cosmonic: NIT: Can we reverse this check (as in `if imported { ... } else { ... }`)

vados-cosmonic (dismissed):
LTGM 🚀 

One readability nit but otherwise good to go!

PR #1516, opened 2026-05-20
feat(jco): add p3 fixtures test runners

andreiltd:
Make upstream p3 fixtures part of regression

### DIFF
--- a/packages/jco/test/p3/fixtures-cli.js
+++ b/packages/jco/test/p3/fixtures-cli.js
@@ -1,0 +1,207 @@
+import { spawn } from "node:child_process";
+import { once } from "node:events";
+import { mkdir, rm } from "node:fs/promises";
+import { createServer } from "node:http";
+import { basename, join } from "node:path";
+import { env as processEnv, execArgv, execPath } from "node:process";
+import { fileURLToPath } from "node:url";
+
+import { suite, test, assert, beforeAll, afterAll } from "vitest";
+
+import { P3_COMPONENT_FIXTURES_DIR } from "../common.js";
+import { setupAsyncTest } from "../helpers.js";
+
+const P3_CLI_FIXTURE_OUTPUT_DIR = fileURLToPath(new URL("../output/p3-cli-fixtures", import.meta.url));
+const CLI_RUNNER_PATH = fileURLToPath(new URL("./cli-runner.js", import.meta.url));
+
+const RUN_TIMEOUT_MS = 15_000;
+
+const P3_CLI_RUN_FIXTURES = [
+    { path: "cli/p3-cli-hello-stdout-post-return.wasm" },
+    { path: "cli/p3-cli-hello-stdout.wasm" },
+    { path: "cli/p3-cli-much-stdout.wasm", args: ["p3_cli_much_stdout.component", "hello, world\n", "1"] },
+    { path: "cli/p3-cli.wasm", args: ["p3_cli.component", "."] },
+    { path: "clocks/p3-clocks-sleep.wasm" },
+    { path: "fs/p3-file-write.wasm" },
+    { path: "fs/p3-filesystem-file-read-write.wasm" },
+    { path: "fs/p3-readdir.wasm" },
+    { path: "http/p3-http-outbound-request-get.wasm" },
+    { path: "http/p3-http-outbound-request-invalid-dnsname.wasm" },
+    { path: "http/p3-http-outbound-request-invalid-header.wasm" },
+    { path: "http/p3-http-outbound-request-invalid-port.wasm" },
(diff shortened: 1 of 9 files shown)


packages/jco/test/p3/fixtures-cli.js
```diff
@@ -0,0 +1,207 @@
...
+import { once } from "node:events";
+import { mkdir, rm } from "node:fs/promises";
+import { createServer } from "node:http";
+import { basename, join } from "node:path";
+import { env as processEnv, execArgv, execPath } from "node:process";
+import { fileURLToPath } from "node:url";
+
+import { suite, test, assert, beforeAll, afterAll } from "vitest";
+
+import { P3_COMPONENT_FIXTURES_DIR } from "../common.js";
```
vados-cosmonic: NIT: can we move the metadata and helper functions to another file? Ideally the test file should contain minimal setup and maybe the tests that should be run (i.e. P3_CLI_RUN_FIXTURES`) and the test harness.

packages/jco/test/p3/fixtures-cli.js
```diff
@@ -0,0 +1,207 @@
...
+
+    afterAll(async () => {
+        await server?.cleanup();
+    });
+
+    for (const fixture of P3_CLI_RUN_FIXTURES) {
+        const { path: relPath, failing } = fixture;
+        const title = `run ${relPath}${failing ? " (currently failing)" : ""}`;
+
+        suite.skipIf(failing)(title, () => {
```
vados-cosmonic: Any reason to make these their own suites? Seems like they could be individual in this suite

packages/jco/test/p3/fixtures-cli.js
```diff
@@ -0,0 +1,207 @@
...
+
+            beforeAll(async () => {
+                ({ esModuleHref, preopenDir, runnerArgs, cleanup } = await setupFixture(fixture));
+            });
+
+            afterAll(async () => {
+                await cleanup?.();
+            });
+
+            test("passes", async () => {
```
vados-cosmonic: could we make these concurrent? And if we collapse the suite, the individual test could have a more meaningful name (maybe the filename of the component in question)

packages/jco/test/p3/fixtures-cli.js
```diff
@@ -0,0 +1,207 @@
...
+                const { code, stdout, stderr } = await runNodeScript(
```
vados-cosmonic: We have something very similar to this -- `exec` in helpers.js -- could you merge with/reuse that? Feel free to refactor as necessary!

packages/jco/test/p3/fixtures-handler.js
```diff
@@ -0,0 +1,112 @@
...
+    }
+}
+
+suite("P3 handler fixtures", () => {
+    for (const fixture of P3_HANDLER_RUN_FIXTURES) {
+        const { path: relPath, title, failing, outbound, expect } = fixture;
+        const labelBase = title ? `${relPath} (${title})` : relPath;
+        const heading = `invoke ${labelBase}${failing ? " (currently failing)" : ""}`;
+
+        suite.skipIf(failing)(heading, () => {
```
vados-cosmonic: same note about component-test-as-suite here

packages/jco/test/p3/fixtures-handler.js
```diff
@@ -0,0 +1,112 @@
...
+    // TODO(tandr): http/p3-http-proxy.wasm
+];
+
+function outputDirFor(fixture) {
+    const { path: relPath, title } = fixture;
+    const base = relPath.replace(/[/\\]/g, "__").replace(/\.wasm$/, "");
+    return title ? `${base}__${title.replace(/[^A-Za-z0-9_-]/g, "-")}` : base;
+}
+
+async function setupFixture(fixture) {
```
vados-cosmonic: same note about moving stuff we can into shared modules for reuse. 

For example, the preamble here feels like it could be part of `setupAsyncTest` if it's not already?

packages/jco/test/p3/helpers/http.js
vados-cosmonic, line 1: NIT: how do you feel about naming this file `p3/util/runner-handler.js` or `p3/helpers/runner-handler.js`, or  `p3/helpers/http.js` instead? 

Looking at this one might think this was a file with tests in it, but it's actually helpers to be used from tests

vados-cosmonic (commented):
This mostly looks good, thanks for the work here @andreiltd  -- only some organizational stuff I'd like to get addressed if we can.

packages/jco/test/p3/fixtures-cli.js
andreiltd: Oh, missed that. Yeah exec works with only a small tweak for stdin.

PR #1519, opened 2026-05-20
fix(p3-shim): drop unconsumed body streams

andreiltd:
Fixes "http/p3-http-outbound-request-response-build.wasm" component.


vados-cosmonic (dismissed):
LGTM 🚀

PR #1547, opened 2026-05-25
test(jco): run remaining p3 http fixtures

andreiltd:
This adds last 3 p3 http handler based tests to regression:
- http/p3-http-middleware.wasm
- http/p3-http-middleware-with-chain.wasm
- http/p3-http-proxy.wasm


vados-cosmonic (approved):
LGTM 🚀

PR #1548, opened 2026-05-25
fix(bindgen): propagate host stream drop/cancel

andreiltd:
Propagate host stream drop/cancel in bindgen lowering, preserve pending host values across cancelled reads, and make tcp send/receive streams survive parent socket disposal.

This fixes p3-sockets-tcp-streams.wasm component test.

### DIFF
--- a/packages/preview3-shim/lib/nodejs/sockets/error.js
+++ b/packages/preview3-shim/lib/nodejs/sockets/error.js
@@ -20,6 +20,7 @@ export const ERROR_MAP = {
   ENONET: "remote-unreachable",
   ECONNREFUSED: "connection-refused",
   EPIPE: "connection-broken",
+  "write EPIPE": "connection-broken",
   ECONNRESET: "connection-reset",
   ECONNABORTED: "connection-aborted",
   EMSGSIZE: "datagram-too-large",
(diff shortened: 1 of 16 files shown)


vados-cosmonic (dismissed):
This is fantastic, and looks good to me -- thanks for getting these fixes in @andreiltd 

I have a couple minor nits:

- it would be nice if the commits were categorized correctly (changes to bindgen separate from changes to p3-shim)
- It would be nice if the `PendingValueQueue` was as separate intrinsic similar to other classes/etc that are used -- it looks to be defined in-line right now which does get optimzed out/hoisted, but would like to deal with that separately. 

That said, I'm fine to merge this and get it in, will leave it up to you to decide on the nits!

packages/preview3-shim/lib/nodejs/sockets/error.js
```diff
@@ -20,6 +20,7 @@ export const ERROR_MAP = {
   ENONET: "remote-unreachable",
   ECONNREFUSED: "connection-refused",
   EPIPE: "connection-broken",
+  "write EPIPE": "connection-broken",
```
vados-cosmonic, line 23: Oh this is another thing that I found weird -- forgot to add to the review, should this just be EPIPE?

andreiltd: Yeah, that is a workaround for when that raw `Error` crosses worker threads, `err.code` is dropped and only message is preserved so this matches string to wasi error string. The cleaner fix here would be to serialize `code` from worker. Let me have a look.

vados-cosmonic: yeah IMO I'm not too hung up on this, if you want to make a quick sort of... demuxer/translation fn for errors from functions (as strings and does whatever) then I'm happy with with that, if it's less complicated.

String encoding the errors coming out is just fine by me, especially if it lets us show more meaningful errors on the other side!

vados-cosmonic: Dismissing in favor of handling this in a follow up since this is internal

vados-cosmonic (dismissed):

vados-cosmonic (dismissed):

Issue #1534, opened 2026-05-21
WIT type aliases in some cases cause runtime panic in js-component-bindgen


jasoncg:
**tl;dr:**

Using WIT type aliases in some cases trigger runtime panics in `js-component-bindgen-1.19.2/src/function_bindgen.rs` `gen_dataview_set_and_check_fn_js_for_numeric_type(..)` apparently because these become `Type::Id(..)` which is not accounted for in the gen_dataview function. I worked around it in a local fork by passing `FunctionBindgen::resolve` into the function and looking up the type by id.

**Background:**

I'm using js-component-bindgen as part of a process transpiling WASM components to WASM modules at runtime client-side in the browser. An update to js-component-bindgen at some point in the past few months is triggering a runtime error:

> panicked at .cargo/registry/src/index.crates.io-1949cf8c6b5b557f/js-component-bindgen-1.19.2/src/function_bindgen.rs:3647:14:
> internal error: entered unreachable code: unsupported type [Id(Id { idx: 94 })] for canonical list lower

I made a local copy of js-component-bindgen to try to diagnose further, and determined that the panic happens for some (but not all) type aliases. 

For example:

```wit

    interface rb {
            type dt-idx = u64; 
            // various functions, records, and variants that use dt-idx
            variant data-type-variant {
                list-arr(dt-idx),
                // many other variants                
            }
    }

```
But oddly it doesn't seem to impact every type alias. I use a lot of type aliases and haven't been able to narrow down the critera that triggers this bug exactly, 

One workaround I tries is to change the alias types to records, which works but there are times when alias types make more sense so this isn't and ideal solution:

```wit

    interface rb {
        // type dt-idx = u64;
        record dt-idx {
            index: u64
        }
        
    }

```

I made a local fork of js-component-bindgen and this is sufficient to make it work for me, but I'm not sure if this would cause other side effects:

```rust
                // Determine what methods to use with a DataView when setting the data
                let (dataview_set_method, check_fn_intrinsic) =
                    gen_dataview_set_and_check_fn_js_for_numeric_type(element, &self.resolve);
```

```rust

fn gen_dataview_set_and_check_fn_js_for_numeric_type(
    ty: &Type,
    resolve: &Resolve,
) -> (&'static str, String) {
    let check_fn = Intrinsic::Conversion(ConversionIntrinsic::RequireValidNumericPrimitive).name();
    match ty {
        // Unsigned Integers
[...]

PR #1549, opened 2026-05-25
fix WIT type alias in js-component-bindgen (see issue #1534)

jasoncg:
This is a fix for issue #1534 where type aliases in some cases can cause a runtime panic. 

> panicked at .cargo/registry/src/index.crates.io-1949cf8c6b5b557f/js-component-bindgen-1.19.2/src/function_bindgen.rs:3647:14:
> internal error: entered unreachable code: unsupported type [Id(Id { idx: 94 })] for canonical list lower

I also included a test case that passes with the fix applied. 

The primary change to `gen_dataview_set_and_check_fn_js_for_numeric_type()` is that it now takes a wit_parser::resolve::Resolve so Type::Id(id) can be looked up. 

I'm not sure if it's actually possible to end up in this state, but one potential risk since `gen_dataview_set_and_check_fn_js_for_numeric_type()` is now called recursively is if a Type::Id is poorly defined such that it references itself it could cause an infinite loop.

### DIFF
--- a/crates/js-component-bindgen/src/function_bindgen.rs
+++ b/crates/js-component-bindgen/src/function_bindgen.rs
@@ -1254,7 +1254,7 @@ impl Bindgen for FunctionBindgen<'_> {
 
                 // Determine what methods to use with a DataView when setting the data
                 let (dataview_set_method, check_fn_intrinsic) =
-                    gen_dataview_set_and_check_fn_js_for_numeric_type(element);
+                    gen_dataview_set_and_check_fn_js_for_numeric_type(resolve, element);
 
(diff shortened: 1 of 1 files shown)


vados-cosmonic (dismissed):
LGTM 🚀 

Thanks @jasoncg for this fix -- for handling `Type::Id(..)`s I think we can recur there rather than panicking so early -- so I'll probably add that change on before merging, but even as-is I think the code is certainly an improvement.

vados-cosmonic (dismissed):

vados-cosmonic:
Resolves https://github.com/bytecodealliance/jco/issues/1534

PR #1551, opened 2026-05-26
chore(p3-shim): downgrade to 0.0.1 before first release

vados-cosmonic:


vados-cosmonic:
This is mostly to enable `0.1.0-rc.0` release(s) (we currently have checks that deny any release *prior* to the current version, so `0.1.0-rc.0` would be "before" `0.1.0` which is the current version), if we choose to do them, and to get an initial low-risk version out there via automation

PR #1560, opened 2026-05-26
fix(p3-shim): enforce outbound request content length

andreiltd:
This fixes request `content-length` handling. Request bodies are now checked against the length from header before accepting oversized guest writes and while transmitting bytes to the worker, so short bodies and oversized bodies reports `HTTP-request-body-size` correctly.

The bindgen side of the fix adds a stream read option that lets the host reject a pending write before copying it, which is required for `write_all` to return the full over size buffer as remaining. The preview3-shim change uses that hook for http request bodies.

### DIFF
--- a/packages/preview3-shim/lib/nodejs/http/request.js
+++ b/packages/preview3-shim/lib/nodejs/http/request.js
@@ -2,7 +2,12 @@ import { HttpError } from "./error.js";
 import { _fieldsLock, Fields } from "./fields.js";
 
 import { FutureReader, future } from "../future.js";
-import { StreamReader, readableByteStreamFromReader } from "../stream.js";
+import {
+  DEFAULT_BYTE_STREAM_CHUNK_SIZE,
+  StreamReader,
+  _byteChunk,
+  readableByteStreamFromReader,
+} from "../stream.js";
 
 const symbolDispose = Symbol.dispose || Symbol.for("dispose");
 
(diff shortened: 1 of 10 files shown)


packages/preview3-shim/lib/nodejs/http/request.js
```diff
@@ -476,6 +481,93 @@ export class Request {
...
+      if (typeof reader.close === "function") {
+        return reader.close();
+      }
+      return reader[symbolDispose]?.();
+    },
+  };
+}
+
+function readOpts(remaining) {
+  if (remaining <= 0n) {
```
vados-cosmonic, line 535: is remaining ever truly negative here? Should we maybe error/assert that isn't the case?

vados-cosmonic (dismissed):
LGTM 🚀

packages/preview3-shim/lib/nodejs/http/request.js
vados-cosmonic: Let's maybe address this in a follow up, if at all, I think this check is definitely not bad, and we don't strictly have to return a sort of... out of band error for the input being out of bounds.

andreiltd: Yeah, a bit defensive as this is computed from: `expectedLength - sent` so in theory it could be negative,

PR #1561, opened 2026-05-26
fix(bindgen): avoid spread calls for host stream buffers

andreiltd:
This fixes RangeError: Maximum call stack size exceeded error when running large post fixture test.


vados-cosmonic (approved):
Good find!

PR #1562, opened 2026-05-27
fix(p3-shim): serialize errors from worker

andreiltd:
This allows preserve error code when corssing worker boundary and allow correct error mapping on the main thread.


vados-cosmonic (approved):
LGTM 🚀

PR #1563, opened 2026-05-27
fix(p3-shim): resume socket on the stream write fail

andreiltd:
This an attempt to fix flaky socket stream test.

### DIFF
--- a/packages/preview3-shim/lib/nodejs/workers/tcp-worker.js
+++ b/packages/preview3-shim/lib/nodejs/workers/tcp-worker.js
@@ -217,8 +217,12 @@ async function handleTcpReceive({ socketId, stream }) {
           } catch {
             settle();
             return;
+          } finally {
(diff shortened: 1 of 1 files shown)


packages/preview3-shim/lib/nodejs/workers/tcp-worker.js
```diff
@@ -217,8 +217,9 @@ async function handleTcpReceive({ socketId, stream }) {
           } catch {
             settle();
             return;
+          } finally {
+            tcp.resume();
```
vados-cosmonic, line 224: maybe we should add a note here as to why this was necessary? If we're getting an error here, what kind of error could it be, etc etc

andreiltd: Yup, added a comment.

vados-cosmonic (approved):

PR #1574, opened 2026-05-29
fix(p2-shim): coerce browser file read bounds to numbers

### LINKED ISSUE #1573: Browser preview2-shim: `Descriptor.read` throws on canonical BigInt arguments
# Browser preview2-shim: `Descriptor.read` throws on canonical BigInt arguments

## Summary

In `@bytecodealliance/preview2-shim`, the browser implementation of
`wasi:filesystem.types.Descriptor.read` passes its `length` and `offset`
arguments straight to `Uint8Array.prototype.slice`. The WIT signature
declares both as `filesize` (`u64`), which the canonical-ABI binding
delivers as `BigInt`, so `slice` always throws:

```
TypeError: Cannot convert a BigInt value to a number
    at Uint8Array.slice (<anonymous>)
    at Descriptor.read (preview2-shim/lib/browser/filesystem.js:191)
```

Any component that performs `read` against a file in the browser shim
fails on the first call. The Node-side filesystem implementation in
this same package is unaffected because its `read` goes through
`fs.read` (which accepts BigInt position).

## Affected version / location

- Package: `@bytecodealliance/preview2-shim` (verified on `0.18.0`)
- File: `packages/preview2-shim/lib/browser/filesystem.js`
- Function: `Descriptor.read(length, offset)` — currently lines 189-192
  on `main` (commit `6c585ac1b0b252e5ec8987a0d7d8f3ffda0cd076`)

## Minimal reproducer

`repro.mjs`:

```js
import { _setFileData, preopens } from
  './node_modules/@bytecodealliance/preview2-shim/lib/browser/filesystem.js';

_setFileData({
  dir: {
    'test.txt': { source: new Uint8Array([0x48, 0x65, 0x6c, 0x6c, 0x6f]) },
  },
});

const [[root]] = [preopens.getDirectories()];
const fd = root.openAt({}, 'test.txt', {}, {});

try {
  const [bytes, eof] = fd.read(5n, 0n);
  console.log('OK', new TextDecoder().decode(bytes), 'eof=' + eof);
} catch (e) {
  console.error('THROWS', String(e));
}
```

(The relative `./node_modules/...` import is used so the file resolves
to the browser implementation under Node; the package's `exports`
field otherwise routes specifier imports to the Node implementation.)

Steps:

```
mkdir repro && cd repro
npm init -y
npm install @bytecodealliance/preview2-shim@0.18.0
# save repro.mjs alongside package.json
node repro.mjs
```

### Before the fix

```
$ node repro.mjs
THROWS TypeError: Cannot convert a BigInt value to a number
exit 1
```

### After the fix

```
$ node repro.mjs
OK Hello eof=true
exit 0
```

## Root cause

`Descriptor.read` in the browser implementation:

```js
read(length, offset) {
  const source = getSource(this.#entry);
  return [source.slice(offset, offset + length),
          offset + length >= source.byteLength];
}
```

`Uint8Array.prototype.slice` requires `Number` arguments and throws
on `BigInt`. The wasi-filesystem WIT declares the arguments as
`filesize = u64`, so the canonical-ABI binding passes them as
`BigInt`. The call therefore always throws.

## Proposed fix

Coerce both arguments to `Number` once at entry:

```js
read(length, offset) {
  const source = getSource(this.#entry);
  const off = typeof offset === "bigint" ? Number(offset) : offset;
  const len = typeof length === "bigint" ? Number(length) : length;
  return [source.slice(off, off + len),
          off + len >= source.byteLength];
}
```

This:

- accepts the canonical-ABI `BigInt` arguments without throwing,
- preserves behavior for any caller that still happens to pass
  `Number` (`typeof` guard makes the coercion idempotent),
- keeps `source.byteLength` and the returned EOF boolean as
  `Number`-typed (matching the existing function's already-Number
  result tuple).

Safe up to `Number.MAX_SAFE_INTEGER` bytes (~9 PB), well above any
practical browser use.

Patch attached as `patch.diff`.

## Discovery context

Found while embedding a WASI component (a WebAssembly-native JVM)
in a browser via `jco transpile` + the `preview2-shim` browser
builds. The component reads a ~15 MB resource file on startup
through `wasi-filesystem.types.read`; every browser run failed
immediately on the first `read`. The same artifact runs without
modification under `jco run` (which uses the Node-side filesystem
implementation, where this code path is not exercised).

## Related code

`packages/preview2-shim/lib/browser/filesystem.js` `Descriptor.write`
at line ~194 also takes a `filesize` `offset` but currently only
checks `offset !== 0` and stores the buffer; it doesn't index with
the value, so the same BigInt issue doesn't manifest there. No
change proposed.

zacharywhitley:
Fixes #1573

wasi-filesystem `read` takes `filesize` (`u64`) arguments, which the canonical-ABI binding passes as `BigInt`. `Uint8Array.slice` rejects `BigInt` and throws `Cannot convert a BigInt value to a number` on the first read against any file in the browser implementation.

Coerce both arguments to `Number` on entry. Safe up to `Number.MAX_SAFE_INTEGER` bytes.

Reproducer and before/after evidence are in #1573.

### DIFF
--- a/packages/preview2-shim/lib/browser/filesystem.js
+++ b/packages/preview2-shim/lib/browser/filesystem.js
@@ -43,6 +43,22 @@ const timeZero = {
   nanoseconds: 0,
 };
 
+/** Coerce the given object to a safe integer */
+function coerceToSafeIntegerNumber(obj) {
+  let n;
+  if (typeof obj === "number") {
+    n = obj;
+  } else if (typeof obj == "bigint") {
+    n = Number(obj);
+  } else {
+    throw new TypeError(`unexpected non-numeric type: ${obj}`);
+  }
+  if (n > Number.MAX_SAFE_INTEGER) {
+    throw new TypeError(`excessively large number: ${n}`);
+  }
+  return n;
+}
+
 function getChildEntry(parentEntry, subpath, openFlags) {
   if (subpath === "." && _rootPreopen && descriptorGetEntry(_rootPreopen[0]) === parentEntry) {
     subpath = _getCwd();
@@ -188,8 +204,8 @@ class Descriptor {
 
   read(length, offset) {
     const source = getSource(this.#entry);
-    const off = typeof offset === "bigint" ? Number(offset) : offset;
-    const len = typeof length === "bigint" ? Number(length) : length;
+    const off = coerceToSafeIntegerNumber(offset);
+    const len = coerceToSafeIntegerNumber(length);
     return [source.slice(off, off + len), off + len >= source.byteLength];
   }
 
--- a/packages/preview2-shim/lib/browser/filesystem.js
+++ b/packages/preview2-shim/lib/browser/filesystem.js
@@ -188,7 +188,9 @@ class Descriptor {
 
   read(length, offset) {
     const source = getSource(this.#entry);
-    return [source.slice(offset, offset + length), offset + length >= source.byteLength];
+    const off = typeof offset === "bigint" ? Number(offset) : offset;
+    const len = typeof length === "bigint" ? Number(length) : length;
+    return [source.slice(off, off + len), off + len >= source.byteLength];
   }
 
   write(buffer, offset) {


vados-cosmonic (dismissed):
LGTM 🚀 

Thanks @zacharywhitley for getting this fixed -- would you mind sharing an outline of what the component was doing (or even code for the component itself?) would love to get it added to a regression test.

We're going to be expanding the coverage of our test suite (currently we only build Rust components for use in testing), but it would be nice to have JVM component code and some ooling to build and use during test.

packages/preview2-shim/lib/browser/filesystem.js
```diff
@@ -188,7 +188,9 @@ class Descriptor {
 
   read(length, offset) {
     const source = getSource(this.#entry);
-    return [source.slice(offset, offset + length), offset + length >= source.byteLength];
+    const off = typeof offset === "bigint" ? Number(offset) : offset;
```
vados-cosmonic: The line after this is probably a good time to check and error loudly if we're over `Number.MAX_SAFE_INTEGER`.

Separately, rather than checking for `bigint` in particular we should check for `number` and do the conversion for non-`number` types (also happy with checking the cases systematically with a `switch` or `if`).

vados-cosmonic (approved):

zacharywhitley:
Glad this was helpful. Give me a bit and i'll get back to you with what exactly surfaced this but it was when I was working on a browser wasi 1/2/3 polyfill if you wanted to take a look at that at https://github.com/tegmentum/wasi-polyfill

What exactly did you have in mind for "JVM component code"? Happy to help

vados-cosmonic:
Thanks for the link to your polyfill -- it looks awesome! We've always had only experimental browser support in Jco so your help here improving our impl is really appreciated.

> What exactly did you have in mind for "JVM component code"? Happy to help

Ah so I meant whatever code you were trying to run -- I read this in your reproducer:

> Found while embedding a WASI component (a WebAssembly-native JVM)
in a browser via jco transpile + the preview2-shim browser
builds. The component reads a ~15 MB resource file on startup
through wasi-filesystem.types.read; every browser run failed
immediately on the first read. The same artifact runs without
modification under jco run (which uses the Node-side filesystem
implementation, where this code path is not exercised).

Do you have a JVM that's WebAssembly Native? I don't think I've seen a popular one up until now with component model support so very curious what toolchain you used!

PR #1586, opened 2026-06-02
fix(bindgen): restore storageLen budget after a list field in records and tuples

### LINKED ISSUE #1585: jco transpile 1.20: record with a list field before a fixed-width field fails to lift (storageLen not restored)
### Environment
- `@bytecodealliance/jco` 1.20.0 (`js-component-bindgen` 1.19.x)
- `jco transpile --instantiation --optimize`
- Node 20 / 22 / 25

### Summary
When a component export returns (or streams) a `record` whose **first field is a `list`** (e.g. `list<u8>`) immediately followed by a **fixed-width field** (e.g. `u32`), the transpiled JS throws while lifting the record:

```
Error: insufficient storage ([0] bytes) for lift (u32 requires 4 bytes)
    at _liftFlatU32 (<component>.js)
    at _liftFlatRecordInner (<component>.js)
    at ManagedBuffer.read (<component>.js)
```

It reproduces both for a plain `-> record` return and for a `record` carried over a `stream<T>`; it's value-marshaling, independent of async mode.

### Minimal WIT
```wit
package example:liftbug;

world w {
  record blob {
    data: list<u8>,
    count: u32,
    label: string,
  }
  export get: func() -> blob;
}
```
Transpile a component implementing this and call `get()`; lifting fails at the `count: u32` field.

### Root cause (in the emitted JS)
In the generated `_liftFlatList` helper, the inner `readValuesAndReset`:
- sets `ctx.storageLen` to the list's byte length and reads the elements (decrementing `ctx.storageLen` toward 0), then
- on return restores `ctx.storagePtr = originalPtr` **but never restores `ctx.storageLen`**.

`_liftFlatRecordInner` advances `storagePtr` per field but does **not** save/restore `storageLen` per field. So the field *after* the list sees `ctx.storageLen === 0`, and `_liftFlatU32`'s guard `if (ctx.storageLen !== undefined && ctx.storageLen < 4) throw …` fires. Any fixed-width lift following a list in the same record hits this.

### Standalone repro (no wasm — reproduces the emitted lift logic)
```js
function _liftFlatU32(ctx) {
  if (ctx.storageLen !== undefined && ctx.storageLen < 4)
    throw new Error(`insufficient storage ([${ctx.storageLen}] bytes) for lift (u32 requires 4 bytes)`);
  const v = new DataView(ctx.memory.buffer).getUint32(ctx.storagePtr, true);
  ctx.storagePtr += 4;
  if (ctx.storageLen !== undefined) ctx.storageLen -= 4;
  return [v, ctx];
}
function _liftFlatU8(ctx) {
  const v = new DataView(ctx.memory.buffer).getUint8(ctx.storagePtr);
  ctx.storagePtr += 1;
  if (ctx.storageLen !== undefined) ctx.storageLen -= 1;
  return [v, ctx];
}
function _liftFlatList({ elemLiftFn, elemSize32 }) {
  const readValuesAndReset = (ctx, originalPtr, dataPtr, len) => {
    for (let i = 0; i < len; i++) { ctx.storagePtr = dataPtr + i * elemSize32; const [, n] = elemLiftFn(ctx); ctx = n; }
    if (originalPtr !== null) ctx.storagePtr = originalPtr; // <-- storageLen NOT restored
    return [[], ctx];
  };
  return (ctx) => {
    ctx.storageLen = 8;
    const [dataPtr, c1] = _liftFlatU32(ctx);
    const [len, c2] = _liftFlatU32(c1);
    let c = c2; const originalPtr = c.storagePtr;
    c.storagePtr = dataPtr; c.storageLen = len * elemSize32;
    return readValuesAndReset(c, originalPtr, dataPtr, len);
  };
}
function _liftFlatRecord(fieldMetas) {
  return (ctx) => { const res = {}; for (const [k, fn] of fieldMetas) { const [v, n] = fn(ctx); res[k] = v; ctx = n; } return [res, ctx]; };
}

const memory = new WebAssembly.Memory({ initial: 1 });
const dv = new DataView(memory.buffer);
dv.setUint32(0, 64, true);   // list data ptr
dv.setUint32(4, 5, true);    // list len
dv.setUint32(8, 42, true);   // count: u32
for (let i = 0; i < 5; i++) dv.setUint8(64 + i, i + 1);

const lift = _liftFlatRecord([
  ['data', _liftFlatList({ elemLiftFn: _liftFlatU8, elemSize32: 1 })],
  ['count', _liftFlatU32],
]);
try { lift({ memory, storagePtr: 0, storageLen: 20 }); console.log('unexpected success'); }
catch (e) { console.log('repro:', e.message); }
// -> repro: insufficient storage ([0] bytes) for lift (u32 requires 4 bytes)
```

### Suggested fix
Save `storageLen` before reading the list elements and restore it alongside the existing `storagePtr` restore in `readValuesAndReset`:
```js
const savedLen = ctx.storageLen;
// … read elements …
if (originalPtr !== null) { ctx.storagePtr = originalPtr; ctx.storageLen = savedLen; }
```
Equivalently, have `_liftFlatRecordInner` save/restore `storageLen` per field. Happy to send a PR if that helps.

temporaryfix:
## Summary

Fixes #1585 — transpiling a component that returns or streams a `record`/`tuple` whose first field is a `list` immediately followed by a fixed-width field throws while lifting the field after the list.

`_liftFlatList`'s `readValuesAndReset` sets `ctx.storageLen` to the element-buffer length while reading the out-of-line elements and restores `ctx.storagePtr` but **not** `ctx.storageLen`. `_liftFlatRecord` / `_liftFlatTuple` already normalize `storagePtr` per field via `ctx.storagePtr = Math.max(ctx.storagePtr, fieldPtr + size32)` but don't track `storageLen`, so the field *after* a list sees `storageLen === 0` and its guard throws:

```
insufficient storage ([0] bytes) for lift (u32 requires 4 bytes)
```

## Fix

Mirror the existing per-field `storagePtr` normalization for `storageLen`: capture the remaining budget before each field/element and set it to `fieldLen - size32` afterwards. A field occupies exactly `size32` bytes of the composite's flat storage regardless of how its own lift fn manipulates `storageLen` internally. This:

- is a **no-op** for fields that already account for their own bytes (primitives do `storageLen -= size`);
- correctly handles a **zero-length leading list** — a "restore to element-buffer length" variant would leave `storageLen` at `0` for an empty list and still throw;
- needs no magic numbers, and covers tuples as well as records.

## Test plan

- New `lift-list-before-fixed` test component (record + tuple with a list before a fixed-width field, plus an empty-list case) and a P3 regression test `test/p3/lift-list-before-fixed.js`.
- Before the fix the same component throws on all three exports; after, all lift correctly.
- Full P3 suite (239 passed) and the classic runtime suite (15 passed) stay green; `cargo fmt` / `clippy` and `oxlint` / `oxfmt` are clean.

## Note

While in `lift.rs` I noticed a separate, likely-unrelated typo in the UTF-8 string lift (~line 689): `ctx.storagelen -= 8` (lowercase `l`) where the UTF-16 path uses `ctx.storageLen`. It's guard-weakening rather than fatal, so I left it out of this PR — happy to address separately if useful.

### DIFF
--- a/crates/js-component-bindgen/src/intrinsics/lift.rs
+++ b/crates/js-component-bindgen/src/intrinsics/lift.rs
@@ -745,12 +745,26 @@ impl LiftIntrinsic {
                                     fieldPtr = ctx.storagePtr;
                                 }}
 
+                                // A field occupies exactly size32 bytes of the record's
+                                // flat storage. Capture the remaining storage budget before
+                                // lifting the field and restore it afterwards: a field's own
+                                // lift fn may repurpose storageLen internally (e.g. a list
+                                // sets it to the element-buffer length while reading
+                                // out-of-line data and never restores it), which would
+                                // otherwise corrupt the budget the next field sees.
+                                // See https://github.com/bytecodealliance/jco/issues/1585.
+                                let fieldLen;
+                                if (ctx.storageLen !== undefined) {{ fieldLen = ctx.storageLen; }}
+
                                 let [val, newCtx] = liftFn(ctx);
                                 res[key] = val;
                                 ctx = newCtx;
 
                                 if (fieldPtr !== undefined) {{
                                     ctx.storagePtr = Math.max(ctx.storagePtr, fieldPtr + size32);
+                                }}
+                                if (fieldLen !== undefined) {{
+                                    ctx.storageLen = fieldLen - size32;
                                 }}
                             }}
 
@@ -992,12 +1006,24 @@ impl LiftIntrinsic {
                                     elemPtr = ctx.storagePtr;
                                 }}
 
+                                // As in _liftFlatRecord: an element occupies exactly size32
+                                // bytes of the tuple's flat storage, so capture and restore
+                                // the storage budget around the element lift to stop a
+                                // field's internal storageLen use (e.g. lists) leaking into
+                                // the next element.
+                                // See https://github.com/bytecodealliance/jco/issues/1585.
+                                let elemLen;
+                                if (ctx.storageLen !== undefined) {{ elemLen = ctx.storageLen; }}
+
                                 const [newValue, newCtx] = liftFn(ctx);
                                 val.push(newValue);
                                 ctx = newCtx;
 
                                 if (elemPtr !== undefined) {{
                                     ctx.storagePtr = Math.max(ctx.storagePtr, elemPtr + size32);
+                                }}
+                                if (elemLen !== undefined) {{
+                                    ctx.storageLen = elemLen - size32;
                                 }}
                             }}
 
--- a/crates/test-components/src/bin/lift_list_before_fixed.rs
+++ b/crates/test-components/src/bin/lift_list_before_fixed.rs
@@ -1,0 +1,42 @@
+mod bindings {
+    use super::Component;
+    wit_bindgen::generate!({
+        world: "lift-list-before-fixed",
+    });
+    export!(Component);
+}
+
+use bindings::Blob;
+
+struct Component;
+
+impl bindings::Guest for Component {
+    /// A record whose first field is a list, followed by a fixed-width `u32`
+    /// and a `string`. Lifting `count` used to fail (jco#1585).
+    async fn get_blob() -> Blob {
+        Blob {
+            data: vec![1, 2, 3, 4, 5],
+            count: 42,
+            label: "hello".into(),
+        }
+    }
+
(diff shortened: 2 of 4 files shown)


vados-cosmonic (approved):
LGTM 🚀 

Thanks for the contribution @temporaryfix  -- this looks good to go and should be protecetd going forward given the new test, will merge once tests are 🟢 .

PR #1597, opened 2026-06-05
fix(p3-shim): align filesystem tcp and http edge cases

andreiltd:
This patch fixes some of the wasi semantics that wasi-testsuite tests. This make the following tests from wasi-testsuite pass:

 - filesystem-hard-links
 - filesystem-is-same-object
 - filesystem-metadata-hash
 - filesystem-mkdir-rmdir
 - filesystem-open-errors
 - filesystem-rename
 - filesystem-unlink-errors
 - http-fields
 - http-request
 - http-response
 - sockets-tcp-receive
 - sockets-udp-send


vados-cosmonic (approved):
LGTM 🚀 

Looks like we are getting some useful edge case testing out of wasi-testsuite!

PR #1598, opened 2026-06-05
fix(bindgen): lower async flag results with js field names

andreiltd:
Normalize flag names when lowering so async result accepts js camel fields and not only raw wit kebab case names. This was found when using name like `mutate-directory` from wasi-testsuite.

### DIFF
--- a/crates/js-component-bindgen/src/intrinsics/lower.rs
+++ b/crates/js-component-bindgen/src/intrinsics/lower.rs
@@ -846,10 +846,14 @@ impl LowerIntrinsic {
 
                             let flagObj = ctx.vals[0];
                             let flagValue = 0;
-                            for (const [idx, name] of names.entries()) {{
-                                if (flagObj[name] === true) {{
-                                    flagValue |= 1 << idx;
+                            if (typeof flagObj === 'object' && flagObj !== null) {{
+                                for (const [idx, name] of names.entries()) {{
+                                    if (flagObj[name] === true) {{
+                                        flagValue |= 1 << idx;
+                                    }}
                                 }}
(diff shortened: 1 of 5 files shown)


crates/js-component-bindgen/src/intrinsics/lower.rs
```diff
@@ -839,17 +839,22 @@ impl LowerIntrinsic {
                 output.push_str(&format!(r#"
                     function {lower_flat_flags_fn}(meta) {{
                         const {{ names, size32, align32, intSizeBytes }} = meta;
+                        const jsNames = names.map(name => name.replace(/-([a-z])/g, (_, c) => c.toUpperCase()));
```
vados-cosmonic: what do you think about doing this on the Rust side?

vados-cosmonic (dismissed):
LGTM 🚀 

Just a small NIT about whether we want to pre-compute the JS names on the Rust side (IIRC we have `heck` pulled in) and have it just be static data

crates/js-component-bindgen/src/intrinsics/lower.rs
andreiltd: Yeah, sounds like a good idea!

vados-cosmonic: yeah at some point we'll optimize bundle size... but not today :)

vados-cosmonic (approved):

crates/js-component-bindgen/src/intrinsics/lower.rs
andreiltd: Much cleaner :chefkiss:

PR #1609, opened 2026-06-08
test(jco): add component model WAT tests

vados-cosmonic:


vados-cosmonic:
I'm going to merge this in favor of getting this machinery added as large refactors may be coming to bindgen, to avoid drift/difficulty merging this PR in the future.

So this PR adds the tests, but *does not* make every fix necessary to bindgen necessary to get all the tested corner cases passing.

PR #1610, opened 2026-06-08
fix(bindgen): host-side async import param lifting

### LINKED ISSUE #1601: transpile/jspi: async import taking a `future` or `stream` param leads to ReferenceError
This example uses a `future` but the exact same issue appears with `stream`s.

`repro.wit`
```wit
package repro:import-async-param;
world root {
  import take-future: async func(f: future<u32>);
}
```

```console
$ wasm-tools component embed --dummy --world root repro.wit | wasm-tools component new -o repro.wasm
$ jco transpile repro.wasm --async-mode jspi -o out
$ grep futureResult0 out/repro.js
        fn: () => takeFuture(futureResult0),
```

Note that there is exactly one line matching `futureResult0`; it is used without being defined. In practice this results in `ReferenceError: futureResult0 is not defined` at runtime.

vados-cosmonic:
Resolves https://github.com/bytecodealliance/jco/issues/1601

### DIFF
--- a/crates/js-component-bindgen/src/function_bindgen.rs
+++ b/crates/js-component-bindgen/src/function_bindgen.rs
@@ -1511,7 +1511,7 @@ impl Bindgen for FunctionBindgen<'_> {
                     prefix = self.tracing_prefix,
                 );
 
-                // Write out whether the caller was host provided
+                // Write out whether the callee was host provided
                 // (if we're calling into wasm then we know it was not)
                 uwriteln!(self.src, "const hostProvided = false;");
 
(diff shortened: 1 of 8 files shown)


vados-cosmonic:
Hey @lann I'm bit delayed, but this PR here should fix the issue you ran into -- feel free to peruse it.

I tried to also capture [the pattern of your example](https://github.com/bytecodealliance/jco/pull/1610/changes#diff-2b7a7774f132e68b177f67eb2c977701a868df19a73de6e5c437fbdb58ed587b) as a test case, feel free to let me know if I missed anything/the spirit of the functionality you were going for (though admittedly narrow -- more tests are planned but I'll leave that for a follow up).

I'll merge this and get it released once tests are :green_circle: but will have some quick follow ups this week.

PR #1611, opened 2026-06-08
fix(bindgen): set typedArray in js lowering functions

lann:
This diff is copied nearly verbatim from the trampoline code: 

https://github.com/bytecodealliance/jco/blob/bf5e5abf55cdb4d59cda1334c106ae27d0efebd2/crates/js-component-bindgen/src/transpile_bindgen.rs#L1372
https://github.com/bytecodealliance/jco/blob/bf5e5abf55cdb4d59cda1334c106ae27d0efebd2/crates/js-component-bindgen/src/transpile_bindgen.rs#L1385
https://github.com/bytecodealliance/jco/blob/bf5e5abf55cdb4d59cda1334c106ae27d0efebd2/crates/js-component-bindgen/src/transpile_bindgen.rs#L1420
https://github.com/bytecodealliance/jco/blob/bf5e5abf55cdb4d59cda1334c106ae27d0efebd2/crates/js-component-bindgen/src/transpile_bindgen.rs#L1437

Note that for my use case I could only trigger this bug after applying the (possibly not entirely correct) patch described [here](https://github.com/bytecodealliance/jco/issues/1601#issuecomment-4641094403) but this fix seems pretty straightforward.

### DIFF
--- a/crates/js-component-bindgen/src/transpile_bindgen.rs
+++ b/crates/js-component-bindgen/src/transpile_bindgen.rs
@@ -6064,6 +6064,7 @@ pub fn gen_flat_lower_fn_js_expr(
                 is_none_type,
                 is_numeric_type,
                 is_async_value,
+                typed_array_js,
             ) = match payload {
                 None => (
                     0,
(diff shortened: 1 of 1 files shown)


vados-cosmonic (approved):
LGTM 🚀 

Thanks @lann!

PR #1612, opened 2026-06-09
fix(p3-shim): handle sub milisecond waits

andreiltd:
Add recheck `monotonicClock.now()` after each wake in wait functions and continue waiting until the target is reached.


vados-cosmonic (approved):
LGTM 🚀 

Thanks for this fix -- I know this code has been a source of issues in p2-shim, glad to get this under more test coverage.

PR #1646, opened 2026-06-15
Fix js component bindgen helper emission

### LINKED ISSUE #1643: js-component-bindgen omits _liftFlatVariant for enum-only import lifting
## Summary

A component can import a host function whose parameter is a WIT `enum`.

Example:

```wit
interface host {
  enum level {
    debug,
    info,
  }

  log: func(lvl: level);
}
```

When `js-component-bindgen` generates JavaScript for this shape, it emits a helper named `_liftFlatEnum`.

That helper does not decode the enum by itself. Instead, it calls the shared variant helper `_liftFlatVariant` and then unwraps the variant tag into a plain JavaScript string.

The generated code looks like this:

```js
function _liftFlatEnum(meta) {
  meta.isEnum = true;
  const f = _liftFlatVariant(meta);
  // ...
}
```

The problem is that, for a small enum-only world, the generated file contains `_liftFlatEnum` but does not contain `_liftFlatVariant`.

So the generated JavaScript refers to a helper that was never emitted. At runtime, calling the imported function can fail with:

```text
ReferenceError: _liftFlatVariant is not defined
```

This does not happen when the same world also contains an `option`, `result`, or ordinary `variant`, because those types already cause `_liftFlatVariant` to be emitted.

That is why the bug is easiest to reproduce in a small world where the enum is the only variant-shaped type.

## Minimal Reproducer

```js
import { componentEmbed, componentNew, transpile } from "@bytecodealliance/jco";

const witSource = `
package test:enum-only;

interface host {
  enum level {
    debug,
    info,
  }

  log: func(lvl: level);
}

world w {
  import host;
  export run: func();
}
`;

const coreModule = await componentEmbed({
  witSource,
  dummy: true,
});

const component = await componentNew(coreModule);
const output = await transpile(component, { name: "enum-only" });

const generatedJs = new TextDecoder().decode(output.files["enum-only.js"]);

console.log({
  hasEnum: generatedJs.includes("function _liftFlatEnum"),
  hasVariant: generatedJs.includes("function _liftFlatVariant"),
  hasU8: generatedJs.includes("function _liftFlatU8"),
});
```

## Actual Output

With `jco 1.24.0` / `js-component-bindgen 2.0.2`:

```js
{
  hasEnum: true,
  hasVariant: false,
  hasU8: false
}
```

## Expected Output

If `_liftFlatEnum(...)` is emitted, `_liftFlatVariant(...)` should also be emitted.

The missing dependency is not limited to `_liftFlatU8`.

`_liftFlatVariant(...)` chooses a discriminant lifter based on the number of cases:

```js
_liftFlatU8   // fewer than 256 cases
_liftFlatU16  // fewer than 65536 cases
_liftFlatU32  // larger variants/enums
```

In this small reproducer, `_liftFlatU8` is the scalar helper that would be used. But the core missing dependency is `_liftFlatVariant`; once that helper is emitted, the existing variant dependency path should also emit the needed scalar lifters.

## Bug Trace

The generated enum helper delegates to `_liftFlatVariant`:

```js
function _liftFlatEnum(meta) {
  meta.isEnum = true;
  const f = _liftFlatVariant(meta);
  // ...
}
```

In `js-component-bindgen`, `LiftFlatOption` and `LiftFlatResult` already cause `LiftFlatVariant` to be emitted.

The current dependency expansion appears to check:

```text
LiftFlatResult -> LiftFlatVariant
LiftFlatOption -> LiftFlatVariant
LiftFlatOption -> LiftFlatVariant
```

The third check is duplicated.

Since `_liftFlatEnum(...)` also delegates to `_liftFlatVariant(...)`, the intended dependency seems to be:

```text
LiftFlatResult -> LiftFlatVariant
LiftFlatOption -> LiftFlatVariant
LiftFlatEnum   -> LiftFlatVariant
```

So the minimal source-level fix appears to be replacing the duplicated `LiftFlatOption` dependency check with `LiftFlatEnum`.

## Notes

This reproducer intentionally avoids a `string` parameter so the issue stays focused on enum/variant helper emission.

This reproducer uses `jco` only as the public API for invoking `js-component-bindgen`.

The actual generated-code issue appears to be in `js-component-bindgen` intrinsic dependency tracking.

### LINKED ISSUE #1644: js-component-bindgen emits invalid const reassignment in _liftFlatStringUTF16
## Summary

`js-component-bindgen` can generate invalid JavaScript for a component that imports a host function with a `string` parameter.

This is not related to enums or variants. A plain string-only WIT world is enough.

The generated UTF-16 string lifting helper declares `offset` as `const`, then immediately tries to reassign it after bigint normalization:

```js
const offset = ctx.params[0];
if (typeof offset === 'bigint') { offset = Number(offset); }
```

That is invalid JavaScript and can fail during parsing/building with errors like:

```text
cannot reassign to a variable declared with `const`
```

## Minimal Reproducer

```js
import { componentEmbed, componentNew, transpile } from "@bytecodealliance/jco";

const witSource = `
package test:string-only;

interface host {
  log: func(msg: string);
}

world w {
  import host;
  export run: func();
}
`;

const component = await componentNew(
  await componentEmbed({
    witSource,
    dummy: true,
  }),
);

const { files } = await transpile(component, {
  name: "string-only",
});

const generatedJs = new TextDecoder().decode(files["string-only.js"]);

console.log({
  hasLetOffset: generatedJs.includes("let offset = ctx.params[0];"),
  hasConstOffset: generatedJs.includes("const offset = ctx.params[0];"),
});
```

## Actual Output

With `jco 1.24.0` / `js-component-bindgen 2.0.2`, the generated JS contains both forms:

```js
{
  hasLetOffset: true,
  hasConstOffset: true
}
```

The `let offset` comes from the UTF-8 string helper.

The invalid `const offset` comes from the UTF-16 string helper.

The problematic generated code is:

```js
const offset = ctx.params[0];
if (typeof offset === 'bigint') { offset = Number(offset); }
```

## Expected Output

The UTF-16 string lifting helper should match the UTF-8 helper and use a mutable binding:

```js
let offset = ctx.params[0];
if (typeof offset === 'bigint') { offset = Number(offset); }
```

Then bigint normalization can safely reassign `offset`.

## Build Error Seen In Practice

A bundler can reject the generated module during parsing/building:

```text
cannot reassign to a variable declared with `const`

const offset = ctx.params[0];
if (typeof offset === 'bigint') { offset = Number(offset); }
                                      ^^^^^^
```

## Bug Trace

The issue appears to be in the generated `_liftFlatStringUTF16(...)` helper.

`_liftFlatStringUTF8(...)` already emits:

```js
let offset = ctx.params[0];
```

`_liftFlatStringUTF16(...)` emits:

```js
const offset = ctx.params[0];
```

but then performs the same reassignment:

```js
offset = Number(offset);
```

So the minimal source-level fix is to change the UTF-16 direct-param branch from `const offset` to `let offset`.

## Regression Test Shape

A focused regression test can transpile the string-only WIT world above and assert:

```js
assert.isOk(generatedJs.includes("let offset = ctx.params[0];"));
assert.isFalse(generatedJs.includes("const offset = ctx.params[0];"));
```

I verified locally that this test fails when the generator still emits `const offset`, and passes after changing the UTF-16 helper to emit `let offset`.

## Notes

This is separate from enum/variant helper dependency tracking.

This reproducer uses `jco` only as the public API. The generated-code issue appears to be in `js-component-bindgen` string lifting code.

### LINKED ISSUE #1645: js-component-bindgen emits Stream class that references undeclared symbolRscRep
## Summary

`js-component-bindgen` can generate a `Stream` class that references `symbolRscRep` without also emitting the `symbolRscRep` declaration.

A small WIT world that only returns a `stream<T>` is enough.

The generated `Stream` constructor contains:

```js
this[symbolRscRep] = args.globalRep;
```

but the generated file does not contain:

```js
const symbolRscRep = Symbol.for("cabiRep");
```

So constructing the returned stream can fail with:

```text
ReferenceError: symbolRscRep is not defined
```

## Minimal Reproducer

```js
import { componentEmbed, componentNew, transpile } from "@bytecodealliance/jco";

const witSource = `
package test:stream-lift-only;

world w {
  export get-stream: func() -> stream<u8>;
}
`;

const component = await componentNew(
  await componentEmbed({
    witSource,
    dummy: true,
  }),
);

const { files } = await transpile(component, {
  name: "stream-lift-only",
});

const generatedJs = new TextDecoder().decode(files["stream-lift-only.js"]);

console.log({
  hasStreamClass: generatedJs.includes("class Stream"),
  hasSymbolRscRepDecl: generatedJs.includes('const symbolRscRep = Symbol.for("cabiRep");'),
  symbolRscRepCount: (generatedJs.match(/symbolRscRep/g) ?? []).length,
});
```

## Actual Output

With `jco 1.24.0` / `js-component-bindgen 2.0.2`:

```js
{
  hasStreamClass: true,
  hasSymbolRscRepDecl: false,
  symbolRscRepCount: 1
}
```

The one `symbolRscRep` occurrence is a use, not a declaration.

The generated code contains:

```js
this[symbolRscRep] = args.globalRep;
```

but does not contain:

```js
const symbolRscRep = Symbol.for("cabiRep");
```

## Expected Output

If the generated external `Stream` class uses `symbolRscRep`, then the generated file should also emit:

```js
const symbolRscRep = Symbol.for("cabiRep");
```

## Important Distinction

`args.globalRep` is the stream representation value.

The generated code correctly checks that this value is present:

```js
if (args.globalRep === undefined) {
  throw new TypeError("missing host stream rep");
}
```

But `symbolRscRep` is not that value.

`symbolRscRep` is the JavaScript variable used as the property key where the representation value is stored:

```js
this[symbolRscRep] = args.globalRep;
```

For that assignment to work, the generated module must also declare the property-key symbol:

```js
const symbolRscRep = Symbol.for("cabiRep");
```

So the failure is not caused by a missing stream representation value. The failure is caused by the missing declaration of the symbol-key variable used to store that value on the JS stream object.

## Bug Trace

The generated external `Stream` class uses `Intrinsic::SymbolResourceRep.name()`.

The relevant generated shape is:

```js
class Stream {
  constructor(args) {
    if (args.globalRep === undefined) {
      throw new TypeError("missing host stream rep");
    }

    this[symbolRscRep] = args.globalRep;
  }
}
```

But the intrinsic dependency expansion for the external stream class does not appear to require `Intrinsic::SymbolResourceRep`.

The likely source-level fix is to make `AsyncStreamIntrinsic::ExternalStreamClass` imply `Intrinsic::SymbolResourceRep`.

Conceptually:

```text
ExternalStreamClass -> SymbolResourceRep
```

## Additional Local Check

This also reproduces for a stream import shape:

```wit
package test:stream-lower-only;

world w {
  import use-stream: func(data: stream<u8>);
  export run: func();
}
```

Generated-output check:

```js
{
  hasStreamClass: true,
  hasSymbolRscRepDecl: false,
  symbolRscRepCount: 1
}
```

So the issue is tied to emitting the external `Stream` class, not only to one export/import direction.

## Notes

This reproducer uses `jco` only as the public API. The generated-code issue appears to be in `js-component-bindgen` stream intrinsic dependency tracking.

gurkerl83:
## Summary

This fixes three small generated-JS correctness issues in `js-component-bindgen`.

1. UTF-16 string lifting now emits a mutable `offset` binding.
   - `_liftFlatStringUTF16` reassigns `offset` when normalizing bigint direct params.
   - The generated helper now uses `let offset`, matching the UTF-8 helper.
   - Fixes #1644

2. Enum lifting now emits its variant helper dependency.
   - `_liftFlatEnum` delegates to `_liftFlatVariant`.
   - `LiftFlatEnum` now requests `LiftFlatVariant`.
   - Fixes #1643

3. External stream generation now emits the resource-rep symbol dependency.
   - The generated external `Stream` class stores `args.globalRep` using `symbolRscRep`.
   - `SymbolResourceRep` is added after all paths that can request `ExternalStreamClass`.
   - This covers both lifted streams and streams created from lift.
   - Fixes #1645

## Verification

- `cargo check -p js-component-bindgen`
- Local generated-output checks for:
  - string-only UTF-16 helper
  - enum-only import lifting
  - stream export/import shapes

### DIFF
--- a/crates/js-component-bindgen/src/intrinsics/lift.rs
+++ b/crates/js-component-bindgen/src/intrinsics/lift.rs
@@ -724,7 +724,7 @@ impl LiftIntrinsic {
 
                         if (ctx.useDirectParams) {{
                             if (ctx.params.length < 2) {{ throw new Error('expected at least two u32 arguments'); }}
-                            const offset = ctx.params[0];
+                            let offset = ctx.params[0];
                             if (typeof offset === 'bigint') {{ offset = Number(offset); }}
                             if (!Number.isSafeInteger(offset)) {{  throw new Error('invalid offset'); }}
                             const len = ctx.params[1];
--- a/crates/js-component-bindgen/src/intrinsics/mod.rs
+++ b/crates/js-component-bindgen/src/intrinsics/mod.rs
@@ -1360,7 +1360,7 @@ pub fn render_intrinsics(args: RenderIntrinsicsArgs) -> Source {
             .contains(&Intrinsic::Lift(LiftIntrinsic::LiftFlatOption))
         | args
             .intrinsics
-            .contains(&Intrinsic::Lift(LiftIntrinsic::LiftFlatOption))
+            .contains(&Intrinsic::Lift(LiftIntrinsic::LiftFlatEnum))
     {
         args.intrinsics
             .extend([&Intrinsic::Lift(LiftIntrinsic::LiftFlatVariant)]);
--- a/crates/js-component-bindgen/src/intrinsics/mod.rs
+++ b/crates/js-component-bindgen/src/intrinsics/mod.rs
@@ -1551,6 +1551,12 @@ pub fn render_intrinsics(args: RenderIntrinsicsArgs) -> Source {
         ]);
     }
 
+    if args.intrinsics.contains(&Intrinsic::AsyncStream(
+        AsyncStreamIntrinsic::ExternalStreamClass,
+    )) {
+        args.intrinsics.insert(Intrinsic::SymbolResourceRep);
+    }
+
     if args
         .intrinsics
         .contains(&Intrinsic::AsyncStream(AsyncStreamIntrinsic::StreamWrite))


vados-cosmonic (approved):
LGTM 🚀 

Thanks @gurkerl83 for contributing the fix, happy to get this out and released ASAP

PR #1650, opened 2026-06-17
feat(p2-shim): migrate to typescript

### LINKED ISSUE #1596: [preview2-shim] Runtime values are exported only as types
I have built a component using  js-component-bindgen (v1.19.4). When trying to consume it together with preview2-shim (v0.18.1) in a TypeScript project, I get the following error:

> error TS2339: Property 'streams' does not exist on type 'typeof import("@bytecodealliance/preview2-shim/io")'.

It seems certain exports (e.g. streams, poll, etc) are actual runtime values, but they are being exported as type-only.

You can find the WASI preview2 component [here](https://github.com/eduardomourar/wasm-web-shell/tree/62f3604cbc8d438896342f74e32cf32af5f94ef4/www/aws-cli) and a workaround in [this commit](https://github.com/eduardomourar/wasm-web-shell/pull/7/changes/f1b7f44db33e58e017ddf1cbef7cbfe2e4d86fcd) where I just include the type definition for those runtimes values in a types.d.ts file (I could have just re-exported those casting as `any` too).

eduardomourar:
Related to #717.

---
Migrate preview2-shim to TypeScript
                                                                                          
This PR completes the TypeScript migration for the @bytecodealliance/preview2-shim package.

Changes

  Source Migration:
  - Migrated all source files from lib/ to src/ directory structure
  - Set up proper TypeScript compilation: src/ → dist/
  - Organized code into src/browser/, src/nodejs/, src/common/, src/io/, and src/synckit/

  Test Migration:
  - Converted all test files to TypeScript (test/*.js → test/*.ts)
  - Added proper type annotations for test utilities and fixtures
  - Fixed type errors related to socket addresses, HTTP requests, bigint conversions, etc.

  Type System Improvements:
  - Fixed AppendVersion type in instantiation.d.ts to correctly handle empty version strings
    - Before: wasi:random/random@ (incorrect)
    - After: wasi:random/random (correct)
  - Added type safety for WASI import objects
  - Proper typing for sandboxing configuration

  Build Configuration:
  - Updated tsconfig.json to compile TypeScript sources
  - Separated source compilation from test type-checking
  - Vitest handles test file type-checking independently

Migration Strategy

Complex low-level files (IO workers, sockets, HTTP, filesystem) are still using `any` for now, allowing gradual type refinement in future PRs while ensuring the codebase compiles successfully.

---
Fixes #1596.


vados-cosmonic:
Hey @eduardomourar this is huge, thanks for taking on this work! Will dig in and review soon

eduardomourar:
I believe having this in TypeScript will help with any confusion between types and values being exported/exposed in the public API.

packages/preview2-shim/.gitignore
```diff
@@ -0,0 +1,2 @@
+# TypeScript build outputs
+/lib/
```
vados-cosmonic: NIT: let's rename this to `dist`. I know the JS comunity isn't consistent on this, but if it's generated output let's put it in `dist` and package from there.

And if we need mutliple subfolders under `dist` (due to generating multiple distinct artifacts), so be it.

packages/preview2-shim/package.json
```diff
@@ -30,35 +30,47 @@
     "lib"
   ],
   "type": "module",
-  "types": "./types/index.d.ts",
   "exports": {
     ".": {
-      "types": "./types/index.d.ts",
-      "node": "./lib/nodejs/index.js",
+      "types": "./lib/browser/index.d.ts",
```
vados-cosmonic: NIT: lib -> dist

packages/preview2-shim/src/browser/filesystem.ts
```diff
@@ -93,84 +106,96 @@ function getChildEntry(parentEntry, subpath, openFlags) {
...
+  #entry!: FileDataEntry;
   #mtime = 0;
 
-  _getEntry(descriptor) {
+  _getEntry(descriptor: Descriptor): FileDataEntry {
     return descriptor.#entry;
   }
 
-  constructor(entry, isStream) {
+  static _create(entry: FileDataEntry | any, isStream?: boolean) {
```
vados-cosmonic, line 152: I've been trying to avoid this pattern, as it hurts tree-shakability... any reason it has to be this way?

packages/preview2-shim/src/browser/filesystem.ts
```diff
@@ -241,8 +271,8 @@ class Descriptor {
   }
 
   stat() {
-    let type = "unknown",
-      size = BigInt(0);
+    let type: TypesNamespace.DescriptorType = "unknown";
+    let size = BigInt(0);
```
vados-cosmonic:
```suggestion
    let size = 0n;
```

Any reason to not use the shorthands here?

packages/preview2-shim/src/browser/filesystem.ts
```diff
@@ -468,10 +516,6 @@ function convertFsError(e) {
       return "unsupported";
     case "ENOTTY":
       return "no-tty";
-    // windows gives this error for badly structured `//` reads
-    // this seems like a slightly better error than unknown given
-    // that it's a common footgun
-    case -4094:
```
vados-cosmonic, line 474: I don't think we should remove this special case, as we probably discovered this via an explicit report.

packages/preview2-shim/src/io/worker-http.ts
```diff
@@ -45,7 +51,7 @@ export async function startHttpServer(id, { port, host }) {
     // create the streams and their ids
     const streamId = createReadableStream(req);
     const responseId = ++responseCnt;
-    parentPort.postMessage({
+    parentPort?.postMessage({
```
vados-cosmonic: are there instances where we're getting here and `parentPort` is not set? we might want to error loudly if that's the case

packages/preview2-shim/src/nodejs/http.ts
```diff
@@ -374,15 +385,18 @@ function outgoingBodyDispose(id) {
...
 delete OutgoingBody._outputStreamId;
 
 const outgoingBodyCreate = OutgoingBody._create;
+// @ts-expect-error - Deleting static method
 delete OutgoingBody._create;
 
-class IncomingResponse {
-  /** @type {Fields} */ #headers = undefined;
+class IncomingResponse implements TypesNamespace.IncomingResponse {
+  #headers: Fields = undefined as any;
```
vados-cosmonic:
```suggestion
  #headers: Fields;
```

Do we need his `undefined as any`?

packages/preview2-shim/test/test.ts
```diff
@@ -3,12 +3,17 @@ import { throws } from "node:assert";
...
+  IpAddress,
+  IpSocketAddress,
+  Ipv4Address,
+} from "../types/interfaces/wasi-sockets-network.js";
 
 const symbolDispose = Symbol.dispose || Symbol.for("dispose");
 
 suite("Node.js Preview2", () => {
   test("Stdio", async () => {
-    const { cli } = await import("@bytecodealliance/preview2-shim");
```
vados-cosmonic, line 11: let's not change this -- using this via it's export name enables changing the version of p2-shim that is under test via `package.json`

vados-cosmonic (dismissed):
LGTM 🚀 

This looks overall good to me -- just a few things to address/discuss (mostly nits) and we can get it in.

packages/preview2-shim/package.json
```diff
@@ -30,35 +30,47 @@
...
+    "build": "tsc --project tsconfig.build.json",
+    "build:watch": "tsc --watch",
     "types:check": "tsc --noEmit",
     "fmt": "oxfmt",
     "fmt:check": "oxfmt --check",
     "lint": "oxlint",
     "lint:fix": "oxlint --fix",
-    "test": "vitest run -c test/vitest.ts"
+    "test": "vitest run -c test/vitest.ts",
+    "prepublishOnly": "pnpm run build"
```
vados-cosmonic: when we release we normally do a `pack`, so if you could change this to `prepack` that would be helpful

packages/preview2-shim/test/test.ts
eduardomourar: Yes, I will put it back. The only downside is that you need to do `npm run build` prior to testing otherwise you don't have anything available in the dist folder.

packages/preview2-shim/src/browser/filesystem.ts
eduardomourar: The types generated from WIT have a private constructor that prevents us from using new ...():
https://github.com/bytecodealliance/jco/blob/543c034e27e5ac295772a7d6f952a20138245dab/packages/preview2-shim/types/interfaces/wasi-filesystem-types.d.ts#L390-L394

If that behavior in the transpiler is changed, then we can move it to the constructor.

packages/preview2-shim/src/browser/filesystem.ts
eduardomourar: Not at all. I am changing it.

packages/preview2-shim/src/nodejs/http.ts
eduardomourar: It is because we don't have a constructor so typescript thinks that this has not been initialized. We know that it will always be a `Fields`. Following a similar process I did elsewhere in the code, I will set this to `#headers!: Fields` instead.

packages/preview2-shim/src/browser/filesystem.ts
vados-cosmonic: Ah thank you -- I was thinking this was only for information hiding purposes, you're right [descriptor doesn't have a constructor](https://github.com/WebAssembly/WASI/blob/main/proposals/filesystem/wit/types.wit#L311) (and isn't supposed to), so yeah this is the only reasonable way to do it.

What I wonder is if we could replicate this with some sort of private signal that only the host side has access to, but I can try that out as an experiment some time in the future

packages/preview2-shim/src/nodejs/http.ts
vados-cosmonic: ah thanks, I was confused but that makes sense

vados-cosmonic: I'll leave these unresolved until you're ready just in case you use these to track what's left to be done!

packages/preview2-shim/src/browser/filesystem.ts
eduardomourar: I will put it back.

packages/preview2-shim/src/io/worker-http.ts
eduardomourar: added error handling for it

eduardomourar:
@vados-cosmonic, all comments have been addressed. I was having some issues in the test/browser.ts file. The tests are all passing. There is a mismatch for options type while transpiling. I have removed some explicit options and the default behavior is in place for instantiation and wasi shim injection.

packages/preview2-shim/src/browser/filesystem.ts
eduardomourar: I would suggest having a optional flag in the transpiler to generated code assuming a default private constructor or not. The fact that we have initialization state tells us that our shim implemention needs a constructor even though the WIT definition does not have one.

vados-cosmonic:
Hey @eduardomourar thanks for the heads up -- I'll take a quick look! 

Please feel free to mark all the conversations that you've addressed as resolved!

vados-cosmonic:
hey @eduardomourar sorry for the delay here -- I really will get to the bottom of the browser tests -- they're unfortunately really unergonomic to debug, trying to make get a jco release out and then will get to this and release p2-shim as well.

eduardomourar:
I was able to get the other tests working, except the one for deno.

vados-cosmonic:
Oh that's fantastic!

Looking a the issues, it looks like something simple: 

```
error: Module not found "file:///home/runner/work/jco/jco/node_modules/@bytecodealliance/preview2-shim/dist/nodejs/cli.js".
    at file:///home/runner/work/jco/jco/crates/jco/tests/virtualenvs/scratch.mjs:1:35
```

~~If I had to guess `preview2-shim` doesn't have the `files` array populated with `dist` ?~~

Nope it's there... wonder if it's getting a stale version somehow or because it's not installed

eduardomourar:
I think is because I changed to the new dist/ folder, but that does not exist on previously published version:
https://github.com/bytecodealliance/jco/blob/3c64835aceecfa5e00c4b678ae435ed018676a0d/package.json#L17-L19

Let me know what you want me to do. Either I revert back that folder change for deno importmap or you merge it and include the new published version.

vados-cosmonic:
Yeah so it's weird because it's *looking* for the new code (so must be using the newer `package.json`), but somehow doesn't have the new package. If it was completely using the old package then it wouldn't be looking for `dist` in the first place, right?

I think the fix *might* be that we need to perform a build of `preview2-shim` during the test itself:

https://github.com/bytecodealliance/jco/blob/3c64835aceecfa5e00c4b678ae435ed018676a0d/.github/workflows/main.yml#L223

Do the deno tests fail locally?

Maybe I'm missing something here -- why would Deno take the newer `package.json` but be referring to an older release/the published version only? This is why I'm leaning towards it actually using the in-workspace version, but the built code jsut not being available @ `dist` in a fresh checkout in CI

eduardomourar:
I already included the preview2-shim build  to the test-wasi-deno, but it still fails. On my local machine, deno tests fail in the same way. Locally I see that the `node_modules/@bytecodealliance/preview2-shim` does have v0.17.9 (as pinned in the package.json file) which does not have a `dist/` folder. I have reverted the path change for deno importmap. That can be put back after a new version is released and pinned in the root package.json file.

.github/workflows/main.yml
```diff
@@ -1,6 +1,7 @@
 name: test
 
 on:
+  workflow_dispatch:
```
vados-cosmonic: we can remove this right? I assume this was from testing?

vados-cosmonic (dismissed):

.github/workflows/main.yml
eduardomourar: This was for testing, but I would advise to leave it there. Basically because it is the only way for PR author to trigger a fork-based workflow (further details here: https://github.com/orgs/community/discussions/176785).

vados-cosmonic: Ah so I'd recommend that on your fork to change the configuration to test CI stuff! Ideally people who fork should make whatever changes they'd like to make to enable testing locally (I often remove all other workflows, for example), then commit only the important changes upstream.

eduardomourar: I did change my fork, but now the forked main branch is not in sync with upstream. This change allows me to trigger a manual workflow in the fork context without having to bother you and without much hassle for any new contributor.

vados-cosmonic: IMO it's not terribly hard to reset your fork's `main` to match upstream. 

BTW, another way to do this that I often use is to open a PR against *your fork*, and you can make unlimited changes to CI stuff there. It's a viable way to test even things like github-pages deployments, and make any change you'd like to the CI scripts (and run them as you like).

Manual running of the main workflow is generally desirable in the scope of a code change -- which is normally in a PR, so I think the current setting is correct in that way... In general in upstream here we do not run the main set of tests manually.

PR #1671, opened 2026-06-24
refactor(jco): use oxc minify instead of terser

vados-cosmonic:


vados-cosmonic:
This change has been made and is working but the results look to be dramatically worse with `oxc-minify` -- what used to be 80,000 characters minified (in test) is now 177,335 lines, over 2x the length. 

The difference seems to be the terser `unsafe` optimizations which were clearly quite substantial. We'll have to wait for upstream to improve before this can be merged.

PR #1700, opened 2026-06-29
fix(bindgen): plug subtask memory leak during sync host calls

### LINKED ISSUE #1675: AsyncSubtask leak in handles table
I'm playing around with components for a node JS project. I'm getting OOMs on a worker thread using components with async support.

From grabbing some heapsnapshots, it looks like we're leaking `AsyncSubtask` on `ComponentAsyncState.handles`. They're added but never removed, so GC can't clear them.

I generated a reduced test case here: https://github.com/eqrion/jco-async-issue

```
world repro {
    import tick: func();
    export run: func(n: u32);
}
```

```
impl Guest for Component {
    fn run(n: u32) {
        for _ in 0..n {
            tick();
        }
    }
}
```
```
component.run(1000);
```

This grows the heap by about 60MiB, even with a manual GC in there.

I don't know the jco internals well, but patching the output so that `AsyncSubTask.onResolve` calls `this.#getComponentState().handles.remove(this.waitableRep());` fixed the issue for me. I don't see any existing code for removing a subtask from the handles.

I might be doing something wrong though, this is my first major jco project.

Tested on jco version: 1.24.3

vados-cosmonic:
This commit fixes a memory leak that occurred during host calls, as they do not call `subtask.drop` similar to async calls. Subtask objects accumulated in the `RepTable` class that tracks them (via `cstate.handles`) and were not being removed when a subtask itself would resolve.

Resolves #1675

### DIFF
--- a/crates/js-component-bindgen/src/intrinsics/mod.rs
+++ b/crates/js-component-bindgen/src/intrinsics/mod.rs
@@ -758,6 +758,7 @@ impl Intrinsic {
                 output.push_str(&format!(r#"
                     class {rep_table_class} {{
                         #data = [0, null];
+                        #size = 0;
                         #target;
 
                         constructor(args) {{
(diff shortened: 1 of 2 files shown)


vados-cosmonic:
One thing to note is that there is still *some* leakage -- there is an increase of roughly 6MB currently (without explicit GC) after running 1000 host calls -- the test checks to ensure that the increase is *below* 10MB for 1000 host calls. 

This isn't great (ideally there should be no increase in the size of MBs), but is dramatically better than the previous state (60MB according to the earlier report), so I'm committing this fix now with the hope of a focus on memory use in the future (just like we also need to focus on code size).

PR #1710, opened 2026-06-29
  fix(bindgen): emit Option/Result type aliases outside the world interface

### LINKED ISSUE #1708: jco transpile --instantiation emits invalid TypeScript .d.ts for exports returning result<T, E>
When transpiling with `--instantiation`, if an exported function returns a
  `result<...>`, jco generates the `Result<T, E>` helper type but places it
  **inside** the generated world interface (`Root`) instead of at the top level.
  A type-alias declaration is not a valid interface member, so the emitted
  `.d.ts` fails to parse under `tsc`.

  In the default (non-instantiation) mode the same `Result` type is emitted
  correctly at the top level, so the `.d.ts` is valid.

  ### Reproduction

  ```sh
  mkdir -p repro/wit && cd repro

  cat > wit/parser.wit <<'EOF'
  package example:parser;

  world parser {
    export parse-config: func(source: string) -> result<string, string>;
  }
  EOF

  # guest impl: return the value for `ok`, throw for `err`
  cat > impl.js <<'EOF'
  export function parseConfig(source) {
    return source;
  }
  EOF

  npx @bytecodealliance/jco@1.24.4 componentize impl.js \
    --wit wit/parser.wit --world-name parser -o parser.component.wasm

  npx @bytecodealliance/jco@1.24.4 transpile parser.component.wasm \
    -o out --name parser --instantiation async

  npx -p typescript tsc --noEmit --skipLibCheck out/parser.d.ts
  ```

This output was followings:

```
out/parser.d.ts:42:3 - error TS1131: Property or signature expected.

42   export type Result<T, E> = { tag: 'ok', val: T } | { tag: 'err', val: E };
     ~~~~~~

out/parser.d.ts:43:1 - error TS1128: Declaration or statement expected.

43 }
```

out/parser.d.ts:
```
 40 export interface Root {
 41   parseConfig(source: string): string,
 42   export type Result<T, E> = { tag: 'ok', val: T } | { tag: 'err', val: E };
 43 }
```

  ### Environment
I confirmed 
  - jco: `1.24.4` (also reproduces on `1.24.3`)
  - Node: `v26.4.0`

walf443:
When transpiling with `--instantiation`, a world's exports are wrapped in an
  `export interface {World} { ... }`. The `Option`/`Result` helper type aliases
  required by exported functions were emitted inside that interface, producing
  invalid TypeScript (a type-alias declaration is not a valid interface member),
  so the generated `.d.ts` failed to parse under `tsc`. The default
  (non-instantiation) mode was unaffected since it emits exports at the top level.

  Fixes #1708

### DIFF
--- a/crates/js-component-bindgen/src/ts_bindgen.rs
+++ b/crates/js-component-bindgen/src/ts_bindgen.rs
@@ -20,6 +20,22 @@ use crate::{
     uwrite, uwriteln,
 };
 
+/// TypeScript declaration for the `Option<T>` helper type alias used to
+/// represent WIT `option<T>` values that cannot be flattened to `T | undefined`.
+///
+/// This is always emitted at the top level of a module (or `.d.ts` file), never
+/// inside an `export interface { ... }` block, since a type-alias declaration is
+/// not a valid interface member
+/// (see https://github.com/bytecodealliance/jco/issues/1708).
+const TS_OPTION_TYPE_ALIAS: &str =
+    "export type Option<T> = { tag: 'none' } | { tag: 'some', val: T };\n";
+
+/// TypeScript declaration for the `Result<T, E>` helper type alias used to
+/// represent WIT `result<T, E>` values. Emitted at the top level for the same
+/// reason as [`TS_OPTION_TYPE_ALIAS`].
+const TS_RESULT_TYPE_ALIAS: &str =
+    "export type Result<T, E> = { tag: 'ok', val: T } | { tag: 'err', val: E };\n";
+
 struct TsBindgen {
     /// The source code for the "main" file that's going to be created for the
     /// component we're generating bindings for. This is incrementally added to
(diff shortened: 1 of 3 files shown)


vados-cosmonic:
Hey @walf443 thanks for the quick PR on this! I think you may need to trim your commit title and then run `cargo fmt` (and/or `pnpm run lint:fix`) and the other tests should run

[EDIT] ah there are some other commit title/body issues:

```
⧗   input:   fix(bindgen): emit Option/Result type aliases outside the world interface

  When transpiling with `--instantiation`, a world's exports are wrapped in an
  `export interface {World} { ... }`. The `Option`/`Result` helper type aliases
  required by exported functions were emitted inside that interface, producing
  invalid TypeScript (a type-alias declaration is not a valid interface member),
  so the generated `.d.ts` failed to parse under `tsc`. The default
  (non-instantiation) mode was unaffected since it emits exports at the top level.

  Fixes #1708
✖   header must not start with whitespace [header-trim]
✖   subject may not be empty [subject-empty]
✖   type may not be empty [type-empty]
```

I think what youw ere looking for should be:

```
fix(bindgen): emit Option/Result type aliases outside the world interface
```

Fairly certain that would pass.

vados-cosmonic (dismissed):
LGTM 🚀 

Happy to get this merged and released once we have the small linting issues fixed -- thanks for adding new tests to ensure the fixed and previous functionality.

walf443:
Thanks for quick review.

> Hey @walf443 thanks for the quick PR on this! I think you may need to trim your commit title and then run `cargo fmt` (and/or `pnpm run lint:fix`) and the other tests should run

@vados-cosmonic I fixed commit message.

I also ran cargo fmt and pnpm run lint:fix. It does not have problem.

vados-cosmonic:
Quick note -- I'll have to disable the new tests briefly to get the new functionality in this PR merged & released. 

This won't be necessary in the future, but it is right now only because of where the tests are (jco, rather than jco-transpile). That will be fixed in the future. I'll make sure to re-enable the new tests!

PR #1722, opened 2026-07-03
fix(js-component-bindgen): treat `undefined` as `none` when lowering options

GamePad64:
`_lowerFlatOption` mapped only JS `null` to the option `none` case, but jco's own generated TypeScript types model `option<T>` none as `T | undefined`. A value of `undefined` was therefore lowered as `some(undefined)`; when `T` contains an `own<resource>`, `_lowerFlatOwn` then throws `missing resource` and aborts the task.

This surfaced with wasi:http p3 `response.consume-body`, whose trailers future is typed `Result<Trailers | undefined, ErrorCode>`: an empty-trailers `Ok(None)` (`{ tag: 'ok', val: undefined }`) crashed lowering at body completion, leaving the enclosing streaming result's `.next()` unresolved.

Widen the none check to accept `undefined`, consistent with the generated types.

### DIFF
--- a/crates/js-component-bindgen/src/intrinsics/lower.rs
+++ b/crates/js-component-bindgen/src/intrinsics/lower.rs
@@ -918,7 +918,7 @@ impl LowerIntrinsic {
                             {debug_log_fn}('[{lower_flat_option_fn}()] args', {{ ctx }});
 
                             const v = ctx.vals[0];
-                            if (v === null) {{
+                            if (v === null || v === undefined) {{
                                 ctx.vals[0] = {{ tag: 'none' }};
                             }} else {{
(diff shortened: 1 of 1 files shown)


vados-cosmonic (approved):
LGTM 🚀 

Thanks for this fix @GamePad64 !

I really think that we should be leaning towards `null` here other than undefined, but the [current type representation documentation](https://bytecodealliance.github.io/jco/wit-type-representations.html#options-option) it looks like we really do use `undefined`. 

This is probably something left for a breaking change in Jco for the future -- `null` is an explicit value (that is none-ish) whereas `undefined` overlaps with a completely missing option, which makes `null` seem like the better representation here (there are some benefits like when making objects for record values)... That's a conversation for a different time, I guess.

PR #1737, opened 2026-07-07
fix(ci): build the release fresh, drop build caches

### LINKED ISSUE #1726: Remove cache use from release action
Currently the release actions use build caches of various sorts and they should not to avoid [cache poisoning attacks](https://docs.zizmor.sh/audits/#cache-poisoning) -- the CI should build everything fresh (we can still use artifacts to shuffle the freshly built artifacts/content we want to cache around).

GamePad64:
Closes #1726. The release workflow restored build caches (Swatinem/rust-cache `target/` + node_modules) instead of compiling fresh. Besides the cache-poisoning vector #1726 describes, the restored `target/` also prevents fat-LTO from re-running its cross-crate codegen, so every publish silently shipped an unoptimized js-component-bindgen to jco-transpile's in-browser consumers.

Remove all build-cache restores from release.yml; keep upload/download artifacts to move freshly-built content between jobs.

Vendored js-component-bindgen (js-component-bindgen-component.core.wasm), measured on [python-env](https://github.com/actpkg/python-env) (~113 MiB wasm):
  cached  (no LTO):  8.9 MiB / 17,616 funcs / ~56 s transpile
  fresh   (fat-LTO): 3.1 MiB /  7,842 funcs / ~7.5 s transpile

A fresh build applies the release profile's `lto=true` (verified with the same rustc, 1.96.1, that shipped 0.4.1); the cache-restored build did not.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -94,14 +94,6 @@ jobs:
         with:
           version: 11
 
-      - name: Cache node modules
-        id: cache-node-modules
-        uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
-        with:
-          key: node-modules-dev-${{ runner.os }}-${{ runner.arch }}-${{ hashFiles('pnpm-lock.yaml') }}
-          path: |
-            node_modules
-
       - name: Install debug packages
         run: |
           pnpm install -D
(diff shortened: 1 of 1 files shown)


vados-cosmonic (approved):
LGTM 🚀

vados-cosmonic:
Hey @GamePad64 would you mind renaming your commit title to something like `fix(ci): ...` ?

We're possibly a bit overly strict about conventional commits, but since the check is automated it's a bit easy to run afoul of it.

PR #1738, opened 2026-07-08
fix(bindgen): Add string intrinsics for flat string lowering

lann:
These dependencies are often fulfilled by e.g. WASI interfaces, but a custom world can be missing them.


vados-cosmonic (approved):
LGTM 🚀 

Thanks for the fix @lann -- I sorely need to rework how these deps are determined and place them a bit closer to the actual use.

vados-cosmonic:
Hey just an update here -- [Jco v1.25.2](https://www.npmjs.com/package/@bytecodealliance/jco/v/1.25.2) is out which contains this fix (via updated `jco-transpile` and `js-component-bindgen` underneath)

PR #1739, opened 2026-07-08
fix(bindgen): async host import stream/future result lowering

lann:
Async host imports returning stream/future were double-lowered: the async return machinery (LowerImport -> subtask.onResolve -> resultLowerFns) lowers the result, but StreamLower/FutureLower also lowered it inline in the trampoline, consuming/locking the host value.

- Skip the inline lower for async imports via is_async && for_import, mirroring the existing StreamLift/FutureLift dispatch.
- Fix _lowerFlatFuture's host-promise branch (never-run while loop, uninitialized list, undefined stringEncoding, const reassignment) to mirror the inline FutureLower codegen.
- Fix gen_flat_lower_fn_js_expr emitting _lowerFlatFuture.bind(null, meta) instead of invoking the factory, matching the stream path.

Adds return-host-async-value test component/world and test/p3/import-result-lowers.ts regression coverage.

### DIFF
--- a/crates/js-component-bindgen/src/intrinsics/lower.rs
+++ b/crates/js-component-bindgen/src/intrinsics/lower.rs
@@ -1039,7 +1039,7 @@ impl LowerIntrinsic {
                         return function {lower_flat_future_fn}Inner(ctx) {{
                             {debug_log_fn}('[{lower_flat_future_fn}()] args', {{ ctx }});
 
-                            const future = ctx.vals[0];
+                            let future = ctx.vals[0];
                             if (!future) {{ throw new Error("missing external future value"); }}
 
                             // As NodeJS will collapse `Promise<Promise<T>>` to `Promise<T>`, enable handling of ordinary values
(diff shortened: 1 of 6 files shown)


crates/js-component-bindgen/src/intrinsics/lower.rs
lann, line 1050: This is what motivated the `const` -> `let` above.

lann:
This is another / similar instance of https://github.com/bytecodealliance/jco/issues/1601. The fix was generated by Claude; the diff looks sensible to me but I don't have enough context to evaluate it in the overall architecture.

crates/js-component-bindgen/src/intrinsics/lower.rs
vados-cosmonic: yeah this is just a bug/typo, thanks

vados-cosmonic:
Appreciate the fix here @lann -- after an initial pass the fixes look right to me (the `.bind` in gen flat lower is definitely a bug, at any rate). Will take another quick look and get this in soon & do a release of jco-transpile.

lann:
Thanks!

lann:
FWIW this came out of getting this working: https://github.com/lann/wasm-component-starter/tree/main/examples/apps/browser-tgz-maker

I have confirmed that this PR along with #1738 fix that example.

vados-cosmonic (approved):
LGTM 🚀 

Thanks for the contribution @lann  -- this does indeed seem to be a case of double-lowering due to async host task handling.

vados-cosmonic:
Hey just an update here -- [Jco v1.25.2](https://www.npmjs.com/package/@bytecodealliance/jco/v/1.25.2) is out which contains this fix (via updated `jco-transpile` and `js-component-bindgen` underneath)

PR #1744, opened 2026-07-09
fix(transpile): stop the debug bootstrap clobbering the release bindgen

GamePad64:
`transpile()` writes the componentized module to
`obj/<name>.component.wasm`, then, on a clean `obj/`, runs a debug bootstrap (`transpile_components(BuildType::Debug)`) to build the `wasm-tools` transpiler that `jco opt` needs. That bootstrap transpiles *all* components, including the one currently being built, into the same `obj/*.component.wasm` paths, overwriting the release component with the 132 MiB debug build. `jco opt` and the final transpile then run on the debug component, so `build:release` emits the debug bindgen (wasm-opt'd to ~8.9 MiB / 17,688 funcs) instead of the release one (~3.1 MiB / 7,836 funcs).

Re-write the release component to `component_path` after the bootstrap so `jco opt` operates on the release build. The bootstrap only runs when `obj/wasm-tools.js` is absent (a clean checkout), which is exactly how the release workflow builds, so every published js-component-bindgen bindgen was the debug one.

### DIFF
--- a/crates/js-component-bindgen-component/src/lib.rs
+++ b/crates/js-component-bindgen-component/src/lib.rs
@@ -169,7 +169,7 @@ impl bindings::Guest for JsComponentBindgenComponent {
             .build();
 
         let files = js_component_bindgen::generate_types(&name, resolve, world, opts)
-            .with_context(|| format!("generating types for [{}]", &name))
+            .with_context(|| format!("generating types for [{}]", name))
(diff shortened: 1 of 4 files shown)


vados-cosmonic (dismissed):
LGTM 🚀 

Thanks @GamePad64 !

PR #1756, opened 2026-07-22
add implements support

ricochet:
Adds support for the component model `implements` proposal (WebAssembly/component-model#613 and WebAssembly/component-model#672):
- `import label: ns:pkg/iface` / `export label: iface` in WIT: named (labeled) imports and exports of interfaces, including importing the same interface under multiple labels. Transpiled bindings import each label under its own specifier, and `--map` entries for the implemented interface id apply to labeled imports (label-specific mappings win)
- resource-bearing interfaces imported under multiple labels resolve each label's resource class through that import instance's own resource table rather than the wit type, which is shared by all labels
- `jco types` aliases each label to the shared interface module
- `@external-id("...")` attributes round-trip through embed/new/print/wit
- `(implements "...")` name validation is enabled during transpile via `WasmFeatures::CM_IMPLEMENTS`

Note: binaries carrying `(external-id ...)` names can be produced and printed but not yet transpiled. wasmtime-environ 47 embeds wasmparser 0.252, which predates external-id name-option parsing (0.253).

This also updates wasmtime-environ from 46.0.1 to 47.0.2, enabling its `anyhow` feature for error conversions. Migration notes:
- the per-instance flags global now holds a single boolean `may_leave` flag rather than a MAY_ENTER|MAY_LEAVE bitmask, so the generated JS global is initialized to `1`
- wasmtime-environ 47's FACT-generated adapters wrap calls in exception barriers (`try_table` + `catch_all`) when the exceptions feature is enabled. The multi-memory augmenter learns to re-encode `try_table`, but the exceptions feature is masked off the transpile validator for now since the resulting core modules would only run behind --experimental-wasm-exnref in today's JS engines.

I went ahead and bumped js-component-bindgen to 2.1.0 as this is a new feature. Is that the preferred flow here?

### DIFF
--- a/crates/js-component-bindgen/src/core.rs
+++ b/crates/js-component-bindgen/src/core.rs
@@ -52,10 +52,6 @@ use wasmparser::{
 use wasmtime_environ::component::CoreDef;
 use wasmtime_environ::{EntityIndex, MemoryIndex, ModuleTranslation, PrimaryMap};
 
-fn unimplemented_try_table() -> wasm_encoder::Instruction<'static> {
-    unimplemented!()
-}
-
 pub enum Translation<'a> {
     Normal(ModuleTranslation<'a>),
     Augmented {
@@ -790,7 +786,7 @@ macro_rules! define_translate {
     (mk F32Const $v:ident) => (F32Const(Ieee32::from(f32::from_bits($v.bits()))));
     (mk F64Const $v:ident) => (F64Const(Ieee64::from(f64::from_bits($v.bits()))));
     (mk V128Const $v:ident) => (V128Const($v.i128()));
-    (mk TryTable $v:ident) => (unimplemented_try_table());
+    (mk TryTable $v:ident) => (TryTable($v.0, $v.1));
     (mk MemoryGrow $($x:tt)*) => ({
         if true { unimplemented!() } Nop
     });
@@ -836,7 +832,7 @@ macro_rules! define_translate {
         $arg.targets().map(|i| i.unwrap()).collect::<Vec<_>>().into(),
         $arg.default(),
     ));
-    (map $self:ident $arg:ident try_table) => {$arg};
+    (map $self:ident $arg:ident try_table) => {$self.try_table($arg)};
     (map $self:ident $arg:ident struct_type_index) => {$self.remap(Item::Type, $arg).unwrap()};
     (map $self:ident $arg:ident field_index) => {$arg};
     (map $self:ident $arg:ident array_type_index) => {$self.remap(Item::Type, $arg).unwrap()};
@@ -893,6 +889,31 @@ impl Translator<'_, '_> {
             wasmparser::BlockType::Type(t) => wasm_encoder::BlockType::Result(valtype(t)),
             wasmparser::BlockType::FuncType(i) => wasm_encoder::BlockType::FunctionType(i),
         }
+    }
+
+    /// Re-encode a `try_table` immediate. Wasmtime's FACT-generated adapters
+    /// wrap lowerings in exception barriers (`try_table` + `catch_all`), and
+    /// tags are not remapped by augmentation, so catches pass through as-is.
+    fn try_table(
+        &self,
+        tt: wasmparser::TryTable,
+    ) -> (
+        wasm_encoder::BlockType,
+        std::borrow::Cow<'static, [wasm_encoder::Catch]>,
+    ) {
+        let catches = tt
+            .catches
+            .iter()
+            .map(|c| match *c {
+                wasmparser::Catch::One { tag, label } => wasm_encoder::Catch::One { tag, label },
+                wasmparser::Catch::OneRef { tag, label } => {
+                    wasm_encoder::Catch::OneRef { tag, label }
+                }
+                wasmparser::Catch::All { label } => wasm_encoder::Catch::All { label },
+                wasmparser::Catch::AllRef { label } => wasm_encoder::Catch::AllRef { label },
+            })
+            .collect::<Vec<_>>();
+        (self.blockty(tt.ty), catches.into())
     }
     fn heapty(&self, _ty: wasmparser::HeapType) -> wasm_encoder::HeapType {
         unimplemented!()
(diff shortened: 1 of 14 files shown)


crates/js-component-bindgen/src/core.rs
```diff
@@ -790,7 +786,7 @@ macro_rules! define_translate {
     (mk F32Const $v:ident) => (F32Const(Ieee32::from(f32::from_bits($v.bits()))));
     (mk F64Const $v:ident) => (F64Const(Ieee64::from(f64::from_bits($v.bits()))));
     (mk V128Const $v:ident) => (V128Const($v.i128()));
-    (mk TryTable $v:ident) => (unimplemented_try_table());
+    (mk TryTable $v:ident) => (TryTable($v.0, $v.1));
```
vados-cosmonic, line 789: Besides the point but a note: so this works for 252, but doesn't actually work for 254 as the structure of this instruction has actually changed (to a struct)

I think we maybe should get the upstream releases of `wasmparser` and `wasmtime-environ` on the same page w/ regard to 254 -- this PR updates only wasmitme-environ (and not wasmtime nor the other upstream deps) and I'd rather update these all at the same time.

That said I think I'm going to just take this as-is for now and complete the update later, since the tests look good!

crates/js-component-bindgen/src/lib.rs
```diff
@@ -128,14 +128,21 @@ pub fn transpile(component: &[u8], opts: TranspileOpts) -> Result<Transpiled> {
     // This does not require the correct execution of the related features post-transpilation,
     // but without the right features specified, components won't load at all.
     let mut validator = wasmtime_environ::wasmparser::Validator::new_with_features(
-        WasmFeatures::WASM3
+        (WasmFeatures::WASM3
```
vados-cosmonic: NIT: What do you think about separatting this set of features to up above, remove the difference to it's own section below that, and then actually control the existetnce of wasm exnref via opts for transpile.

So something like:

```rust
let mut features = ...
if opts.supports_wasm_exnref {
    features = features.difference(...);
}
let mut validator = ...
```

crates/js-component-bindgen/src/transpile_bindgen.rs
```diff
@@ -2981,6 +2983,50 @@ impl<'a> Instantiator<'a, '_> {
...
+    /// method/static or the target of a constructor).
+    ///
+    /// The same interface -- and thus the same `wit-parser` resource type --
+    /// may be imported more than once under different labels via the
+    /// component model `implements` feature, each import with its own
+    /// resource table. The class must therefore be resolved through the
+    /// specific import instance rather than through the wit type, which is
+    /// shared by all labels.
+    #[allow(clippy::too_many_arguments)]
+    fn imported_resource_name(
```
vados-cosmonic: I *think* his should probably be a method on the instantiator itself (to avoid the too-many-arguments), did that not work? Most of these are coming from the instantiatior, but maybe you ran into some mut and immut borrow issues?

crates/js-component-bindgen/src/transpile_bindgen.rs
```diff
@@ -3598,54 +3669,124 @@ impl<'a> Instantiator<'a, '_> {
...
+                        break 'search;
+                    }
+                    _ => {}
+                }
+            }
+
+            let (world_key, iface_name) = match import_entry {
+                // Resource provided by an imported instance: derive the
+                // interface member name from the world key shape.
+                Some((imp_name, true)) => {
```
vados-cosmonic:
```suggestion
                Some((imp_name, _is_from_instance @ true)) => {
```

Personal preference here -- when we have bare booleans I like to leave a descriptive unused name to remind us of what it means

crates/js-component-bindgen/src/transpile_bindgen.rs
```diff
@@ -3598,54 +3669,124 @@ impl<'a> Instantiator<'a, '_> {
...
-                            )
+            // Resolve the specific import that provides this resource table.
+            // The same interface may be imported under multiple labels
+            // (component model `implements` feature), each label with its own
+            // resource table, so the import is identified through the
+            // wasmtime resource index rather than through the wit type's
+            // owning interface, which is shared by all labels.
+            let component = self.component;
+            let types = self.types;
+            let mut import_entry: Option<(&str, bool)> = None;
```
vados-cosmonic:
```suggestion
            let mut imported_resource_entry: Option<(&str, bool)> = None;
```

crates/js-component-bindgen/src/transpile_bindgen.rs
```diff
@@ -3598,54 +3669,124 @@ impl<'a> Instantiator<'a, '_> {
...
-                                    WorldKey::Interface(_) => None,
-                                },
-                            )
+            // Resolve the specific import that provides this resource table.
+            // The same interface may be imported under multiple labels
+            // (component model `implements` feature), each label with its own
+            // resource table, so the import is identified through the
+            // wasmtime resource index rather than through the wit type's
+            // owning interface, which is shared by all labels.
+            let component = self.component;
```
vados-cosmonic: NIT: IMO this block should probably be factored out into a helper

crates/js-component-bindgen/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "js-component-bindgen"
-version = "2.0.11"
+version = "2.1.0"
```
vados-cosmonic: Yeah we normally change this version @ release time -- but good to know this is a minor release.

packages/jco-transpile/test/cm/implements.ts
vados-cosmonic, line 1: These tests are excellent, but for something like this we generally also have a component test with actual usage to match as kind of an "E2E" test, could you add one real quick? 

Also, there is actually a spot for where CM specific tests and type generation tests go: 

`test/cm/implements.ts` is probably a decent spot, and `test/typegen.ts` is probably another good place.

vados-cosmonic (dismissed):
LGTM 🚀 

This is basically ready, nothing but nits here -- thanks for knocking this out @ricochet. Shame we can't fully be sure it will work given the lack of support of try_table/exnref across engines (TIL), but certainly happy to land support for this sooner rather than later.

vados-cosmonic (approved):
LGTM 🚀

PR #1760, opened 2026-07-24
fix(bindgen): Fix dropping multiple borrow types in one func

lann:
Two different versions of the borrow cleanup code were emitted depending on whether one (random) type was host-owned, which breaks when a func takes both host- and guest-owned resources at once.

This change makes the emitted code handle either kind of resource conditionally at runtime.


vados-cosmonic (dismissed):
LGTM 🚀 

Great catch @lann , and thanks for adding a regression test!

vados-cosmonic:
Did a quick lint and fixed the commit title. 

Note that the perf check test is flaky because CI perf varies, but I'll hold off on raising the CI-only limit unless it consistently doesn't pass (and do a quick local check to ensure we didn't actually regress, though of course this change should absolutely not regress anything).

PR #1761, opened 2026-07-25
feat(jco): add --wizer-bin option to componentize command

Sorixelle:
ComponentizeJS supports a `wizerBin` option to override the path to the Wizer binary, but jco doesn't expose any way to control that option. This is important for environments where it isn't possible to use the prebuilt Wizer binaries from NPM, such as NixOS machines (which [can't run dynamically linked binaries built outside of Nix](https://nix.dev/guides/faq#how-to-run-non-nix-executables)).

This PR adds a `--wizer-bin` command line argument to the `componentize` command, following the same pattern as the existing `--weval-bin` flag.


vados-cosmonic (approved):
LGTM 🚀 

Thanks for the contribution @Sorixelle -- happy to get this in ASAP in the next release

PR #1766, opened 2026-07-25
feat(jco): convert jco to typescript

### LINKED ISSUE #717: Moving Jco to Typescript

vados-cosmonic:
Resolves #717


packages/jco/package.json
vados-cosmonic:
```suggestion
"build": "cargo xtask build debug && npm run build:ts",
"build:release": "cargo xtask build release && npm run build:ts",
```

PR #1768, opened 2026-07-25
fix(bindgen): lowered 8-bit writes actually write 8 bits

lann:
Lowered bool, u8, and s8 were using set(U)Int32, which accidentally works most of the time since Wasm is little-endian and values are typically written in memory order. Where it can fail is at the end of a destination buffer where it can cause up to three bytes after the buffer to be overwritten with zeros.


lann:
Regression tests written by Claude, reviewed by me.

vados-cosmonic (approved):
LGTM 🚀 

Yeah this is definitely a bug, must have been a copypasta mistake somewhere, thanks for the fix.

PR #1775, opened 2026-07-26
fix(bindgen): reject task on driver loop exception

lann:
Internal exceptions (e.g. bindgen bugs) get caught and swallowed by the task driver loop. This makes them invisible to the caller an completely silent unless JCO_DEBUG is set.

This change rejects the task with the caught error.

### DIFF
--- a/crates/js-component-bindgen/src/intrinsics/p3/async_task.rs
+++ b/crates/js-component-bindgen/src/intrinsics/p3/async_task.rs
@@ -2072,6 +2072,8 @@ impl AsyncTaskIntrinsic {
                                 }},
                                 err,
                             }});
+                            task.setErrored(err);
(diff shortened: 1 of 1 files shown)


lann:
This _seems_ straightforward to me, but I don't really understand why this was omitted in the first place so I may be missing something. 🤷

vados-cosmonic:
Hey so @lann I think this was the case before, but *any* error during driver loop execution would cause it, and bindgen was buggier at the time (IIRC subtask rejection would cause spurrious errors). Either way I think this is right and probably something that was just not put back that should have been.

vados-cosmonic (approved):
LGTM 🚀 

Thanks for the fix -- this looks just fine to me.

PR #1785, opened 2026-07-27
fix(bindgen): generate named stream and future types

andreiltd:
Emit TS aliases for named wit streams and futures in world and interface declarations.

### DIFF
--- a/crates/js-component-bindgen/src/ts_bindgen.rs
+++ b/crates/js-component-bindgen/src/ts_bindgen.rs
@@ -262,8 +262,8 @@ pub fn ts_bindgen(
                             generator.type_fixed_size_list(*id, name, t, len, &ty.docs)
                         }
                         TypeDefKind::Type(t) => generator.type_alias(*id, name, t, None, &ty.docs),
-                        TypeDefKind::Future(_) => todo!("(async impl) generate for future"),
-                        TypeDefKind::Stream(_) => todo!("(async impl) generate for stream"),
+                        TypeDefKind::Future(t) => generator.type_future(name, t, &ty.docs),
+                        TypeDefKind::Stream(t) => generator.type_stream(name, t, &ty.docs),
                         TypeDefKind::Unknown => unreachable!("(async impl) generate for unknown"),
                         TypeDefKind::Resource => {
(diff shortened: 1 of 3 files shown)


crates/js-component-bindgen/src/ts_bindgen.rs
```diff
@@ -1441,6 +1441,26 @@ impl<'a> TsInterface<'a> {
...
+            name.to_upper_camel_case()
+        ));
+        self.print_optional_ty(element_ty.as_ref());
+        self.src.push_str(">;\n");
+    }
+
+    fn type_stream(&mut self, name: &str, element_ty: &Option<Type>, docs: &Docs) {
+        self.docs(docs);
+        self.src.push_str(&format!(
+            "export type {} = ReadableStream<",
```
vados-cosmonic: Let's be more accepting here and use `AsyncIterator<T>`, IIRC you should be able to use that from Typescript.

andreiltd: I assumed you meant `AsyncIterable<T>`  here?

vados-cosmonic: Ah yes -- `AsyncIterable<T>`!

crates/js-component-bindgen/src/ts_bindgen.rs
```diff
@@ -1441,6 +1441,26 @@ impl<'a> TsInterface<'a> {
         }
     }
 
+    fn type_future(&mut self, name: &str, element_ty: &Option<Type>, docs: &Docs) {
+        self.docs(docs);
+        self.src.push_str(&format!(
+            "export type {} = Promise<",
```
vados-cosmonic: Hey sorry I forgot to mention it, but here we actually want `PromiseLike<T>`, because of the whole nesting futures thing. We can write `Promise<Promise<T>>` but the thing we hand back is eventually not going to be a Promise object exactly -- it will have to be something `.then`-able

vados-cosmonic (dismissed):
LGTM 🚀 

One nit about the `Promise<T>` and this should be ready to go

PR #1798, opened 2026-07-30
chore(transpile): add regression test for filesystem-open-errors

vados-cosmonic:


vados-cosmonic:
So it turns out this isn't the bug -- the real bug was inside p3-shim

PR #1838, opened 2026-08-05
fix(bindgen): fix string flat lowering of multibyte utf-8 and multiunit utf-16 chars

lann:
UTF-8 length was lowered as codepoints instead of bytes, breaking on multibyte chars.

UTF-16 length was lowered as codepoints instead of code _units_, breaking on multi-unit chars. Additionally, the UTF-16 lowering was redundantly re-encoding the string - possibly a refactoring mistake?


vados-cosmonic (approved):
LGTM 🚀 

Thanks for fixing this and updating the test!

PR #1839, opened 2026-08-05
fix(bindgen): fix TypeError from sync-lowered async funcs

lann:
These just needed `new Promise()` - always an error - to be replaced with `PromiseWithResolversPonyfill` (I don't understand the kids these days).


vados-cosmonic (approved):
LGTM 🚀 

This definitely snuck in there, good catch. and thanks for using the ponyfill :). I also don't know what the kids are thinking these days but we're ponyfillin' round here.

PR #1840, opened 2026-08-05
fix(bindgen): fix utf-8 to utf-16 transcode length calculation

lann:
This was mistakenly doubling the length of the array view (presumably thinking the Uint16Array length was in bytes). I renamed the var from `strlen` to `codeUnits` so that the next person to see it will at least stop and think "what the heck is a code unit?".

This only rarely triggered a real bug; if the destination was near the end of the Memory and/or the string was long enough then it could overflow, triggering `RangeError`.


vados-cosmonic (approved):
LGTM 🚀 

Thanks or the fix here -- there is an issue out for a UTF16 component test, so I'll try to add that as a follow up at some point

PR #1842, opened 2026-08-09
feat(p3-shim): Add p3 browser shim skeleton

MendyBerger:
This PR adds a p3 browser shim skeleton, so the shim can be filled out incrementally over time.

The skeleton has no actual implementations, but it's still useful: components generated by p3-targeting toolchains import these interfaces whether or not they end up calling them, and stubs let those components load rather than fail to resolve.

It also leans on TypeScript to enforce that every interface is fully covered.

### DIFF
--- a/.oxlintrc.json
+++ b/.oxlintrc.json
@@ -21,5 +21,17 @@
       "error",
       "all"
     ]
-  }
+  },
+  "overrides": [
+    {
+      // Allow unused vars since most of the browser shims just throw for now.
+      // TODO: remove once browser shims are mostly covered.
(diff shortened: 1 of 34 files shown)


vados-cosmonic (approved):
LGTM :rocket:

This is a great start, thanks @MendyBerger!

vados-cosmonic:
BTW, your commit likely needs the post-scope leading capital "A" turned into a "a"

andreiltd (approved):
Looks great 🎉

vados-cosmonic (approved):

Issue #1898, opened 2026-08-19
wasi 0.3 sync-lowered-async transpile codegen bugs (5 issues with repro)


evanlhatch:
Found while running a `wasm32-wasip3` component under `jco transpile` (default `--async-mode sync`) + `@bytecodealliance/preview3-shim` on Node 24 (`--experimental-wasm-jspi`). All five bugs are in js-component-bindgen's wasi-0.3 / sync-lowered-async codegen path, and **all five reproduce with bindgen built from `main` (faca7e5d)** as well as with the released `@bytecodealliance/jco@1.29.0` (`jco-transpile@0.8.0`).

## Repro

Public repro component (wasm32-wasip3, imports `wasi:filesystem/*@0.3.0`, `wasi:clocks/*@0.3.0`, exports a driver interface incl. a function named `eval`):

- component: https://github.com/evanlhatch/wasmtron/blob/main/components/runtimes/otto/otto.component.wasm
- full runnable harness (with the patches we carry, and the evidence notes): https://github.com/evanlhatch/wasmtron/tree/main/tools/web (`patch-out.mjs`) and https://github.com/evanlhatch/wasmtron/blob/main/notes/web-jco-spike.md

```sh
npm i -g @bytecodealliance/jco@1.29.0
jco transpile otto.component.wasm -o out

# bug 1: grep -n "async function eval(" out/otto.component.js        -> 1 hit
# bug 2: grep -n "subtask.isResolvedState()" out/otto.component.js   -> present; importFn never called in that branch
# bug 3: grep -n "trampoline47 = async function" out/otto.component.js -> open-at reads params via dataView(memory0) while the core import is 7 flat i32
# bug 4: grep -c "return await task.completionPromise();" out/otto.component.js -> 20 sites
# bug 5: grep -c "class FutureReadableEnd" out/otto.component.js     -> 0 (InternalFuture IS emitted)
```

At runtime (stock output, Node 24 + preview3-shim 0.3.0): bug 1 is a `SyntaxError` at module load; after renaming, bug 2 throws on the first fs call; after fixing that, bug 3 misreads params; bug 4 then hangs; bug 5 throws on the first `future`-typed value.

---

## 1. Exported function named `eval` is not sanitized

**WIT shape:** any exported interface function named `eval`:

```wit
package repro:eval@0.1.0;
interface driver { eval: func(src: string) -> string; }
world w { export driver; }
```

[...]

PR #1899, opened 2026-08-19
fix(bindgen): sort RESERVED_KEYWORDS so binary_search finds `eval`

evanlhatch:
Fixes bug 1 of #1898.

## Problem

`RESERVED_KEYWORDS` in `crates/js-component-bindgen/src/names.rs` was not sorted: `"eval"` was placed before `"else"`/`"enum"`, but sorts after both. `is_js_reserved_word()` uses `RESERVED_KEYWORDS.binary_search(...)`, and with the misordered list the search never finds `"eval"` — replaying Rust's `slice::binary_search` probe sequence over the list shows `eval` was the **only** keyword missed.

As a result, `to_js_identifier("eval")` returned `"eval"` unchanged, and a component export named `eval` was emitted verbatim into generated ESM:

```js
async function eval(arg0) { ... }
```

which fails to parse (ESM is strict mode):

```
SyntaxError: Unexpected eval or arguments in strict mode
```

## Repro

WIT:

```wit
package repro:eval@0.1.0;
interface driver { eval: func(src: string) -> string; }
world w { export driver; }
```

Embed into any core module (`wasm-tools component embed` + `wasm-tools component new`), then `jco transpile` — output contains `function eval(`. Reproduced with `@bytecodealliance/jco@1.29.0` and with bindgen built from `main` (see #1898 for full details and a public repro component).

## Fix

Move `"eval"` to its sorted position (after `"enum"`). With the fix, `to_js_identifier("eval")` produces `_eval` and the export mapping becomes `eval: _eval`, which is valid strict-mode JS.

## Tests

Added `#[cfg(test)]` unit tests in `names.rs`:

- `reserved_keywords_are_sorted_and_searchable` — asserts the list is sorted and that `binary_search` finds every entry (prevents this class of bug for any future mis-insertion).
- `eval_is_sanitized` — regression test that `eval` is treated as reserved and sanitized to a valid identifier.

No local Rust toolchain was available; relying on CI for `cargo test`.

### DIFF
--- a/crates/js-component-bindgen/src/names.rs
+++ b/crates/js-component-bindgen/src/names.rs
@@ -178,9 +178,9 @@ pub(crate) const RESERVED_KEYWORDS: &[&str] = &[
     "default",
     "delete",
     "do",
-    "eval",
     "else",
     "enum",
+    "eval",
     "export",
     "extends",
     "false",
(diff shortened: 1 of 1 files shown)


vados-cosmonic (approved):
LGTM 🚀 

Great catch @evanlhatch thanks for the contribution.

PR #1946, opened 2026-08-22
test(bindgen): finish enabling wast test machinery

vados-cosmonic:


vados-cosmonic:
Note that this *does not* mean that all the WAST tests actually pass :) See follow up issue #1947

PR #1979, opened 2026-08-28
feat(jco): inject node api WIT imports, support node:child_process

vados-cosmonic:


vados-cosmonic:
export overhead calls are failing here, and we haven't done the perf exploration yet, so going to merge this for now!

PR #1994, opened 2026-09-01
feat(std): support node fs api

vados-cosmonic:


vados-cosmonic:
NOTE: this PR is absolutely massive because it *also* includes a fix to the `node:dns` API which was using a worker when it didn't need to be, given JSPI support.

PR #2045, opened 2026-09-05
fix(preview2-shim): disposing a settled future incoming response aborts its still-streaming body

### LINKED ISSUE #2044: preview2-shim: disposing a settled future incoming response aborts its still-streaming body
Found this while verifying the would-block fix from #2042 against a real AWS SDK client (aws-cli compiled to a component, running in the browser via JSPI). `aws s3 ls` would fail partway through with:
> response error
Caused by:
0: reading incoming body stream
1: BodyStreamBuffer was aborted

`packages/preview2-shim/src/browser/http.ts`, `FutureIncomingResponse[Symbol.dispose]()`:
 ```js
[symbolDispose]() {
   this.#controller?.abort();
   this.#controller = null;
   this.#promise = null;
}
```

This unconditionally aborts the AbortController tied to the original fetch() call - including after the fetch already resolved successfully. The guest typically drops the future-incoming-response handle right after extracting the response (well before the body is fully read), since ownership of the body moves to the separate incoming-response/incoming-body resources. That drop calls dispose, which aborts the controller, which kills the underlying connection and its body stream out from under whoever is still reading it - even though the request actually succeeded.

Confirmed via a real Chromium browser (this doesn't reproduce under Node's native fetch/undici in the vitest suite, since it tears down streams differently - the regression test I added asserts on the call directly instead):
- Traced the abort call stack straight to [Symbol.dispose] being invoked right after the AWS SDK's connector obtained the response and started reading the body.
- Confirmed via AbortController.prototype.abort instrumentation that abort() fires even though the fetch promise had already resolved with a 200.
- The equivalent code in the published 0.23.0 doesn't call abort() in dispose at all - this looks like a regression introduced since then.

eduardomourar:
Fixes #2044.

The browser IncomingBody's [Symbol.dispose] unconditionally called controller.abort(), even after the fetch had already resolved successfully and its response body was still being streamed elsewhere. Since the guest typically drops the future-incoming-response handle right after extracting the response - well before the body is fully read. This aborted the underlying connection out from under whoever was still reading it, surfacing as a bogus "BodyStreamBuffer was aborted" error on requests that actually succeeded (e.g. aws-cli's `s3 ls`, discovered while validating the would-block fix in #2043). Only abort if the request never settled.

### DIFF
--- a/packages/preview2-shim/src/browser/http.ts
+++ b/packages/preview2-shim/src/browser/http.ts
@@ -873,6 +873,7 @@ class FutureIncomingResponse implements TypesNamespace.FutureIncomingResponse {
     #result: any = undefined;
     #promise: Promise<void> | null = null;
     #controller: AbortController | null = null;
+    #settled = false;
 
     subscribe(): Pollable {
         return pollableCreate(this.#promise!);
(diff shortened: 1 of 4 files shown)


vados-cosmonic (approved):
LGTM :rocket:

Added some tests

vados-cosmonic:
There are a *ton* of commits to various Jco packages that need to go out, but we will try and get this out ASAP, thanks again @eduardomourar

PR #2047, opened 2026-09-06
feat(preview2-shim): implement basic advise method for browser in-memory-filesystem

eduardomourar:
The `advise()` on the browser in-memory filesystem was a complete no-op — the advisory hint was accepted and
discarded. This tracks the last hint per descriptor, so future backends with real storage (e.g. an OPFS adapter) can act on it. Node's `advise()` already forwards to a native `fadvise` call and is unaffected.

- src/browser/in-memory-filesystem.ts: advise() stores the hint (#advice, default "normal").
- test/browser/filesystem.ts: covers hint retention and the existing bad-descriptor behavior for directories.


vados-cosmonic (approved):
LGTM 🚀

PR #2048, opened 2026-09-06
feat(preview2-shim): support follow symlink for node get-full-path

eduardomourar:
Most other call sites (statAt, setTimesAt, openAt) already handled follow/no-follow correctly via statSync/lstatSync or O_NOFOLLOW directly, independent of this parameter — this fix specifically closes the gap for linkAt's old path and any other caller relying on #getFullPath to do the resolution.

- src/nodejs/filesystem.ts: #getFullPath now actually uses its followSymlinks parameter. When set, it resolves the syntactically-joined path via realpathSync (falling back to the unresolved path if resolution fails, e.g. for open-at with create on a not-yet-existing file). This fixes linkAt with `oldPathFlags.symlinkFollow: true`, which previously always hard-linked the symlink itself instead of its target (Node's linkSync doesn't dereference symlinks on its own).
- Added a test in test/node.ts creating a real file + symlink in a temp dir and verifying linkAt with `symlinkFollow: true` hard-links the resolved target.

### DIFF
--- a/packages/preview2-shim/src/nodejs/filesystem.ts
+++ b/packages/preview2-shim/src/nodejs/filesystem.ts
@@ -303,7 +303,7 @@ class Descriptor implements IDescriptor {
     }
 
     createDirectoryAt(path) {
-        const fullPath = this.#getFullPath(path, undefined);
+        const fullPath = this.#getFullPath(path);
         try {
             mkdirSync(fullPath);
         } catch (e) {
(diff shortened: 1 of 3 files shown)


vados-cosmonic (approved):
LGTM 🚀 

Running this through AI it looks like there were some issues with trailing slash handling, and a resolution issue where linkAt should *not* succeed due to the semantics of the WIT interface along with some tests to make sure we don't regress

vados-cosmonic:
Merge queue failure is a flake in `test-extended`, so going to merge this

PR #2055, opened 2026-09-06
feat(p2-shim): add missing socket error-code mapping

eduardomourar:
Finish the `catch`/`switch` arms mapping Node/browser socket errors to the correct `wasi:sockets/network#error-code` variants.

- packages/preview2-shim/src/io/worker-sockets.ts: filled in the previously-TODO'd error mappings in both convertSocketError (string-code path) and convertSocketErrorCode (numeric-code path) for new-socket-limit (EMFILE/ENFILE), remote-unreachable (EHOSTUNREACH/EHOSTDOWN/ENETUNREACH/ENETDOWN), and datagram-too-large (EMSGSIZE). Left name-unresolvable/temporary-resolver-failure out of convertSocketErrorCode with an explanatory comment — Node has no portable numeric errno for DNS resolution failures; those are already handled correctly via string .code values in socketResolveAddress.
- Added packages/preview2-shim/test/socket-error-codes.ts covering the new mappings for both functions.

### DIFF
--- a/packages/preview2-shim/src/io/worker-sockets.ts
+++ b/packages/preview2-shim/src/io/worker-sockets.ts
@@ -24,7 +24,13 @@ import {
     ECONNABORTED,
     ECONNREFUSED,
     ECONNRESET,
+    EHOSTUNREACH,
     EINVAL,
+    EMFILE,
+    EMSGSIZE,
+    ENETDOWN,
+    ENETUNREACH,
+    ENFILE,
     ENOBUFS,
     ENOMEM,
     ENOTCONN,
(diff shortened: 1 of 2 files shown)


vados-cosmonic (approved):
LGTM 🚀 

Thanks for adding these @eduardomourar !

vados-cosmonic:
Ah hey so @eduardomourar the scope for preview2-shim is actually `p2-shim` -- if you check `commitlint.config.mjs` to see the lists, but to be honest we might want an AGENTS.md with this kind of stuff to make it easier to work on, so if you want to contribute one, would be happy to have it (I'll add on to it later as well)!

eduardomourar:
Apologies for that mistake. I modified the commit title in both open PRs.

vados-cosmonic:
No worries, I often have to look these up as well since they're sometimes not intuitive, given some are shorter and some are not, hopefully we can get this change in quickly (the rebase should have also gotten rid of some issues in tests)

PR #2088, opened 2026-09-09
test(transpile): add failing test for cooperative threading intrinsics

TartanLlama:
I've ported over one of the [POSIX Open Test suite tests](https://github.com/bytecodealliance/wasi-sdk/blob/main/src/wasi-libc/test/open-posix-test-suite/conformance/interfaces/pthread_mutex_lock/1-1.c) that we use in testing wasi-libc's threading implementation. 

My idea is that we can use this as basic validation for the jco implementation, then port over more tests from there to build more confidence in it and find edge cases.


vados-cosmonic (dismissed):
LGTM 🚀 

Thanks for updating the SDK version!

vados-cosmonic:
BTW commitlint error:

```
input: tests(transpile): skip test
✖   type must be one of [build, chore, ci, debug, docs, feat, fix, perf, refactor, release, revert, style, test, sec] [type-enum]

✖   found 1 problems, 0 warnings
```

looks like `tests` -> `test` on that list commit title (again sorry this is so strict!)

vados-cosmonic:
OK, so it looks like there might actually be an unexpected bug that has nothing to do with this PR but triggers it because of the sdk update? Will look into it

vados-cosmonic:
OK, so it looks like there was a problem calling destructors in components that have this sort of... outside-of-task execution via core functions, fix is https://github.com/bytecodealliance/jco/pull/2089 -- once it merges I think you should be able to rebase to `main` and this PR should be 🟢

vados-cosmonic (approved):

PR #2093, opened 2026-09-09
feat(std): implement Node.js readline APIs

vados-cosmonic:


vados-cosmonic:
We still need `node:process` support to land before we can fully implement readline APIs

Issue #144, opened 2023-09-02
Leverage File System Standard from browsers


eduardomourar:
In regards to the Preview2 shim (more specifically `wasi:filesystem`) for browsers, should we leverage the [File System Standard](https://fs.spec.whatwg.org/#idl-index) as defined by the WHATWG?

guybedford:
It's a good question, and a tough one as well when working through the various concerns. The best I can come up with so far is that I think we should **integrate** with it as opposed to treating it as the entire interface, for example allowing for mounting a local directory as a preopen within the virtual filesystem, but while still being able to have other types of virtual filesystem management existing in parallel. Generic virtualization might allow other kinds of integrations as well similarly to the virt work.

eduardomourar:
Where do you think the integration can/should be placed? Pluggable at a higher level (JavaScript in the shim) or lower level (e.g. Rust similar to the preview1 adapter)?

guybedford:
I think it's best suited as a feature of the prevew2-shims? See eg https://github.com/bytecodealliance/jco/pull/147.

PR #2095, opened 2026-09-10
feat(p2-shim): add opfs filesystem adapter

eduardomourar:
Adds a first-party [OPFS](https://developer.mozilla.org/en-US/docs/Web/API/File_System_API/Origin_private_file_system)-backed filesystem adapter for `preview2-shim`'s browser `wasi:filesystem` implementation, plus two non-JSPI improvement candidates (advisory locking, in-memory symlink support).

- **`OpfsFilesystemAdapter`** (`src/browser/opfs-filesystem.ts`) - loads a preopened OPFS directory into an in-memory tree once (`loadOpfsCapability()`), so every `Descriptor` op stays fully synchronous - no JSPI required. Mutations flush back to OPFS automatically (debounced to a microtask); `flush()`/`dispose()` force it explicitly.
- **Symlinks** survive a flush + reload via a hidden sidecar file (`.__wasi_symlinks__.json`) per root, since OPFS has no native symlink concept.
- **Advisory locking** (`lockShared`/`lockExclusive`/`tryLock*`/`unlock`) on `BrowserFilesystemDescriptor`, backed by a same-process reader/writer lock in `InMemoryFilesystemAdapter`. `new OpfsFilesystemAdapter({ lockManager: navigator.locks })` opts into best-effort cross-tab coordination via the Web Locks API - off by default. `lockManager` takes a narrow `BrowserLockManager` interface rather than the global `navigator` directly, so it's easy to fake in tests.
- **`_addPreopenWithAdapter`** (`src/browser/filesystem.ts`) lets a host wire a custom adapter into the top-level `wasi:filesystem/preopens` singleton that transpiled components import statically.

Related to: #144

### DIFF
--- a/packages/preview2-shim/src/browser/opfs-filesystem.ts
+++ b/packages/preview2-shim/src/browser/opfs-filesystem.ts
@@ -1,5 +1,5 @@
 import type { BrowserFilesystemAdapter, BrowserFilesystemDescriptor } from "./filesystem.js";
-import { InMemoryFilesystemAdapter, _onTouch } from "./in-memory-filesystem.js";
+import { InMemoryFilesystemAdapter, UNWRAP_DESCRIPTOR, _onTouch } from "./in-memory-filesystem.js";
 import type { FileData, FileDataEntry } from "./in-memory-filesystem.js";
 
 /**
@@ -218,6 +218,9 @@ function withCrossTabLocking(
 
     return new Proxy(descriptor, {
         get(target, prop, receiver) {
+            if (prop === UNWRAP_DESCRIPTOR) {
+                return target;
+            }
             const value = Reflect.get(target, prop, receiver);
             if (prop === "openAt") {
                 return (...args: Parameters<BrowserFilesystemDescriptor["openAt"]>) => {
--- a/packages/preview2-shim/src/browser/opfs-filesystem.ts
+++ b/packages/preview2-shim/src/browser/opfs-filesystem.ts
@@ -19,9 +19,79 @@ type IterableDirectoryHandle = FileSystemDirectoryHandle & {
     entries(): AsyncIterableIterator<[string, FileSystemFileHandle | FileSystemDirectoryHandle]>;
 };
 
+// OPFS has no native symlink concept. All symlinks under a root are recorded in a
+// single hidden sidecar file at that root, keyed by their path relative to it, so
+// they survive a flush + reload instead of only living for the lifetime of one
+// capability.
+const SYMLINKS_FILE = ".__wasi_symlinks__.json";
+
+async function readSymlinksFile(handle: FileSystemDirectoryHandle): Promise<Record<string, string>> {
+    try {
+        const fileHandle = await handle.getFileHandle(SYMLINKS_FILE);
+        const file = await fileHandle.getFile();
+        return JSON.parse(await file.text());
+    } catch {
+        return {};
+    }
+}
+
+async function writeSymlinksFile(
+    handle: FileSystemDirectoryHandle,
+    symlinks: Record<string, string>,
+): Promise<void> {
+    if (Object.keys(symlinks).length === 0) {
+        try {
+            await handle.removeEntry(SYMLINKS_FILE);
+        } catch {
+            // no sidecar to remove
+        }
+        return;
+    }
+    const fileHandle = await handle.getFileHandle(SYMLINKS_FILE, { create: true });
+    const writable = await fileHandle.createWritable();
+    await writable.write(JSON.stringify(symlinks));
+    await writable.close();
+}
+
+/** Collect every symlink under `dir` into a flat path -> target map, relative to `dir` itself. */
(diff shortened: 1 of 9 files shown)


packages/preview2-shim/test/browser/opfs-filesystem.ts
vados-cosmonic, line 1: In this file, what do you think about creating the adapter explicitly rather than loading and using the default bindings? I'd like for as many tests as we can have to be `.concurrent` and I *think* if these tests actually ran at the same time they'd interfere with each other

packages/preview2-shim/test/browser/filesystem.ts
```diff
@@ -298,4 +298,39 @@ suite("Browser filesystem", () => {
 
         assert.deepStrictEqual(calls, ["openAt", "advise", "syncData", "sync"]);
     });
+
+    test("advisory locking: shared locks stack, exclusive locks are exclusive", async () => {
+        const { _setFileData, preopens } = await import("../../src/browser/filesystem.js");
```
vados-cosmonic: can we follow the  convention in the other tests where we make an explicit `filesystem` via `createFileSystem` ? I think it would make it easier for these tests to be concurrent

packages/preview2-shim/src/browser/opfs-filesystem.ts
```diff
@@ -0,0 +1,285 @@
...
+                    }
+                    return result;
+                };
+            }
+            return typeof value === "function" ? value.bind(target) : value;
+        },
+    });
+}
+
+export interface OpfsFilesystemAdapterOptions {
```
vados-cosmonic, line 259: Can we also make `navigator` an option? I think this would really help testability and possibly some other weird use cases (browser is never going to have multiple AFAIK, but I'd like to reserve the flexibility!)

For example passing through something that proxies `navigator`, rather than every adapter always being tied to the "real" accessible `navigator`.

vados-cosmonic (dismissed):
This is awesome! Thanks for working on this -- most of my nits are more about testing and reserving some future optionality. 

One more thing I'd really love to have is an E2E test in the browser! We have [machinery and browser automation already set up](https://github.com/bytecodealliance/jco/blob/a0414eb81d4b1a4db9d440fafb86a1984ae0776b/packages/preview2-shim/test/common.ts#L71), but for some reason we don't actually *use it* from any of the tests... I think you should be able to copy the setup in other projects (`jco`, primarily), but also happy to leave that as a follow up (we can make a new issue to track that separately, it would be very easy for anyone else to pick up), if you'd rather get this in as-is!

packages/preview2-shim/test/browser/opfs-filesystem.ts
eduardomourar: I think that it makes sense, yes.

eduardomourar: Done — every test in this file is now test.concurrent. The cross-tab-locking test no longer mutates globalThis.navigator either (which would've broken under real concurrency); it now passes a fake lockManager directly via `new OpfsFilesystemAdapter({ lockManager: fakeLockManager })`, using the option added in response to your comment below.

packages/preview2-shim/test/browser/filesystem.ts
eduardomourar: Done — both advisory-locking tests now build an isolated `createFilesystem({ adapter: new InMemoryFilesystemAdapter(), preopens: {...} })` and are test.concurrent, matching the pattern a few tests above them.

packages/preview2-shim/src/browser/opfs-filesystem.ts
eduardomourar: Added — OpfsFilesystemAdapter now takes a lockManager option (folded the earlier crossTabLocking/navigator split into one — passing a lock manager is the opt-in). It's typed as BrowserLockManager (Omit<LockManager, "query">) rather than tied to the global navigator, so it can be faked in tests or proxied for other cases, per your suggestion.

eduardomourar:
Added e2e coverage using the harness in test/common.ts (startTestServer/runBasicHarnessPageTest), in opfs-filesystem-e2e.ts:
- one suite componentizes/transpiles a small wasm component, runs it against a live OpfsFilesystemAdapter, flushes, reloads the same OPFS directory into a fresh adapter, and re-runs the component — confirming writes/symlinks made across the component boundary actually survive in real OPFS storage, not just in one adapter's in-memory tree.
- a second suite exercises cross-tab locking directly against the browser's real navigator.locks (no wasm component needed there) — verifying two independent adapters actually queue/grant through the browser's own Web Locks manager, not a mock.

vados-cosmonic (approved):
LGTM :rocket:

I think some of the new tests could do without the `beforeAll` (since they are testing results from a single run), but that can definitely get fixed in a follow up!)

vados-cosmonic (approved):

PR #2132, opened 2026-09-16
feat(bindgen): introduce skip task management mode in function bindgen

andreiltd:
Using `skip_task_management(true)` mode allows user to skip the task machinery and control execution and abi call lifetimes.

See also https://github.com/bytecodealliance/ComponentizeJS/issues/350#issuecomment-5692932194 for more details on rationale for introducing this.


andreiltd:
@vados-cosmonic I'm verifying this locally with `ComponentizeJS` and let you know the result as soon as it's finished.

Edit: `ComponentizeJS` regression is passing locally with those changes.

vados-cosmonic:
Wow that was quick -- Thanks @andreiltd !

crates/js-component-bindgen/src/function_bindgen.rs
```diff
@@ -50,6 +50,25 @@ impl ErrHandling {
...
+    /// Let a native embedding manage execution and canonical ABI call lifetimes.
+    ///
+    /// Calls, result conversion, borrowed-resource cleanup, dealloc, and
+    /// `post_return` are generated without the host-side task runtime. Async
+    /// JavaScript implementations can still be awaited with
+    /// [`FunctionBindgen::requires_async_porcelain`].
+    ///
+    /// This mode supports the synchronous canonical ABI and guest resource
+    /// tables without async metadata.
+    Embedding,
```
vados-cosmonic: If you end up leaving this in, maybe `EmbeddingManaged`? If the enum changes to `FuncCallManager` then maybe it could be only `Embedding` ?

crates/js-component-bindgen/src/function_bindgen.rs
```diff
@@ -50,6 +50,25 @@ impl ErrHandling {
     }
 }
 
+/// Execution environment used by generated function calls.
+#[derive(Debug, Default, Clone, Copy, PartialEq, Eq)]
+#[non_exhaustive]
+pub enum CallRuntime {
```
vados-cosmonic: I'm not sure I love this name -- it doesn't really convey what it's doing... 

Something like `FuncCallManagement`/`FuncCallBookeeping`? Maybe even going with a boolean might be enough here, something like `bypass_task_management` or `skip_task_management` might be clearer as an option.

crates/js-component-bindgen/src/function_bindgen.rs
```diff
@@ -357,7 +380,41 @@ fn unambiguous_result_wrapper_tags(resolve: &Resolve, ty: Option<&Type>) -> (boo
     }
 }
 
+fn instruction_requires_task_runtime(inst: &Instruction<'_>) -> bool {
```
vados-cosmonic, line 368: I think this could be a method on `Instruction` actually

vados-cosmonic (dismissed):
LGTM 🚀 

Really only nits here. 

One thing I do wonder about is if we actually want the option to be able to do this on a function-by-function basis? Right now it's all or nothing, but if we do stick with the enum that would be one benefit you could take a list of function names or something to do the simple call for? Maybe that's silly and we don't need it though (like I said in a comment I do think we could actually get away with a boolean here actually)

crates/js-component-bindgen/src/function_bindgen.rs
andreiltd: The `Instruction` type comes from `wit-bindgen` crate, or do you mean method on `FunctionBindgen`?

crates/js-component-bindgen/src/function_bindgen.rs
andreiltd: I'm happy to go with `skip_task_management(true)` 🙂

andreiltd:
> One thing I do wonder about is if we actually want the option to be able to do this on a function-by-function basis?

I'm not sure if I follow 😅, the API is currently per function so each `FunctionBindgen`  has its own `call_runtime` (or `skip_task_management: bool` if we go with bool). I think it's reasonable granularity.

crates/js-component-bindgen/src/function_bindgen.rs
vados-cosmonic: Yeah I was thinking of an extension trait here, but a bare function is also fine -- probably too much!

crates/js-component-bindgen/src/function_bindgen.rs
vados-cosmonic: yeah, especially if we don't really see it going past two cases.

vados-cosmonic:
Ah yeah sorry, I was thinking of this from the `jco-transpile` side and wanting to specify this up front all at once -- definitely don't need that since you're using the `FunctionBindgen` implementationss directly

crates/js-component-bindgen/src/function_bindgen.rs
```diff
@@ -139,6 +139,11 @@ pub type ResourceMap = BTreeMap<TypeId, ResourceTable>;
 #[derive(bon::Builder)]
 #[non_exhaustive]
 pub struct FunctionBindgen<'a> {
+    /// Skip generated task management when the caller already manages execution
+    /// and canonical ABI call lifetimes. Defaults to `false`.
```
vados-cosmonic: SUPER NIT: It would be nice to explain a case (like... exactly your use case) when someone would need this as a tiny comment under the main description here.

I'm imagining something like "This is normally used by embedders when they want to ...."

crates/js-component-bindgen/src/function_bindgen.rs
```diff
@@ -139,6 +139,11 @@ pub type ResourceMap = BTreeMap<TypeId, ResourceTable>;
...
+    #[builder(default)]
+    pub skip_task_management: bool,
```
vados-cosmonic:
```suggestion
    pub skip_fn_call_task_management: bool,
```

Another super nit here, maybe too lengthy but it might be nice to be extremely explicit here

vados-cosmonic (dismissed):
OK, really nothing but (super) nits at this point!

vados-cosmonic (approved):

Issue #2134, opened 2026-09-17
--instantiation async: WIT imports named reject/resolve collide with helper locals


GreenBusDriver:
Discussed briefly on Zulip in [`jco` thread](https://gist.github.com/GreenBusDriver/4b75336dd1246704d8c687792c346c1e); filing here as requested.
## Summary
With `jco transpile --instantiation async`, importing a WIT function named
`reject` (also `resolve`, and likely `promise` / `gen` / `runNext` /
`maybeSyncReturn`) produces JS that fails to load:
SyntaxError: Identifier 'reject' has already been declared

`reject` / `resolve` are legal WIT identifiers (not WIT keywords, not JS
keywords). The collision is with locals that the instantiation helper emits
in the same function scope.
## WIT shape
```wit
package demo:jco;
interface api {
    reject: func(id: u32);
}
world demo {
    import api;
    export run: func(id: u32);
}
```

## Observed
Generated instantiate() contains, in one scope:
```
const { reject } = imports['demo:jco/api'];
// ...
let promise, resolve, reject;   // instantiation helper
```

## Expected
Either the import or the helper locals are deconflicted (e.g. reject$1), same as other name clashes — module loads cleanly.

Root cause (best guess)
Imports go through LocalNames in render_imports, but the helper is a raw string in transpile_bindgen.rs (let promise, resolve, reject; plus gen / runNext / maybeSyncReturn) and those names are never registered via exclude_globals / create_once. So the deconflicter never sees them.

Specific to --instantiation mode (default ESM puts the helper in a $init IIFE, so this is shadowing rather than redeclaration).

Related precedent: #1898 / #1899 (exported eval not sanitized) — same “bindings must produce valid JS for any legal WIT name” principle.

## Minimal repro
No wasm toolchain needed:

https://gist.github.com/GreenBusDriver/4b75336dd1246704d8c687792c346c1e
```
npm install
npm run demo
# FAIL: Identifier 'reject' has already been declared
```
Control: rename reject → anything else in demo.wit → OK.

## Suggested fix
Add to `RESERVED_KEYWORDS` list https://github.com/bytecodealliance/jco/blob/2da9d4b3271e039af0cc6c8bd8281f2b1759fe4d/crates/js-component-bindgen/src/names.rs#L169
or
Register the helper locals in LocalNames (or add them to exclude_globals), so an imported reject becomes reject$1 automatically — no WIT rename needed.

## Environment
[...]

PR #2135, opened 2026-09-19
fix(bindgen): deconflict instantiation helper locals from WIT import names

GreenBusDriver:
Addresses #2134 
The `instantiate()` / `$init` helper hard-coded its locals (`promise`,
`resolve`, `reject`, `gen`, `runNext`, ...) as literal text, so a component
importing a WIT function with one of those names produced JS that fails to
load with `Identifier 'reject' has already been declared`.

Allocate the helper locals through `LocalNames` like every other generated
binding, so the collision is resolved as `reject$1` automatically.

Zulip link: [#jco > reject/resolve import name collision w instantiation helper @ 💬](https://bytecodealliance.zulipchat.com/#narrow/channel/409526-jco/topic/reject.2Fresolve.20import.20name.20collision.20w.20instantiation.20helper/near/625034686)


vados-cosmonic (dismissed):
LGTM 🚀 


Thanks @GreenBusDriver this looks great, happy to get this into a release ASAP!

